PR #6, opened 2019-10-21
Update Tapyrus specifications and help messages

Naviabheeman:
Updated Tapyrus specifications to reflect changes due to segwit removal.
Updated help in generate and generatetoaddress rpc.
Changed "Bitcoin" to "Tapyrus" in all comments and text messages.


src/rpc/mining.cpp
```diff
@@ -97,7 +97,8 @@ void ParsePrivateKeyList(const UniValue& privkeys_hex, std::vector<CKey>& vecKey
     for(int i=0; i< privkeys_hex.size(); i++) {
         const UniValue& privkey = privkeys_hex[i];
         std::string keyHex = privkey.get_str();
-        if(keyHex.length() % 64 != 0) {
+        if(keyHex.length() != 64 
+        && keyHex.length() != 52) {
```
azuchi: What is the length 52 check for?

Naviabheeman: this change should not be here, sorry. it was made while testing the issue reported on error message.

azuchi:
@Naviabheeman please rebase this request.

PR #20, opened 2019-11-21
Remove support for BIP9 soft forks from Tapyrus

Naviabheeman:
Remove BIP9 based soft fork infrastructure completely.
Make tapyrus block version 1. Any block other version number is rejected.


src/consensus/params.h
```diff
@@ -55,7 +28,6 @@ struct Params {
      */
     uint32_t nRuleChangeActivationThreshold;
     uint32_t nMinerConfirmationWindow;
```
azuchi: `nRuleChangeActivationThreshold` and `nMinerConfirmationWindow`  seem to be BIP9 parameters.

Naviabheeman:
> `vbparams` in `init.cpp` BIP9 related parameter.

Removed vbparams, nRuleChangeActivationThreshold and nMinerConfirmationWindow. 
Removed bit soft forks from rpc help test.

PR #37, opened 2020-01-08
NetworkId parameter to start multiple Tapyrus-networks 

Naviabheeman:
Added "networkid" command line option to configure separate network id.
 - network magic bytes are calculated based on this network id
 - datadir name depends on this parameter: it is now `<chain>-<networkid> `and genesis file name is `genesis.<networkid>`

Refactoring:
CChainParams is all constants (different but fixed for main and regtest modes.)
CBaseChainParams contains all the network parameters - magic bytes, genesis blocks, seeds and datadir etc. This PR does not make seeds configurable. I'll add it in the next PR.

Removed -testnet and -paradium modes.
Moved tapyrus modes to new header and defined default network Ids for two modes - MAIN and REGTEST.

Updated makefile - changed target binaries and link binaries to resolve build errors 
Updated unit tests and functional tests to use networkid parameter.

PR #39, opened 2020-01-09
Rename MAIN & REGTEST to PROD & DEV 

Naviabheeman:
Since Tapyrus has only two modes they are being renamed from MAIN and REGTEST to PROD and DEV. 
Class CChainParamsBase is renamed to CFederationParams
Classes CMainParams and CRegtestParams are renamed to CProductionParams and CDevelopmentParams.

These changes are made over #37 code base. This PR can be merged after #37


src/test/chainparamsbase_tests.cpp
```diff
@@ -15,7 +15,7 @@ extern void noui_connect();
 
 struct ChainParamsBaseTestingSetup {
```
azuchi: You should also rename to FederationParamsBaseTestingSetup. filename and following code too.

Naviabheeman: yes! will correct it.

src/util.h
```diff
@@ -261,8 +261,8 @@ class ArgsManager
     void ForceSetArg(const std::string& strArg, const std::string& strValue);
 
     /**
-     * Looks for -regtest and returns the appropriate BIP70 chain name.
-     * @return CBaseChainParams::MAIN by default; raises runtime error if an invalid combination is given.
+     * Looks for -dev and returns the appropriate BIP70 chain name.
+     * @return CFederationParams::MAIN by default; raises runtime error if an invalid combination is given.
```
azuchi: Is'nt it `TAPYRUS_OP_MODE::PROD`?

azuchi:
In `doc/tapyrus` still remain main, regtest.

PR #41, opened 2020-01-27
Fix tapyrus-cli client don't work in testproposeblock

rantan:
Current implementation try to convert unnecessary string as JSON string to value.


Naviabheeman:
I guess the same is happening in `combineblocksigs` because of the line  
`{ "combineblocksigs", 0, "block_hex"},`
Should we change that too?

rantan:
> I guess the same is happening in `combineblocksigs` because of the line
> `{ "combineblocksigs", 0, "block_hex"},`
> Should we change that too?

Yes. and also `{ "combineblocksigs", 1, "signature_list"},` need to be removed.

I'll update this PR. thanks.

PR #48, opened 2020-02-21
Relaxing Tapyrus standard rules

Naviabheeman:
Made witness script types available only in debug mode to use for testing. Witness related functional tests are run only in debug mode.
Added TX_CUSTOM script type to accept all scripts which do not use reserved op codes.
Updated unit tests, util and functional tests with matching output.


src/test/multisig_tests.cpp
```diff
@@ -203,7 +203,10 @@ BOOST_AUTO_TEST_CASE(multisig_IsStandard)
     malformed[5] << OP_1 << ToByteVector(key[0].GetPubKey()) << ToByteVector(key[1].GetPubKey());
 
     for (int i = 0; i < 6; i++)
-        BOOST_CHECK(!::IsStandard(malformed[i], whichType));
+    {
+        BOOST_CHECK(::IsStandard(malformed[i], whichType));
+        BOOST_CHECK(whichType == TX_CUSTOM);
+    }
```
Naviabheeman: Please not this case - malformed multisig scripts are also accepted in TX_CUSTOM category. Should we check this more strictly?

Naviabheeman: Improved this in the new commits - Now I try to identify the type of script (p2pkh, p2sh and multis) from the script op code and flag it as nonstandard before checking disallowed opcodes to tag the script as custom.

src/script/standard.cpp
```diff
@@ -106,7 +171,7 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
     std::vector<unsigned char> witnessprogram;
     if (scriptPubKey.IsWitnessProgram(witnessversion, witnessprogram)) {
         typeRet = TX_NONSTANDARD;
-        return true;
+        return false;
     }
```
azuchi: leave it true. Treated as a script pushing two elements.

src/script/standard.cpp
```diff
@@ -120,37 +185,51 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
...
+    if ((typeRet == TX_PUBKEY ||
+        typeRet == TX_PUBKEYHASH ||
+        typeRet == TX_MULTISIG ||
+        typeRet == TX_SCRIPTHASH) &&
+        vSolutionsRet.empty())
+        {
+            typeRet = TX_NONSTANDARD;
+            return false;
+        }
+
```
azuchi: not need this check. It is the responsibility of the user who created the script to make sure the script is correct.

src/test/multisig_tests.cpp
azuchi: malformed multisig scripts should be accepted. we do not check it.

Naviabheeman: Then I'll revert back the newer commits. it will restore the expected behaviour.

src/script/standard.cpp
```diff
@@ -88,6 +98,61 @@ static bool MatchMultisig(const CScript& script, unsigned int& required, std::ve
...
+            opcode == OP_RESERVED1 ||
+            opcode == OP_RESERVED2 ||
+            opcode ==  OP_NOP1 ||
+            opcode ==  OP_NOP4 ||
+            opcode ==  OP_NOP5 ||
+            opcode ==  OP_NOP6 ||
+            opcode ==  OP_NOP7 ||
+            opcode ==  OP_NOP8 ||
+            opcode ==  OP_NOP9 ||
+            opcode ==  OP_NOP10)
```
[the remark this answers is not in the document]
Naviabheeman: OP_NOP are rejected in a different location https://github.com/chaintope/tapyrus-core/blob/master/src/script/interpreter.cpp#L485-L370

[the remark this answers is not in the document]
azuchi: oh, I see.

PR #50, opened 2020-03-02
Remove Coinbase maturity limitation from Tapyrus.

Naviabheeman:
Removed Coinbase maturity definition and usage from transaction validation code.
As a result tapyrus wallet does not need to remember immature balance. 
Updated unit tests to create chain of length "5" instead of 100. (some tests needed multiple outputs)
Removed all coinbase maturing code from functional tests i.e create only one block (not 100) to get a spendable coinbase output.
Updated wallet balances in some functional tests - as blocks rewards mature immediately there are tests where wallet balances had to take into account the transaction fee and block reward.

### DIFF
--- a/src/federationparams.cpp
+++ b/src/federationparams.cpp
@@ -159,8 +159,6 @@ CPubKey CFederationParams::ReadAggregatePubkey(const std::vector<unsigned char>&
 
 bool CFederationParams::ReadGenesisBlock(std::string genesisHex)
 {
-    ECCVerifyHandle globalVerifyHandle;
-
     CDataStream ss(ParseHex(genesisHex), SER_NETWORK, PROTOCOL_VERSION);
     unsigned long streamsize = ss.size();
     ss >> genesis;
--- a/src/federationparams.cpp
+++ b/src/federationparams.cpp
@@ -159,6 +159,8 @@ CPubKey CFederationParams::ReadAggregatePubkey(const std::vector<unsigned char>&
 
 bool CFederationParams::ReadGenesisBlock(std::string genesisHex)
 {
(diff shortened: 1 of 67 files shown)


src/federationparams.cpp
```diff
@@ -159,6 +159,8 @@ CPubKey CFederationParams::ReadAggregatePubkey(const std::vector<unsigned char>&
 
 bool CFederationParams::ReadGenesisBlock(std::string genesisHex)
 {
+    ECCVerifyHandle globalVerifyHandle;
+
```
azuchi: What is this use for?

src/interfaces/wallet.h
```diff
@@ -321,7 +320,7 @@ struct WalletBalances
     bool balanceChanged(const WalletBalances& prev) const
     {
         return balance != prev.balance || unconfirmed_balance != prev.unconfirmed_balance ||
-               immature_balance != prev.immature_balance || watch_only_balance != prev.watch_only_balance ||
+               watch_only_balance != prev.watch_only_balance ||
                unconfirmed_watch_only_balance != prev.unconfirmed_watch_only_balance ||
                immature_watch_only_balance != prev.immature_watch_only_balance;
```
azuchi: Is `immature_watch_only_balance` still necessary?

src/federationparams.cpp
Naviabheeman: It was needed in unit tests for initialising verification context of signature verification. IT should be in the constructor so I removed from here.

src/interfaces/wallet.h
Naviabheeman: no, I removed it

PR #54, opened 2020-03-10
snapify

metalefty:


metalefty:
@rantan done. merge at your convenience.

PR #63, opened 2020-04-06
snap: add missing run-time libraries

metalefty:
snap definition fixes only

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -65,3 +65,9 @@ parts:
       - libssl1.0-dev
     stage-packages:
       - libboost-system1.65.1
+      - libboost-filesystem1.65.1
+      - libboost-thread1.65.1
+      - libboost-chrono1.65.1
+      - libdb5.3++
+      - libevent-2.1-6
+      - libevent-pthreads-2.1-6


metalefty:
I missed that was shown in snap build log.

```
./configure --prefix= --disable-dependency-tracking --enable-zmq --enable-glibc-back-compat --enable-reduce-exports --with-gui=no --with-incompatible-bdb --disable-bench --disable-tests
make -j4
make install DESTDIR=/build/snapcraft-tapyrus-core-8e96cd/parts/tapyrus-core/install
Staging tapyrus-core
Priming tapyrus-core
The 'tapyrus-core' part is missing libraries that are not included in the snap or base. They can be satisfied by adding the following entries to the existing stage-packages for this part:
- libboost-chrono1.65.1
- libboost-filesystem1.65.1
- libboost-thread1.65.1
- libdb5.3++
- libevent-2.1-6
- libevent-pthreads-2.1-6
Could not find '/snap/core/current/usr/bin/snap', validation of the license string will only take place once pushed to the store.
Snapping 'tapyrus-core' ...

Snapped tapyrus-core_0+git.3d1eaa285_multi.snap
Revoking proxy token...
RUN: /usr/share/launchpad-buildd/bin/in-target scan-for-processes --backend=lxd --series=bionic --arch=i386 SNAPBUILD-880533
Scanning for processes to kill in build SNAPBUILD-880533
```

Issue #61, opened 2020-04-01
Remove Openssl dependency from Tapyrus


Naviabheeman:
Openssl is not a dependency of bitcoin anymore. We can remove it from tapyrus too.
Port https://github.com/bitcoin/bitcoin/pull/17265

Check if https://github.com/bitcoin/bitcoin/pull/10299 is relevant.

Naviabheeman:
Following PRNG related improvements and some bug fixes can to be merged to remove dependency on openssl random.

- https://github.com/bitcoin/bitcoin/pull/14955
- https://github.com/bitcoin/bitcoin/pull/17507
- https://github.com/bitcoin/bitcoin/pull/17507
- https://github.com/bitcoin/bitcoin/pull/17265
- https://github.com/bitcoin/bitcoin/pull/17270

metalefty:
Is this to be done until 0.4.0?

Naviabheeman:
yes, it is in progress (#64)  and will be in 0.4.0.

PR #64, opened 2020-04-07
Remove openssl and Improve rand

Naviabheeman:
Ported random number generation and openssl removal code from bitcoin core.

PR #67, opened 2020-04-09
snap: add run-time OpenSSL dependency

metalefty:
OpenSSL is required until #61 is finished.

While here, quit build on i386. It is almost unnecessary.


metalefty:
@rantan dependency on OpenSSL is to be removed later but to make snap package right now, could you merge this? After OpenSSL removal, I'll revert this.

metalefty:
Overcome by #64.

PR #68, opened 2020-04-09
60 federation params aggpubkey

winnie-chaintope:


src/federationparams.h
```diff
@@ -35,14 +40,42 @@ class CFederationParams
...
+    const uint& GetHeightFromAggregatePubkey(std::vector<unsigned char> aggpubkey) const {
+        for (auto& c : aggregatePubkeyHeight)
+            if (c.aggpubkey == CPubKey(aggpubkey.begin(), aggpubkey.end())) {
+                return c.height;
+                break;
+            } else {
+                continue;
+            }
+    };
+    const CPubKey& GetAggPubkeyFromHeight(uint height) const {
```
Naviabheeman, line 60: `if((height < 1) || (aggregatePubkeyHeight.size() == 1))` double parentheses '((' are not necessary.
`decltype(aggregatePubkeyHeight.size())` is accurate, but unnecessary, we can use `unsigned int `instead. 

I'm not sure if it is safe to access element (i+1) in a loop running from i = 0 to size - 1.  for new blocks, when i is size - 1, we'll try to compare it to the height at 'size' element which is out of the vector. we can eliminate many lines from the loop and handle that case with this reordering:
```
const CPubKey& GetAggPubkeyFromHeight(uint height) const 
{
    if(height < 1 || aggregatePubkeyHeight.size() == 1) {
        return aggregatePubkeyHeight.at(0).aggpubkey; 

    if(height > aggregatePubkeyHeight.end().height)
        return aggregatePubkeyHeight.end().aggpubkey;

    for(int i = 0; i < aggregatePubkeyHeight.size(); i++) {
        if(height == aggregatePubkeyHeight.at(i).height)
            return aggregatePubkeyHeight.at(i).aggpubkey;

        else if(aggregatePubkeyHeight.at(i).height < height) && (height < aggregatePubkeyHeight.at(i+1).height)) 
            return aggregatePubkeyHeight.at(i).aggpubkey;
    }
}
```

src/federationparams.h
```diff
@@ -35,14 +40,42 @@ class CFederationParams
...
     /** Return the list of hostnames to look up for DNS seeds */
     const std::vector<std::string>& DNSSeeds() const { return vSeeds; }
+    const uint& GetHeightFromAggregatePubkey(std::vector<unsigned char> aggpubkey) const {
+        for (auto& c : aggregatePubkeyHeight)
+            if (c.aggpubkey == CPubKey(aggpubkey.begin(), aggpubkey.end())) {
+                return c.height;
+                break;
+            } else {
+                continue;
+            }
```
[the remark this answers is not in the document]
Naviabheeman: We need a default case when aggpubkey is not in the list.

src/federationparams.h
```diff
@@ -35,14 +40,42 @@ class CFederationParams
...
-    CPubKey ReadAggregatePubkey(const std::vector<unsigned char>& pubkey);
-    const CPubKey& GetAggregatePubkey() const{ return aggregatePubkey; }
+    CPubKey ReadAggregatePubkey(const std::vector<unsigned char>& pubkey, uint height);
+    const CPubKey& GetLatestAggregatePubkey() const { return aggregatePubkeyHeight.back().aggpubkey; }
     bool ReadGenesisBlock(std::string genesisHex);
     const CBlock& GenesisBlock() const { return genesis; }
     const std::string& getDataDir() const { return dataDir; }
     const std::vector<SeedSpec6>& FixedSeeds() const { return vFixedSeeds; }
     /** Return the list of hostnames to look up for DNS seeds */
     const std::vector<std::string>& DNSSeeds() const { return vSeeds; }
```
Naviabheeman, line 50: can we move the function definitions of GetHeightFromAggregatePubkey and GetAggPubkeyFromHeight to .cpp file keep only the declarations in the .h file?

test/functional/p2p_invalid_block_aggpubkey.py
```diff
@@ -0,0 +1,81 @@
...
+import copy
+
+from test_framework.blocktools import create_block, create_coinbase
+from test_framework.messages import COIN
+from test_framework.mininode import P2PDataStore
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal
+from test_framework.util import hex_str_to_bytes
+
+class InvalidBlockRequestTest(BitcoinTestFramework):
```
Naviabheeman, line 23: we can make this FederationBlockTest?

src/validation.cpp
```diff
@@ -4129,7 +4153,12 @@ bool LoadExternalBlockFile(FILE* fileIn, CDiskBlockPos *dbp)
                     while (range.first != range.second) {
                         std::multimap<uint256, CDiskBlockPos>::iterator it = range.first;
                         std::shared_ptr<CBlock> pblockrecursive = std::make_shared<CBlock>();
-                        if (ReadBlockFromDisk(*pblockrecursive, it->second))
+
+                        BlockMap::iterator miSelf = mapBlockIndex.find(hash);
+                        CBlockIndex *pindex = nullptr;
+                        pindex = miSelf->second;
```
Naviabheeman, line 4159: this code could be unsafe, `miSelf` maybe null when the block is not in `mapBlockIndex`. 

since we have a block pointer here, can we try to read height from the block itself, `pblockrecursive->vtx[0]->vin[0].prevout.n`

src/federationparams.cpp
```diff
@@ -139,21 +140,24 @@ CFederationParams::CFederationParams(const int networkId, const std::string data
...
     if(!pubkey.size())
         throw std::runtime_error("Aggregate Public Key for Signed Block is empty");
     
     if (pubkey[0] == 0x02 || pubkey[0] == 0x03) {
-        aggregatePubkey = CPubKey(pubkey.begin(), pubkey.end());
-        if(!aggregatePubkey.IsFullyValid()) {
+        aggPubkeyAndHeight p;
+        p.aggpubkey = CPubKey(pubkey.begin(), pubkey.end());
+        p.height = height;
+        aggregatePubkeyHeight.push_back(p);
```
Naviabheeman, line 152: We have to handle this call cautiously. 
I think we need to move this line:
`aggregatePubkeyHeight.push_back(p); `
do it only after the p.aggpubkey is validated. Otherwise the list is going to have an invalid aggpubkey.

If `p.aggpubkey.IsFullyValid() ` fails while loading genesis block its ok to throw std::runtime_error which stops the node. we cannot start the node without correct genesis block. When we are trying to setup the network this is ok.

But when the network is online and a new federation block is being processed, if it has invalid aggpubkey we should not throw. we have to reject the block and continue. otherwise the node will stop and could bring the whole network down!

src/validation.cpp
```diff
@@ -2867,7 +2870,15 @@ static bool CheckBlockHeader(const CBlockHeader& block, CValidationState& state,
     if(!proofSize)
         return state.Error("No proof in block");
 
-    const CPubKey& aggregatePubkey = FederationParams().GetAggregatePubkey();
+    int height = 0;
+    if(chainActive.Tip())
+        height = chainActive.Tip()->nHeight;
+
+     CPubKey aggregatePubkey;
+    if(nHeight >= height)
```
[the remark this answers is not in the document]
winnie-chaintope: u mean remove the if else and only use 
FederationParams().GetAggPubkeyFromHeight(nHeight); ?
when create block coinbase it will reach the else and take
FederationParams().GetLatestAggregatePubkey();

src/federationparams.h
winnie-chaintope: in your for loop look like both the if and else if return the same thing 
so i group them together

src/federationparams.cpp
winnie-chaintope: i will move it after
`if(!p.aggpubkey.IsFullyValid())` and
`if (p.aggpubkey.size() != CPubKey::COMPRESSED_PUBLIC_KEY_SIZE)`
before `return`

src/federationparams.h
```diff
@@ -52,7 +85,9 @@ class CFederationParams
     CMessageHeader::MessageStartChars pchMessageStart;
     std::string strNetworkID;
     std::string dataDir;
-    CPubKey aggregatePubkey;
+    std::vector<CPubKey> aggregatePubkey;
+    std::vector<aggPubkeyAndHeight> aggregatePubkeyHeight;
```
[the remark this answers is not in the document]
winnie-chaintope: how to eliminate the cast? i change to mutable but try to remove the case still giving me error

src/validation.cpp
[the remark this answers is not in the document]
Naviabheeman: you mean height 0 represents new block? then we can leave it as it is.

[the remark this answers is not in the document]
winnie-chaintope: yes ok

src/federationparams.h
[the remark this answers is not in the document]
Naviabheeman: That is because ReadAggregatePubkey is not const.
`CPubKey ReadAggregatePubkey(const std::vector<unsigned char>& pubkey, uint height)` **const**;

src/validation.cpp
```diff
@@ -3093,8 +3107,11 @@ bool CChainState::AcceptBlockHeader(const CBlockHeader& block, CValidationState&
                 return state.Invalid(error("%s: block %s is marked invalid", __func__, hash.ToString()), 0, "duplicate");
             return true;
         }
+        uint height = 0;
+        if(pindex)
+           height = pindex->nHeight;
 
-        if (!CheckBlockHeader(block, state))
+        if (!CheckBlockHeader(block, state, height))
```
Naviabheeman, line 3115: Height is always 0 here? then we can just leave out the parameter

src/federationparams.h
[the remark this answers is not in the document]
winnie-chaintope: throw error when aggpubkey not in the list?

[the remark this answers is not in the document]
winnie-chaintope: `uint& CFederationParams::GetHeightFromAggregatePubkey(std::vector<unsigned char> aggpubkey) const
{
    for (auto& c : aggregatePubkeyHeight) {
        if (c.aggpubkey == CPubKey(aggpubkey.begin(), aggpubkey.end()))
            return c.height;
    }
    throw std::runtime_error("GetHeightFromAggregatePubkey: aggpubkey not found");
}`

i change it to throw error so it will not process after it gives error ok?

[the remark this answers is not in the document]
Naviabheeman: no throw runtime_error will stop Tapyrusd. we should just return a default value. I think for this we can return -1 so that we know its not in the list. but to return we have to change the return type to "int" from "uint"

src/federationparams.cpp
```diff
@@ -139,21 +140,27 @@ CFederationParams::CFederationParams(const int networkId, const std::string data
...
             throw std::runtime_error(strprintf("Aggregate Public Key for Signed Block is invalid: %s", HexStr(pubkey)));
         }
 
-        if (aggregatePubkey.size() != CPubKey::COMPRESSED_PUBLIC_KEY_SIZE) {
+        if (p.aggpubkey.size() != CPubKey::COMPRESSED_PUBLIC_KEY_SIZE) {
             throw std::runtime_error(strprintf("Aggregate Public Key for Signed Block is invalid: %s", HexStr(pubkey)));
         }
-        return aggregatePubkey;
+
+        if((aggregatePubkeyHeight.size() > 0 && GetHeightFromAggregatePubkey(pubkey) == -1) || (aggregatePubkeyHeight.size() == 0))
```
Naviabheeman, line 160: I don't understand why this check is added

winnie-chaintope: i realised the readAggPubKey will set duplicate 
so i added this line to prevent duplication

Naviabheeman: ok, I got it, ReadAggregatePubkey is called twice for every new block, once in ProcessNewBlock and once in ConnectTip. so the same aggpubkey is push twice. I think we can remove the call from ProcessNewBlock.

PR #88, opened 2020-05-27
getTokenBalance

winnie-chaintope:


src/wallet/test/wallet_tests.cpp
```diff
@@ -379,4 +379,34 @@ BOOST_FIXTURE_TEST_CASE(generate_with_incorrect_privkey, TestingSetup)
...
+    {
+        CWallet wallet("dummy", WalletDatabase::CreateDummy());
+        AddKey(wallet, coinbaseKey);
+        WalletRescanReserver reserver(&wallet);
+        reserver.reserve();
+        CreateAndProcessBlock({}, CP2PKHScriptPubKey);
+
+        std::vector<CBalance> walletBalance = wallet.GetAvailableTokenBalance();
+        // printf("wallet size after 1: %u\n", walletBalance.size());
+
```
azuchi, line 407: Isn't it necessary to check if the result of `wallet.GetAvailableTokenBalance()` is what you expected?

src/test/script_tests.cpp
```diff
@@ -2471,5 +2471,27 @@ BOOST_AUTO_TEST_CASE(bitcoinconsensus_verify_script_invalid_flags)
...
+
+    // CP2SH(Colored P2SH)：
+    // <COLOR identifier> OP_COLOR OP_HASH160 <H(redeem script)> OP_EQUAL
+    CScript CP2SHScriptPubKey = CScript() << ParseHex("011863143c14c5166804bd19203356da136c985678cd4d27a1b8c6329604903262") << OP_COLOR << OP_HASH160 << ParseHex("da8a647bba351bbae4cee0089d373c97ec240580") << OP_EQUAL;
+
+    //check CP2SHScriptPubKey match ColoredPayToScriptHash not match ColoredPayToPubkeyHash
+    // BOOST_CHECK(!MatchColoredPayToPubkeyHash(CP2SHScriptPubKey, data, colorId));
+    BOOST_CHECK(CP2SHScriptPubKey.IsColoredPayToScriptHash());
+}
+
```
azuchi, line 2495: In order to check the conditions comprehensively, you need to test the following cases

* Check if all token types are valid
* Check when an undefined token type is specified.
* Check for cases with different data lengths.

src/wallet/test/wallet_tests.cpp
winnie-chaintope: ya this one i havent completed yet

src/test/script_tests.cpp
winnie-chaintope: ok

src/wallet/test/wallet_tests.cpp
azuchi: Do you add a commit to this pull request to complete it?

winnie-chaintope: i will add a commit later

winnie-chaintope: azuchi san 
i temporary commented this unit test 
because found out there is one dependency not yet make changes
GetScriptForDestination might affect alot of places
so i will create another branch working on that

PR #92, opened 2020-06-04
Issue/77 fixed size color identifier

Naviabheeman:
Changed ColorIdentifier to fixed size. payload is 32 bytes hash of scriptpubkey or utxo.
Updated tests.


src/test/coloridentifier_tests.cpp
```diff
@@ -0,0 +1,152 @@
...
+    BOOST_CHECK_EQUAL(TokenToUint(c0.type), TokenToUint(TokenTypes::NONE));
+    BOOST_CHECK(memcmp(&c0.payload[0], &str[0], 32) == 0);
+
+    try {
+        CDataStream ss00(ParseHex("0100"), SER_NETWORK, INIT_PROTO_VERSION);
+        ss00 >> c0;
+        BOOST_CHECK_MESSAGE(false, "We should have thrown");
+    } catch (const std::ios_base::failure& e) {
+    }
+
```
azuchi, line 31: Can you add a test for 33 bytes of data starting with an invalid color type(0x04)?

PR #95, opened 2020-06-16
Fixed failing functional test and crash on macOS

Naviabheeman:
Fixed failing functional test.
Fixed tapyrusd startup failure on macOS.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1046,9 +1046,6 @@ static bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, 
         if(tpcin <= 0)
             return state.Invalid(false, REJECT_INSUFFICIENTFEE, "bad-txns-token-without-fee");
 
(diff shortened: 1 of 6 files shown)


src/validation.cpp
```diff
@@ -1046,9 +1046,6 @@ static bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state,
         if(tpcin <= 0)
             return state.Invalid(false, REJECT_INSUFFICIENTFEE, "bad-txns-token-without-fee");
 
-        if(tpcin - tpcout <= 0)
-            return state.Invalid(false, REJECT_INSUFFICIENTFEE, "bad-txns-token-insufficient");
-
```
azuchi, line 1051: Isn't this check necessary?

Naviabheeman: This is done in CheckTxInputs (and cause 'bad-txns-in-belowout' error). when I keep this check some test cases were failing because of the difference in error code.

PR #96, opened 2020-06-17
added colorId argument

winnie-chaintope:
isMine(), ExtractDestination(), GetScriptForDestination() functions added colorId params to get colored scripts


src/key_io.cpp
```diff
@@ -68,11 +68,18 @@ class DestinationEncoder : public boost::static_visitor<std::string>
...
 };
 
-CTxDestination DecodeDestination(const std::string& str, const CChainParams& params)
+CTxDestination DecodeDestination(const std::string& str, const CChainParams& params, ColorIdentifier* colorId = nullptr)
 {
     std::vector<unsigned char> data;
     uint160 hash;
     if (DecodeBase58Check(str, data)) {
+
+        if (data[35] == OP_COLOR) {
```
azuchi, line 77: It is better to use the existing decision logic for colored coin like `CScript::IsColoredScript()`, as it is better to have a single logic to determine them.

src/script/descriptor.cpp
```diff
@@ -322,6 +322,8 @@ class ConvertorDescriptor : public Descriptor
 };
 
 CScript ConvertP2SH(const CScript& script) { return GetScriptForDestination(CScriptID(script)); }
+ColorIdentifier colorID = ColorIdentifier();
+CScript ConvertCP2SH(const CScript& script) { return GetScriptForDestination(CScriptID(script), &colorID); }
```
azuchi, line 326: `ConvertCP2SH` does not appear to be used.

src/keystore.cpp
```diff
@@ -29,7 +29,12 @@ void CBasicKeyStore::ImplicitlyLearnRelatedKeyScripts(const CPubKey& pubkey)
 #ifdef DEBUG
         script = GetScriptForDestination(WitnessV0KeyHash(key_id));
 #else
-        script = GetScriptForDestination(key_id);
+        if (isColored) {
+            ColorIdentifier colorID = ColorIdentifier();
+            script = GetScriptForDestination(key_id, &colorID);
```
azuchi, line 34: In this case, the type of `colorID` is `TokenTypes::NONE`?

src/rpc/mining.cpp
```diff
@@ -106,13 +106,14 @@ static UniValue generatetoaddress(const JSONRPCRequest& request)
 
     int nGenerate = request.params[0].get_int();
 
-    CTxDestination destination = DecodeDestination(request.params[1].get_str());
+    ColorIdentifier* colorId = nullptr;
+    CTxDestination destination = DecodeDestination(request.params[1].get_str(), colorId);
```
azuchi, line 110: Since the following definitions are in `key_io.h`, is it okay to keep the original code?

```cpp
CTxDestination DecodeDestination(const std::string& str, ColorIdentifier* colorId = nullptr);
```

other code also same.

winnie-chaintope: actually the colorId will be use at line 116 GetScriptForDestination function
`coinbaseScript->reserveScript = GetScriptForDestination(destination, colorId);`

doc/release-notes.md
```diff
@@ -487,7 +487,7 @@ Support for Python 2 has been discontinued for all test files and tools.
 - #13063 `6378eef` Use shared pointer to retain wallet instance (promag)
 - #13142 `56fe3dc` Separate IsMine from solvability (sipa)
 - #13194 `fd96d54` Remove template matching and pseudo opcodes (sipa)
-- #13252 `c4cc8d9` Refactor ReserveKeyFromKeyPool for safety (Empact)
+- #13252 `c4cc8d9` Refactor  for safety (Empact)
```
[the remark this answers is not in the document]
winnie-chaintope: ok sorry i will change back
i think i accidentally remove the word it without realising

src/key_io.cpp
winnie-chaintope: ok

src/rpc/mining.cpp
azuchi: Oh, Isee.

src/keystore.cpp
winnie-chaintope: because we will not know the colorId during add key so set to a default colorId

src/script/descriptor.cpp
winnie-chaintope: i will remove this code since it is not being used

src/key_io.cpp
```diff
@@ -68,11 +68,18 @@ class DestinationEncoder : public boost::static_visitor<std::string>
...
+CTxDestination DecodeDestination(const std::string& str, const CChainParams& params, ColorIdentifier* colorId = nullptr)
 {
     std::vector<unsigned char> data;
     uint160 hash;
     if (DecodeBase58Check(str, data)) {
+
+        CScript scriptPubKey(data.begin(), data.end());
+        if(scriptPubKey.IsColoredScript()) {
+            ColorIdentifier cid(GetColorIdFromScript(scriptPubKey));
+            colorId = &cid;
```
Naviabheeman, line 80: this is not safe. the pointer is uninitialised and we are assigning the address of a stack variable. make it a reference,
```ColorIdentifier& colorId = ColorIdentifier()```

then by default its type will be TokenType::NONE

File: src/keystore.cpp [hunk 7]
```diff
@@ -53,11 +58,11 @@ bool CBasicKeyStore::GetPubKey(const CKeyID &address, CPubKey &vchPubKeyOut) con
     return true;
 }
 
-bool CBasicKeyStore::AddKeyPubKey(const CKey& key, const CPubKey &pubkey)
+bool CBasicKeyStore::AddKeyPubKey(const CKey& key, const CPubKey &pubkey, bool* isColored)
 {
     LOCK(cs_KeyStore);
     mapKeys[pubkey.GetID()] = key;
-    ImplicitlyLearnRelatedKeyScripts(pubkey);
+    ImplicitlyLearnRelatedKeyScripts(pubkey, isColored);
```
Naviabheeman, line 65: isColored is not initialised. None of AddKeyPubKey calls have the third parameter, its a pointer, we are accessing it here without checking if it is initialised. it would crash. 
I think we can remove the flag completely and just add two scripts 
```
ImplicitlyLearnRelatedKeyScripts(pubkey);
ImplicitlyLearnRelatedKeyScripts(pubkey, true);

Comment (Naviabheeman) [reply src/keystore.cpp:65]: Actually we don't need two calls to ImplicitlyLearnRelatedKeyScripts. since OP_COLOR is not allowed inside the redeem script it will be the same for P2SH and CP2SH. we need not change ImplicitlyLearnRelatedKeyScripts or AddKeyPubKey. Sorry, I didn't realise earlier.

Comment (winnie-chaintope) [reply src/keystore.cpp:65]: ok

File: src/script/standard.cpp [hunk 8]
```diff
@@ -345,15 +355,24 @@ class CScriptVisitor : public boost::static_visitor<bool>
...
     }
 
     bool operator()(const CScriptID &scriptID) const {
         script->clear();
-        *script << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+        if (colorID != NULL) {
+            *script << colorID->toVector() << OP_COLOR << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+        } else {
+            *script << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+        }
```
Naviabheeman, line 372: as we don't need OP_COLOR inside redeem script CScriptID need not be changed.

PR #99, opened 2020-06-25
Change prefix of ColorIdentifier types

Naviabheeman:
Changed ColorIdentifier Type from  '0x01' to '0xc1', '0x02' to '0xc2', '0x03' to '0xc3'.
Added check against 0 valued token outputs. Since IsDust is not checked for tokens 0 values are being allowed.


src/script/script.cpp
```diff
@@ -229,7 +229,7 @@ bool CScript::IsColoredPayToScriptHash() const
     // <COLOR identifier> OP_COLOR OP_HASH160 <H(redeem script)> OP_EQUAL
     if(this->size() == 58) // <COLOR identifier> : TYPE = 1 byte and 32 byte PAYLOAD
         return ((*this)[0] == 0x21 &&
-                ((*this)[1] == 0x01 || (*this)[1] == 0x02 || (*this)[1] == 0x03)&&
+                ((*this)[1] == 0xc1 || (*this)[1] == 0xc2 || (*this)[1] == 0xc3)&&
```
azuchi, line 232: It is better to use constant like `TokenTypes::REISSUABLE`.

PR #109, opened 2020-11-13
Fix color id size check in colored custom script

rantan:
The length of color ids is fixed 33 bytes. But the `MatchCustomColoredScript()` expected length 34bytes.

I guess the condition didn't consider that results of `std::distance()` include last item too.

### DIFF
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -410,7 +410,7 @@ bool MatchCustomColoredScript(const CScript& script, std::vector<unsigned char>&
     if(iterOpColor == script.end())
         return false;
 
-    if(iterColorId1 != script.end() && std::distance(iterColorId1 + 1, iterOpColor - 1) == 33 &&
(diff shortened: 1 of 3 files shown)


src/script/script.cpp
```diff
@@ -410,7 +410,7 @@ bool MatchCustomColoredScript(const CScript& script, std::vector<unsigned char>&
     if(iterOpColor == script.end())
         return false;
 
-    if(iterColorId1 != script.end() && std::distance(iterColorId1, iterOpColor) == 34 &&
+    if(iterColorId1 != script.end() && std::distance(iterColorId1 + 1, iterOpColor - 1) == 33 &&
```
azuchi, line 413: length `33` should be a constant.

azuchi (changes requested):

src/script/script.cpp
rantan: Done at c7ccadd

azuchi (approved):

PR #111, opened 2020-12-02
Make it possible to build tapyrus-qt on CMake

rantan:


azuchi (commented):
It is prefer to `BUILD_BITCOIN_QT` in `src/CMakeLists.txt` is also renamed to `BUILD_TAPYRUS_QT`.

rantan:
This can build GUI on CMake from clean environment now.
Now I'm adding CI settings for the CMake building. After this I'm going to back this to be reviewed.

PR #121, opened 2021-02-26
Unify private key format that is used in interfaces in WIF 

### LINKED ISSUE #120: Unify private key format as WIF in places where treats private key
The current situation is the cause of confusion for users who want to start up a new network.

* `generatetoaddress` RPC accepts a hex private key, but `tapyrus-genesis` command accepts a WIF private key.
* The help document of `tapyrus-genesis` command doesn't show details of `-signblockprivatekey`. Actually, the description is placed in the code but it is set as a hidden option.
* Other places where are before forked uses WIF format. but they don't show to use WIF explicitly.

So, I think we should do below.

* Unify the format of private keys as WIF.
* Make the `-signblockprivatekey` help document is shown.
* Make `importprivkey` RPC and `tapyrus-tx` command show to use WIF explicitly.

rantan:
This PR fixes #120. see it before review.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -117,7 +117,7 @@ static UniValue generatetoaddress(const JSONRPCRequest& request)
 
     CKey cPrivKey = DecodeSecret(request.params[2].get_str());
     if(!cPrivKey.IsValid())
-        throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "No private key given or invalid private key.");
+        throw JSONRPCError(RPC_WALLET_INVALID_PRIVATE_KEY, "No private key given or invalid private key.");
 
     if(cPrivKey.GetPubKey() != FederationParams().GetLatestAggregatePubkey())
         throw JSONRPCError(RPC_WALLET_INVALID_AGGREGATE_KEY, "Given private key doesn't correspond to the Aggregate Key.");
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -96,12 +96,12 @@ static UniValue generatetoaddress(const JSONRPCRequest& request)
             "\nArguments:\n"
             "1. nblocks      (numeric, required) How many blocks are generated immediately.\n"
             "2. address      (string, required) The address to send the newly generated bitcoin to.\n"
-            "3. private key  (hex string, required) aggregate private key to sign the generated blocks.\n"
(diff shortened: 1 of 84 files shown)


src/rpc/mining.cpp
```diff
@@ -115,10 +115,7 @@ static UniValue generatetoaddress(const JSONRPCRequest& request)
     std::shared_ptr<CReserveScript> coinbaseScript = std::make_shared<CReserveScript>();
     coinbaseScript->reserveScript = GetScriptForDestination(destination, colorId);
 
-    std::vector<unsigned char> privkeyraw = ParseHex(request.params[2].get_str());
-    CKey cPrivKey;
-    cPrivKey.Set(privkeyraw.begin(), privkeyraw.end(), true);
-
+    CKey cPrivKey = DecodeSecret(request.params[2].get_str());
     if(!cPrivKey.size())
```
azuchi, line 119: It is more intuitive to use `cPrivKey.IsValid()` to check the validity of the private key. In addition, it should be checked if the private key corresponds to `FederationParams().GetLatestAggregatePubkey()`.

src/wallet/rpcwallet.cpp
```diff
@@ -3952,21 +3952,18 @@ UniValue generate(const JSONRPCRequest& request)
...
     }
 
     int num_generate = request.params[0].get_int();
 
-    std::vector<unsigned char> privkeyraw = ParseHex(request.params[1].get_str());
-    CKey cPrivKey;
-    cPrivKey.Set(privkeyraw.begin(), privkeyraw.end(), true);
-
+    CKey cPrivKey = DecodeSecret(request.params[1].get_str());
     if(!cPrivKey.size())
```
azuchi, line 3967: It is more intuitive to use `cPrivKey.IsValid()` to check the validity of the private key. In addition, it should be checked if the private key corresponds to `FederationParams().GetLatestAggregatePubkey()`.

azuchi (changes requested):

src/rpc/mining.cpp
rantan: fixed at 37cd8f627f64269901a0954934ecaa7e614a6407

src/wallet/rpcwallet.cpp
rantan: fixed at 37cd8f627f64269901a0954934ecaa7e614a6407

azuchi (approved):

PR #127, opened 2021-04-18
Support to create colored coin transactions with correct change using Tapyrus wallet and report both TPC and coloured coin balances

Naviabheeman:
This PR adds support to maintain coloured coin token balances in Tapyrus wallet. 
It fixes the issue in coloured coin transaction creation where the fee was sometimes taken from token balance instead of TPC.
All RPC outputs that report Amount now specify a "token" stating if it is TPC/Colored coin
It enhances the functional test to send tokens using sendToAddress RPC and verify token/TPC balances in tapyrus wallet.


test/functional/feature_segwit.py
```diff
@@ -527,8 +527,6 @@ def mine_and_test_listunspent(self, script_list, ismine):
         elif (ismine == 1):
             assert_equal(watchcount, len(script_list))
             assert_equal(spendcount, 0)
-        else:
-            assert_equal(watchcount, 0)
         return txid
```
azuchi, line 530: What kind of case is 0?
If we can remove this, can we also remove the code that calls this method with 0?

src/wallet/rpcwallet.cpp
```diff
@@ -2430,16 +2436,31 @@ static UniValue gettransaction(const JSONRPCRequest& request)
...
+        CAmount nFee = (wtx.IsFromMe(filter) ? wtx.tx->GetValueOut(colorId) - nDebit : 0);
 
+        if(colorId.type == TokenTypes::NONE)
+            entry.pushKV("token", CURRENCY_UNIT);
+        else
+        {
+            assert(nFee == 0);
+            std::string cid(colorId.toString());
+            entry.pushKV("token", HexStr(cid.begin(), cid.end()));
+        }
```
azuchi, line 2459: You can simplify by creating a function that converts color-type to string.

azuchi (approved):

PR #128, opened 2021-05-13
Support using coloured coin addresses in tapyrus wallet and sendToAdress RPC

Naviabheeman:
Added colored address support in python
Added new types with colorid inside CTxDestination. and added 'token' name output in RPCs where the address has a colorid


test/functional/wallet_coloredcoin.py
```diff
@@ -0,0 +1,218 @@
...
+        walletinfo = self.nodes[1].getwalletinfo()
+        assert_equal(len(walletinfo['balance']), 2)
+        assert_equal(walletinfo['balance']['TPC'], 10)
+        assert_equal(walletinfo['balance'][bytes_to_hex_str(colorid1)], 0)
+
+        walletinfo = self.nodes[2].getwalletinfo()
+        assert_equal(len(walletinfo['balance']), 1)
+        assert_equal(walletinfo['balance']['TPC'], 101)
+
+        #  PART 1: using cp2pkh address
```
azuchi, line 126:
```suggestion
        #  PART 2 using cp2sh address
```

test/functional/wallet_coloredcoin.py
```diff
@@ -0,0 +1,218 @@
...
+        walletinfo = self.nodes[1].getwalletinfo()
+        assert_equal(len(walletinfo['balance']), 3)
+        assert_equal(walletinfo['balance']['TPC'], 20)
+        assert_equal(walletinfo['balance'][bytes_to_hex_str(colorid1)], 0)
+        assert_equal(walletinfo['balance'][bytes_to_hex_str(colorid2)], 0)
+
+        walletinfo = self.nodes[2].getwalletinfo()
+        assert_equal(len(walletinfo['balance']), 1)
+        assert_equal(walletinfo['balance']['TPC'], 152)'''
+
```
azuchi, line 216: Will these comment-out be fixed in subsequent PRs?

azuchi (changes requested):

test/functional/wallet_coloredcoin.py
Naviabheeman: yes, these will be uncommented in PR127 after the fee calculation is fixed. Now the balances are not correct.

src/wallet/rpcwallet.cpp
```diff
@@ -555,10 +555,11 @@ static UniValue sendtoaddress(const JSONRPCRequest& request)
...
     }
     ColorIdentifier colorId;
-    if(dest.which() == 3 || dest.which() == 4)
-    {
+    if(dest.which() == 3)
         colorId = boost::get<CColorKeyID>(dest).color;
-    }
+    else if(dest.which() == 4)
+        colorId = boost::get<CColorScriptID>(dest).color;
+
```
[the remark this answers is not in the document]
Naviabheeman: done

azuchi (approved):

PR #142, opened 2021-06-28
ReissueToken RPC and Detailed colored coin functional test

Naviabheeman:
Added implementation of reissue token RPC using two transaction, first to create a UTXO with the respective scriptpubkey and second to create a coloured coin spending the previous UTXO.
Changed issuetoken RPC to use the two transaction method for issuing REISSUABLE tokens to be consistent.
Improved functional test to include all address types and coloured coin types. All coloured coin RPCs are tested with all combinations


test/functional/wallet_coloredcoin.py
```diff
@@ -7,19 +7,23 @@
...
+        3. create transaction to issue/send/burn colored coins
     
     RPCs:
         1. sendtoaddress
         2. issuetoken
         3. transfertoken
         4. burntoken
         5. getcolor
+        6. reissuetoken
+        createrawtransaction
```
azuchi, line 19:
```suggestion
        7. createrawtransaction
```

test/functional/wallet_coloredcoin.py
```diff
@@ -7,19 +7,23 @@
...
+        sendtoaddress
```
azuchi, line 20:
```suggestion
        8. sendtoaddress
```

azuchi (changes requested):

src/wallet/rpcwallet.cpp
```diff
@@ -32,6 +32,7 @@
 #include <wallet/wallet.h>
 #include <wallet/walletdb.h>
 #include <wallet/walletutil.h>
+#include <wallet/fees.h>
```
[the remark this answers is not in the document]
Naviabheeman: No, its not needed. I'll remove it.

Issue #141, opened 2021-06-18
Add a configuration parameter to allow multiple OP_RETURNs


azuchi:
In some cases, the Layer 2 protocol with metadata, requires multiple OP_RETURNs for a single transaction.
To support such cases, we would like to allow multiple OP_RETURNs in the configuration parameters.

Naviabheeman:
@azuchi, Multiple OP_RETURN meaning which of these: 
  transaction with more than one outputs having exactly one OP_RETURN in the script
  transaction with one output having more than one OP_RETURN in the script
  transaction with more than one output having more than one OP_RETURN in the script

Naviabheeman:
We now have `-datacarrier` and `-datacarriersize `parameters to control the max data size. we can add more options like: 

1. '-datacarriermultiple' - Allow multiple OP_RETURN. Default:False
2. `-datacarriercount ` - Max allowed OP_RETURN ops in one script. This is used only when datacarriermultiple is true. default:1
3. . `-datacarrieroutputcount`- Max allowed OP_RETURN outputs in one tx. This is used only when datacarriermultiple is true. default:1

Each OP_RETURN will still allow only `-datacarriersize` bytes of data.

azuchi:
@Naviabheeman I am assuming this case.

> transaction with one output having more than one OP_RETURN in the script

In debug builds we can bypass the standard rule check with the `-accept nonstdtxn` option. I think that making this option available, not just for debug builds, will allow us to handle multiple OP_RETURNs.

PR #144, opened 2021-07-02
New option  to allow multiple OP_RETURN in a tapyrus script

Naviabheeman:
(#141) New command line option -datacarriermultiple is added to allow tapyrus scripts to have multiple OP_RETURNs. its default is false.

### DIFF
--- a/src/script/standard.cpp
+++ b/src/script/standard.cpp
@@ -18,6 +18,7 @@
 typedef std::vector<unsigned char> valtype;
 
 bool fAcceptDatacarrier = DEFAULT_ACCEPT_DATACARRIER;
+bool fAcceptMultipleDatacarrier = DEFAULT_ACCEPT_MULTIPLE_DATACARRIER;
 unsigned nMaxDatacarrierBytes = MAX_OP_RETURN_RELAY;
 
 CScriptID::CScriptID(const CScript& in) : uint160(Hash160(in.begin(), in.end())) {}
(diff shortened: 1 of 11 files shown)


azuchi (changes requested):
`multiple OP_RETURN` means tx has multiple outputs that contains `OP_RETURN`, not multiple `OP_RETURN` in a script. multiple `OP_RETURN` in a script is not necessary.

src/script/standard.cpp
```diff
@@ -190,9 +191,13 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
...
     // script.
-    if (scriptPubKey.size() >= 1 && scriptPubKey[0] == OP_RETURN && scriptPubKey.IsPushOnly(scriptPubKey.begin()+1)) {
+    if (scriptPubKey.size() >= 1
+     && scriptPubKey[0] == OP_RETURN){
         typeRet = TX_NULL_DATA;
-        return true;
+        if(scriptPubKey.IsPushOnly(scriptPubKey.begin()+1))
+            return true;
+        else
+            return false;
```
azuchi, line 200: Why is this change needed?

Naviabheeman: without this change scripts with multiple `OP_RETURN` were **not** marked as `TX_NULL_DATA` and were allowed as standard by the default `TX_CUSTOM` case.

azuchi (approved):

PR #147, opened 2021-07-14
146 colored coin scale in rpc

Naviabheeman:
Changed amount outputs to include 'token' key value pair.
Amount is integer for coloured coins and tapyrus scale string for TPC


src/tapyrus-tx.cpp
```diff
@@ -32,6 +32,8 @@ static bool fCreateBlank;
 static std::map<std::string,UniValue> registers;
 static const int CONTINUE_EXECUTION=-1;
 
+const std::string CURRENCY_UNIT = "TPC";
```
azuchi, line 35: CURRENCY_UNIT is also defined in `feederate.cpp`. Similarly, in `tapyrusunits.cpp`. It would be good to define it in a common header file somewhere.

src/core_write.cpp
```diff
@@ -227,7 +227,9 @@ void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry,
 
         UniValue out(UniValue::VOBJ);
 
-        out.pushKV("value", ValueFromAmount(txout.nValue));
+        ColorIdentifier colorId(std::move(GetColorIdFromScript(txout.scriptPubKey)));
+        out.pushKV("token", colorId.toHexString());
```
azuchi, line 231: The description of the newly added `token` field should also be added to the help for each RPC.

azuchi (changes requested):

src/tapyrus-tx.cpp
Naviabheeman: CURRENCY_UNIT is defined in `feederate.cpp` and  'tapyrus-tx.cpp' for inclusion in 'tapyrusd' and 'tapyrus-tx'. As this is a string constant If I put it in a header, it would cause multiple definitions in each executable and linker error during compilation. I can add a new cpp file that can be included in both executables. but it is a lot more work just for one string.

azuchi (approved):

PR #153, opened 2021-08-10
RPC updates part -3: Removing witness support from wallet

Naviabheeman:
Removed addwitnessaddress RPC
Removed witnessifier from tapyrus wallet
Removed witness script support from tapyrus wallet and functional tests

Not in this PR:
Complete segwit support is not removed in this PR. Some more code remains in the peer to peer layer and script processing. These can be done after the wallet is completed


src/script/ismine.cpp
```diff
@@ -68,9 +68,6 @@ IsMineResult IsMineInner(const CKeyStore& keystore, const CScript& scriptPubKey,
     {
```
azuchi, line 68: `IsMineSigVersion` has `WITNESS_V0`.

```cpp
enum class IsMineSigVersion
{
    TOP = 0,        //! scriptPubKey execution
    P2SH = 1,       //! P2SH redeemScript
    WITNESS_V0 = 2  //! P2WSH witness script execution
};
```

src/script/sign.cpp
```diff
@@ -121,9 +121,6 @@ static bool SignStep(const SigningProvider& provider, const BaseSignatureCreator
     {
```
azuchi, line 121: It seems that there is still some processing related to scriptWitness.
Remove `witness_script` and `scriptWitness` from SignatureData in `script/sign.h` to clarify the target.

src/script/standard.cpp
```diff
@@ -438,20 +385,7 @@ CScript GetScriptForMultisig(int nRequired, const std::vector<CPubKey>& keys)
 
 CScript GetScriptForWitness(const CScript& redeemscript)
```
azuchi, line 386: Isn't this function itself unnecessary?

src/script/standard.h
```diff
@@ -127,24 +84,12 @@ struct WitnessUnknown
  *  A CTxDestination is the internal data type encoded in a bitcoin address
  */
```
azuchi, line 86: Remove following comments too.

```cpp
 *  * WitnessV0ScriptHash: TX_WITNESS_V0_SCRIPTHASH destination (P2WSH)
 *  * WitnessV0KeyHash: TX_WITNESS_V0_KEYHASH destination (P2WPKH)
 *  * WitnessUnknown: TX_WITNESS_UNKNOWN destination (P2W???)
 ```

File: src/wallet/wallet.cpp [hunk 5]
```diff
@@ -2798,7 +2759,7 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CTransac
                 bool fFirst = true;
 
                 // vouts to the payees
-                coin_selection_params.tx_noinputs_size = 11; // Static vsize overhead + outputs vsize. 4 nVersion, 4 nLocktime, 1 input count, 1 output count, 1 witness overhead (dummy, flag, stack size)
+                coin_selection_params.tx_noinputs_size = 10; // Static vsize overhead + outputs vsize. 4 nVersion, 4 nLocktime, 1 input count, 1 output count
```
azuchi, line 2762: What is the fix for this?

src/outputtype.cpp
```diff
@@ -48,14 +48,8 @@ std::vector<CTxDestination> GetAllDestinationsForKey(const CPubKey& key)
 {
```
azuchi, line 48: Following type is necessary?

```cpp
static const std::string OUTPUT_TYPE_STRING_P2SH_SEGWIT = "p2sh-segwit";
static const std::string OUTPUT_TYPE_STRING_BECH32 = "bech32";
```

azuchi (changes requested):

src/script/standard.cpp
Naviabheeman: no, it is scarcely used. I was able to remove it.

src/rpc/rawtransaction.cpp
```diff
@@ -1746,19 +1742,15 @@ UniValue createpsbt(const JSONRPCRequest& request)
```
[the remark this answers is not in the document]
Naviabheeman: I hadn't touched any  PSBT code. Just updated it and tests.

src/script/descriptor.cpp
```diff
@@ -533,25 +517,11 @@ std::unique_ptr<Descriptor> ParseScript(Span<const char>& sp, ParseScriptContext
         }
         return MakeUnique<MultisigDescriptor>(thres, std::move(providers));
     }
```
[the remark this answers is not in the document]
Naviabheeman: This is in descriptor feature. I think we need to review this separately, if we need to keep this feature it should be extended to support coloured coins. otherwise we can remove it. I'll create a separate issue for this?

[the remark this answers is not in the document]
azuchi: I see. it is better to create another issue.

azuchi (commented):
@Naviabheeman Since `vsize` and `weight` are specific to segwit, can't they be removed?

Naviabheeman:
> @Naviabheeman Since `vsize` and `weight` are specific to segwit, can't they be removed?

@azuchi, it can be removed. but I think it can be a separate PR. as some consensus rules like max transaction size now depend on transaction weight, witness scale factor etc. To remove them cleanly we need to need to check how it was calculated before segwit. As its not a small task I'll better create another issue and another PR.

azuchi (approved):

PR #154, opened 2021-08-23
148 rpc updates 4

Naviabheeman:
- Removed minconf parameter from all RPCs. It is replaced by the default value used in the respective scenario.
- Added color argument to getbalance
- Updated output the following RPCs to include colorId: 

> listreceivedbyaddress
> getreceivedbyaddress
> getaddressinfo


src/wallet/rpcwallet.cpp
```diff
@@ -746,13 +748,26 @@ static UniValue getreceivedbylabel(const JSONRPCRequest& request)
...
         }
     }
 
-    return ValueFromAmount(nAmount);
+    UniValue balances(UniValue::VOBJ);
+    for (auto const& wb : lable_balance) {
+        if (wb.first.type == TokenTypes::NONE)
+            balances.pushKV(CURRENCY_UNIT.c_str(), ValueFromAmount(wb.second));
+        else
+            balances.pushKV(HexStr(wb.first.toVector()).c_str(), wb.second);
```
azuchi, line 768: You can use `wb.first.toHexString` instead of `wb.first.toVector()` and `CURRENCY_UNIT`.

src/wallet/rpcwallet.cpp
```diff
@@ -1110,7 +1125,7 @@ static UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bo
                 continue;
 
             tallyitem& item = mapTally[address];
-            item.nAmount += txout.nValue;
+            item.nAmount+= txout.nValue;
```
azuchi, line 1128: this change unnecessary.

azuchi (changes requested):
Although not directly related to this PR, the following code on line 1778 is also `-nFee` for tokens, is this necessary?

```cpp
entry.pushKV("amount", (colorId.type == TokenTypes::NONE) ? ValueFromAmount(nNet - nFee) : nNet - nFee);
```

Naviabheeman:
> Although not directly related to this PR, the following code on line 1778 is also `-nFee` for tokens, is this necessary?
> 
> ```c++
> entry.pushKV("amount", (colorId.type == TokenTypes::NONE) ? ValueFromAmount(nNet - nFee) : nNet - nFee);
> ```

no. this was a mistake. I'll fix it.

PR #156, opened 2021-09-05
fixed bug in getnewaddress when colorID constructor throws exception

Naviabheeman:


azuchi (changes requested):
Addresses will be generated even for values larger than 33 bytes, which is not correct for color id.

example:  34 bytes argument
`c11863143c14c5166804bd19203356da136c985678cd4d27a1b8c632960490326212`

azuchi:
There are other codes that need to be checked, so it would be nice to have a Color ID validation function.

src/wallet/rpcwallet.cpp
```diff
@@ -150,6 +150,19 @@ static void addTokenKV(const CTxDestination& address, const CAmount nAmount, Uni
...
 
+static bool checkColorIdParam(const UniValue& param, ColorIdentifier& colorId) {
+    try {
+        const std::vector<unsigned char> vColorId(ParseHex(param.get_str()));
+        if(vColorId.size() != 33) return false;
+        colorId = ColorIdentifier(vColorId);
+        if(colorId.type == TokenTypes::NONE) return false;
+    }
+    catch(...) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid color parameter.");
```
azuchi, line 161: It should return false here, not an JSONRPCError.

src/wallet/rpcwallet.cpp
```diff
@@ -150,6 +150,20 @@ static void addTokenKV(const CTxDestination& address, const CAmount nAmount, Uni
...
 
+static bool checkColorIdParam(const UniValue& param, ColorIdentifier& colorId) {
+    try {
+        const std::vector<unsigned char> vColorId(ParseHex(param.get_str()));
+        if(vColorId.size() != 33) return false;
+        colorId = ColorIdentifier(vColorId);
+        if(colorId.type == TokenTypes::NONE) return false;
+    }
+    catch(...) {
+        colorId.type = TokenTypes::NONE;
```
azuchi, line 161: Why this expression need?

Naviabheeman: hmmm...its not needed. I was thinking that this would be useful after return... but its not effective.

PR #163, opened 2021-10-18
Add coloured coin transactions

Shurabthini:
coloured coin transactions have been added to blocks for test resurrection.


azuchi (commented):
The current test is testing the  colored Tx after the preceding test is finished and the mempool is empty.
I think it would be better to test with a mix of colored tx and non-colored Tx. Therefore, how about including the colored Tx in the preceding test?

@Naviabheeman  what do you think?

azuchi (changes requested):
The commit message explains what the commit does, which is important for other users who will later see the commit log. The `commit` doesn't tell anything, so fix it to an appropriate message.

PR #170, opened 2021-11-06
Fix string to integer conversion error in RPCs when using cli

Naviabheeman:
Add all new token RPCs to client table for conversion. 
Enable cli mode testing of wallet colored coin test


src/rpc/client.cpp
```diff
@@ -148,6 +148,14 @@ static const CRPCConvertParam vRPCConvertParams[] =
...
     { "createwallet", 1, "disable_private_keys"},
     { "testproposedblock", 1, "nonstandard"},
+    { "issuetoken", 0, "type"},
+    { "issuetoken", 1, "value"},
+    { "issuetoken", 3, "index"},
+    { "getcolor", 0, "type"},
+    { "getcolor", 2, "index"},
+    { "reissuetoken", 1, "value"},
+    { "burntoken", 1, "value"},
+    { "transfertoken", 1, "amount"},
```
azuchi, line 158: `issuetoken` and `burntoken` use `value`, not `amount`. Is there any reason to use `amount` only `transfertoken`?

Naviabheeman: transfertoken is only a wrapper around sendtoaddress. So I just used the same name as in send to address. We can change it to keep it uniform as well. please let me know your preference.

azuchi: I see. `sendtoaddress` take amount as string because it may be used to send TPC right? 
If sp, and `transfertoken` always take an integer, bacause it transfer only token(not TPC), `value` is better in order to know receive integer as well as others.

azuchi (approved):

Issue #174, opened 2021-11-28
Error while signing colored coin transaction


Naviabheeman:
signing colored coin transaction causes "Multiple OP_COLOR  found in script" error. It is a bug introduced by #173

PR #175, opened 2021-11-28
Fix multiple OP_COLOR in script error while signing

Naviabheeman:
bug #174 which was introduced by #173 changes is fixed in this PR. 

Functional tests are not included in this PR. I'll Create a separate PR for them
Edit(6 Dec)

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -843,8 +843,7 @@ UniValue SignTransaction(CMutableTransaction& mtx, const UniValue& prevTxsUnival
         }
         const CScript& prevPubKey = coin.out.scriptPubKey;
         const CAmount& amount = coin.out.nValue;
-        ColorIdentifier colorId = GetColorIdFromScript(coin.out.scriptPubKey);
-        inBalances[colorId] += amount;
+        inBalances[GetColorIdFromScript(coin.out.scriptPubKey)] += amount;
 
         SignatureData sigdata = DataFromTransaction(mtx, i, coin.out);
         // Only sign SIGHASH_SINGLE if there's a corresponding output:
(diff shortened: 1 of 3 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -854,15 +853,18 @@ UniValue SignTransaction(CMutableTransaction& mtx, const UniValue& prevTxsUnival
...
-        if (!VerifyScript(txin.scriptSig, prevPubKey, nullptr, STANDARD_SCRIPT_VERIFY_FLAGS, TransactionSignatureChecker(&txConst, i, amount), colorId, &serror)) {
+        if (!VerifyScript(txin.scriptSig, prevPubKey, nullptr, STANDARD_SCRIPT_VERIFY_FLAGS, TransactionSignatureChecker(&txConst, i, amount), tempColorId, &serror)) {
             if (serror == SCRIPT_ERR_INVALID_STACK_OPERATION) {
                 // Unable to sign input and verification failed (possible attempt to partially sign).
                 TxInErrorToJSON(txin, vErrors, "Unable to sign input, invalid stack size (possibly missing key)");
             } else {
                 TxInErrorToJSON(txin, vErrors, ScriptErrorString(serror));
             }
         }
+        }
```
azuchi, line 867: Do you need outer braces? Also can you fix indentation?

test/functional/mempool_reorg_.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
```
azuchi, line 1: What is the difference between this file and `mempool_reorg.py`?

azuchi (changes requested):

test/functional/mempool_reorg_.py
Naviabheeman: sorry! this is a mistake. it was local backup.

azuchi (approved):

Issue #191, opened 2022-04-27
Modify transaction validation rule for txs which issue colored coins


rantan:
[Here](https://github.com/chaintope/tapyrus-core/blob/master/src/validation.cpp#L1050) checks txs which issues colored coin have enough TPC amount in their inputs.

It intends to allow any positive value of TPC to issue new tokens. So it is enough to satisfy below condition

```cpp
 tpcin - tpcout - ::minRelayTxFee.GetFee(nSize) <= 0
```

However it is defferent from this. So the current implementation should be modify to the above condition.

Naviabheeman:
PR #192 fixes this

PR #192, opened 2022-04-27
Fixed minimum TPC required in token issue transactions

Naviabheeman:
Fixed issue #191 minimum TPC required in token issue transactions 
added new change output to functional test.

### DIFF
--- a/test/functional/feature_coloredcoin.py
+++ b/test/functional/feature_coloredcoin.py
@@ -369,6 +369,7 @@ class ColoredCoinTest(BitcoinTestFramework):
         txSuccess7 = CTransaction()
         txSuccess7.vin.append(CTxIn(COutPoint(coinbase_txs[5].malfixsha256, 0), b""))
         txSuccess7.vout.append(CTxOut(1000, script_reissuable1))
+        txSuccess7.vout.append(CTxOut(4999998000, change_script))
         sig_hash, err = SignatureHash(coinbase_txs[5].vout[0].scriptPubKey, txSuccess7, 0, SIGHASH_ALL)
         signature = self.coinbase_key.sign(sig_hash) + b'\x01'
         txSuccess7.vin[0].scriptSig = CScript([signature])
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1047,7 +1047,7 @@ static bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, 
             if(iter == inColoredCoinBalances.end())
             {
                 //if TPC input is sufficiently large this is a token issue. 
-                if(tpcin < 0 || tpcin - tpcout - ::minRelayTxFee.GetFee(nSize) < 1)
+                if(tpcin < 0 || tpcin - tpcout - ::minRelayTxFee.GetFee(nSize) <= 0)
                     return state.Invalid(false, REJECT_INSUFFICIENTFEE, "bad-txns-token-insufficient");
             }
             //output valus is more than input


test/functional/feature_coloredcoin.py
```diff
@@ -369,6 +369,7 @@ def run_test(self):
         txSuccess7 = CTransaction()
         txSuccess7.vin.append(CTxIn(COutPoint(coinbase_txs[5].malfixsha256, 0), b""))
         txSuccess7.vout.append(CTxOut(1000, script_reissuable1))
+        txSuccess7.vout.append(CTxOut(4999998000, change_script))
```
azuchi, line 372: The code before the change seems to pass this test.

test/functional/feature_coloredcoin.py
```diff
@@ -369,6 +369,7 @@ def run_test(self):
...
         sig_hash, err = SignatureHash(coinbase_txs[5].vout[0].scriptPubKey, txSuccess7, 0, SIGHASH_ALL)
         signature = self.coinbase_key.sign(sig_hash) + b'\x01'
         txSuccess7.vin[0].scriptSig = CScript([signature])
```
azuchi, line 375: There does not seem to be any test code where `"bad-txns-token-insufficient"` occurs, so it needs to be added.

azuchi (changes requested):

test/functional/feature_coloredcoin.py
Naviabheeman: Right, it is because there was no change output and the all the leftover TPC was taken as fee. I meant this change to check that TPC less than 1 to be accepted.

test/functional/feature_coloredcoin.py
Naviabheeman: To my surprise I can't simulate this error! it gets caught in the 'min relay fee not met' when the TPC input is low. 
So I changed the code to allow positive cases. negative cases  do not get to this point.

azuchi: @rantan Did not this error actually occur? or other errors?

rantan: I got an error `{"response_code":"500","response_msg":"Internal Server Error","rpc_error":{"code":-26,"message":"bad-txns-token-insufficient (code 66)"}}` from `sendrawtransaction` RPC if it pay fee that is same with min relay fee.

This is an example tx of the error.
`0100000003ea1500c01a327baf53221827cd0867db86953d7c7aae1cfa543d9dee36cad6ea000000006441f9f27b063b40c9bb123758e4bd3ad42cc269f48d8e54deb5427ba5827c5b0798fbecba09cfdff60bcc32cb3b73359082ec82c04ddf994c0a50b40127ed8e40ef0121033b098877fc2b1d24b1848eec9e09953cb4ea2b8d726b802b92d2a0fe2647caccffffffff15a93dba3473d1bf05936ec9451a78870dc4cfe9fafca31453d92811922430170100000064417a2662b95bbf75f21b225f51d09c5a1345693517a1f95b1d75553f1b44d77e986052efdea662b5b885cd1494fdf919bfc1ca72dce7be10f5ba8ea4f3958d937c01210288d109dd1cf995aa2eb6befd8afa351c05f84ac0b03e38ee8a83b46260a5fb80ffffffff15a93dba3473d1bf05936ec9451a78870dc4cfe9fafca31453d92811922430170e000000644139b362972f47e57a810ae296140afc15703ba7d2bb31edaadd28142119a68be8b1c8c4de4428ae57eab2e56a9bb5f7f7eb0acb13a097eba21284d33dd6b12f8e01210288d109dd1cf995aa2eb6befd8afa351c05f84ac0b03e38ee8a83b46260a5fb80ffffffff6564000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac64000000000000003c21c163347bf5418aa15d822e7128c0f564dbaa0d7d5b9bf3a2830e339696b8500fd3bc76a9147697e36bb9a38113c56cccbee0569be49d3cd5dd88ac490a0000000000001976a914b661961615d8973d33e53f2861c7ac48a863523588ac00000000`

The unit test that produce above tx is https://github.com/rantan/glueby/blob/fix_dummy_tx/spec/functional/token_spec.rb#L249
It try to produce 100 colored coin issuing outputs with minimum fee.
To reproduce kind of above tx, this row(https://github.com/rantan/glueby/blob/fix_dummy_tx/lib/glueby/contract/fee_estimator/auto.rb#L34) need to be modified to eliminate `+ 1` .

azuchi: @rantan thanks!
@Naviabheeman Please check based on the sample above.

Naviabheeman: @azuchi, yes, I'm able to simulate this error with the old code. After fixing the code as suggested by 191 it does not happen anymore.  I'll create a commit with the modified python test first.  Then add the c++ fix alone to make the test fail.

azuchi (changes requested):

azuchi:
> Now the test fails because "bad-txns-token-insufficient" does not occur.

@Naviabheeman I see. But I am curious, why is it that in some cases the test is passed by CI?

azuchi:
> That is why I would prefer to remove the error and allow token issue with this condition ` if(tpcin > tpcout + ::minRelayTxFee.GetFee(nSize)) continue;`

What is the meaning of continue here?
There is only this check in this if-branch, and whether or not the check matches and continues or not, it moves to the next loop.

Naviabheeman:
> 
> @Naviabheeman I see. But I am curious, why is it that in some cases the test is passed by CI?

The test did not have a check for "bad-txns-token-insufficient" in the beginning. After changing the condition I added transaction with min relay fee and one paying extra fee less than 1 TPC. so it was always passing. 

>What is the meaning of continue here?
>There is only this check in this if-branch, and whether or not the check matches and continues or not, it moves to the >next loop.

the continue is meant to allow the token issue transaction. Control comes to this case only when there is a token in output but the same token was not in the input. i.e token issue. Here we try to see if there is enough TPC to allow the token issue. "continue" will allow that output and move to the next output in the loop.

test/functional/feature_coloredcoin.py
```diff
@@ -432,6 +434,22 @@ def run_test(self):
...
         test_transaction_acceptance(node, TxFailure8, accepted=False, reason=b'invalid-colorid')
 
+        #TxSuccess9 - (UTXO-17) - issue 10 REISSUABLE1, change
+        txSuccess9 = CTransaction()
+        txSuccess9.vin.append(CTxIn(COutPoint(txSuccess7.malfixsha256, 1), b""))
+        txSuccess9.vout.append(CTxOut(10, script_reissuable1))
+        txSuccess9.vout.append(CTxOut(4999999649, change_script))
+        sig_hash, err = SignatureHash(txSuccess7.vout[1].scriptPubKey, txSuccess9, 1, SIGHASH_ALL)
+        signature = self.coinbase_key.sign(sig_hash) + b'\x01'
+        txSuccess9.vin[0].scriptSig = CScript([signature])
```
azuchi, line 444: @Naviabheeman Why are you setting your signature here?
It looks like the code after this is changing the output value and re-signing it.

Naviabheeman: the signature is needed to calculate the size of the transaction. We need to calculate the fee so that the amount can be set to 'min transaction fee - 2'. Only then the error is actually generated every time the test is run. When I make the amount an absolute integer the test fails intermittently  if the signature size is different.

azuchi: I see.

azuchi (approved):

PR #194, opened 2022-05-21
QT GUI support in tapyrus GUI

Naviabheeman:
This PR adds token/colored coin support in all the qt GUI dialogs:
1. In Overview page token balance can be viewed using prev and next buttons
2. In send coins dialog token would be shown based on the address given
3. In receive funds dialog token is chosen in the drop down and colorid is given in the new text box
4. Transaction page contains a column called token to show the colorid


azuchi (changes requested):
@Naviabheeman This PR contains other PR's commits, remove them.

Naviabheeman:
Closing this in favour of multiple Pas starting with 200, 201, 202

PR #200, opened 2022-06-14
Tapyrus Qt Mac compilation and updates

Naviabheeman:
Rename bitcoin to tapyrus in parts of QT code.
Make macOS compilation with boost libraries less ambiguous.


src/qt/README.md
```diff
@@ -60,7 +60,7 @@ Represents the view to a single wallet.
 
 * UI elements like BitcoinAmountField, which inherit from QWidget.
 * `bitcoinstrings.cpp`: automatically generated
-* `bitcoinunits.(h/cpp)`: TPC / mTPC / etc handling
+* `TapyrusUnits.(h/cpp)`: TPC / mTPC / etc handling
```
azuchi, line 63:
```suggestion
* `tapyrusunits.(h/cpp)`: TPC / mTPC / etc handling
```

src/qt/trafficgraphwidget.cpp
```diff
@@ -9,7 +9,7 @@
 #include <QPainter>
 #include <QColor>
 #include <QTimer>
-
+#include <QPainterPath>
```
azuchi, line 12: Is this `include` necessary?

azuchi (changes requested):

src/qt/trafficgraphwidget.cpp
Naviabheeman: It is not necessary now with QT5.9, but on my laptop with QT5.15.3 the build fails. It can be done later when we actually upgrade too.

azuchi (approved):

PR #201, opened 2022-06-14
Coloured coin support in tapyrus QT overview page

Naviabheeman:
- The overview page now contains a new box listing Tokens and their balances. When there are no colored coins in the wallet, this box is not visible.
- The transaction table page now lists all colored coin transactions also. New categories "Token Issue" and "Token Burn" are added. Token transfer is the same as "Send to" or "Payment to self" or "Received with" with a color id. The transaction tool tip and transaction details from the right click menu now show token output and TPC output. 
- `gettransaction` rpc output is also enhanced to show all the credits and debits to the wallet separately. functional tests are updated according to the output.

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -389,6 +389,7 @@ struct WalletBalances
 
     void prev()
     {
+        if (tokens.empty()) return;
         if(tokenIndex != tokens.begin())
             tokenIndex--;
         else
@@ -400,6 +401,7 @@ struct WalletBalances
 
     void next()
     {
+        if (tokens.empty()) return;
         if(tokenIndex != tokens.end())
         {
             tokenIndex++;
@@ -410,11 +412,13 @@ struct WalletBalances
 
     bool isToken()
     {
+        if (tokens.empty()) return false;
         return (*tokenIndex).type != TokenTypes::NONE;
     }
 
     std::string getTokenName()
     {
+        if (tokens.empty()) return "";
         return (*tokenIndex).toHexString();
     }
 };
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -200,11 +200,10 @@ public:
     virtual bool tryGetBalances(WalletBalances& balances, int& num_blocks) = 0;
 
     //! Get balance.
-    virtual CAmount getBalance() = 0;
+    virtual CAmount getBalance(ColorIdentifier colorId = ColorIdentifier()) = 0;
 
     //! Get available balance.
-    virtual CAmount getAvailableBalance(const CCoinControl& coin_control) = 0;
-    virtual CAmount getAvailableBalance(const CCoinControl& coin_control, const ColorIdentifier& colorId) = 0;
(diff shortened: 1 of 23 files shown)


test/functional/wallet_coloredcoin.py
```diff
@@ -390,15 +380,13 @@ def test_sendRPC(self, rpcname):
...
+                                {"txid": txid6, "category": "send"},
+                                {"token" : self.colorids[5],
                                 "amount": -1,
                                 "confirmations": 0})
             assert_array_result(self.nodes[1].listtransactions(),
-                                {"txid": txid6},
-                                {"category": "receive",
-                                "token" : self.colorids[5],
+                                {"txid": txid6, "category": "receive"},
+                                {"token" : self.colorids[5],
```
azuchi, line 389: With this change, the following tests should be added:

```python
# Verify that multiple entries exist for the same txid after sending a token
txlist = self.nodes[0].listtransactions(count=50)
entries = [e for e in txlist if e['txid'] == txid1]
assert len(entries) >= 2, "Token tx should have multiple entries (send + fee)"

# Verify the existence of a fee entry
assert_array_result(txlist,
{"txid": txid1, "category": "fee"},
{"amount": lambda x: x < 0}) # The fee is a negative value
```

src/interfaces/wallet.h
```diff
@@ -368,6 +370,53 @@ struct WalletBalances
...
+        for(auto pair:watch_only_balances)
+            tokens.insert(pair.first);
+        for(auto pair:unconfirmed_watch_only_balances)
+            tokens.insert(pair.first);
+
+        tokenIndex = tokens.begin();
+    }
+
+    void prev()
+    {
```
azuchi, line 391: Add guard:

```suggestion
    {
        if (tokens.empty()) return;
```

src/interfaces/wallet.h
```diff
@@ -368,6 +370,53 @@ struct WalletBalances
...
+            tokenIndex--;
+        else
+        {
+            tokenIndex = tokens.end();
+            tokenIndex--;
+        }
+    }
+
+    void next()
+    {
```
azuchi, line 403: Add guard:


```suggestion
    {
        if (tokens.empty()) return;
```

src/interfaces/wallet.h
```diff
@@ -368,6 +370,53 @@ struct WalletBalances
...
+        if(tokenIndex != tokens.end())
+        {
+            tokenIndex++;
+            if(tokenIndex == tokens.end())
+                tokenIndex = tokens.begin();
+        }
+    }
+
+    bool isToken()
+    {
```
azuchi, line 414: Add guard:


```suggestion
    {
        if (tokens.empty()) return false;
```

src/interfaces/wallet.h
```diff
@@ -368,6 +370,53 @@ struct WalletBalances
...
+        return (*tokenIndex).type != TokenTypes::NONE;
+    }
+
+    std::string getTokenName()
+    {
```
azuchi, line 420: Add guard:


```suggestion
    {
        if (tokens.empty()) return "";
```

azuchi (changes requested):

azuchi (approved):

PR #209, opened 2022-06-27
Restrict Tapyrus Network id to Four bytes

Naviabheeman:
Added range check to networkId argument to tapyrusd. 
Added unit test for the same

### DIFF
--- a/src/federationparams.cpp
+++ b/src/federationparams.cpp
@@ -102,7 +102,11 @@ std::unique_ptr<CFederationParams> CreateFederationParams(const TAPYRUS_OP_MODE 
 {
     gArgs.SelectConfigNetwork(TAPYRUS_MODES::GetChainName(mode));
 
-    const int networkId = gArgs.GetArg("-networkid", TAPYRUS_MODES::GetDefaultNetworkId(mode));
+    int64_t nid;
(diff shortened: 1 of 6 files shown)


src/federationparams.cpp
```diff
@@ -102,7 +102,11 @@ std::unique_ptr<CFederationParams> CreateFederationParams(const TAPYRUS_OP_MODE
 {
     gArgs.SelectConfigNetwork(TAPYRUS_MODES::GetChainName(mode));
 
-    const int networkId = gArgs.GetArg("-networkid", TAPYRUS_MODES::GetDefaultNetworkId(mode));
+    int64_t nid;
+    bool inRange = gArgs.IsGetArgInRange("-networkid", 1, 4294967295,  TAPYRUS_MODES::GetDefaultNetworkId(mode), nid);
```
azuchi, line 106: `UINT_MAX` can be used instead of `4294967295`?

azuchi (changes requested):

src/federationparams.cpp
Naviabheeman: oh yes!

azuchi (approved):

PR #213, opened 2022-07-12
change p2p port

nakajo2011:
Dockerfile define `EXPOSE 2357 12383 2377 12381`, but this configuration was setuped `12383` as p2p port. Should change to same port.


azuchi:
The port in dev mode is `12383`.

PR #218, opened 2022-07-21
Sample tapyrus config file

Naviabheeman:
Mandatory tapyrus options are added to the file. 
Few others may be useful in configuring are added.

### DIFF
--- a/share/examples/tapyrus.conf
+++ b/share/examples/tapyrus.conf
@@ -140,7 +140,7 @@
 #paytxfee=0.00
 
 # A fee rate (in %s/kB) that will be used when fee estimation has insufficient data. It is recommended to have a fallback fee rate in every node especially in new networks.
-#fallbackfee=
+#fallbackfee=0.00001
 
 # Enable pruning to reduce storage requirements by deleting old blocks. 
 # This mode is incompatible with -txindex and -rescan.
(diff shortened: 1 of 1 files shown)


azuchi (changes requested):
The `fallbackfee` option should be set, especially when configuring a custom network in Tapyrus. Without this value, and without enough transactions, a transaction with 0 fee will be created, which will not be relayed.

share/examples/tapyrus.conf
```diff
@@ -126,22 +134,57 @@
...
 #keypool=100
 
-# Pay an optional transaction fee every time you send bitcoins.  Transactions with fees
+# Pay an optional transaction fee every time you send transations.  Transactions with fees
 # are more likely than free transactions to be included in generated blocks, so may
 # be validated sooner.
 #paytxfee=0.00
 
+# A fee rate (in %s/kB) that will be used when fee estimation has insufficient data. It is recommended to have a fallback fee rate in every node especially in new networks.
+#fallbackfee=
```
azuchi, line 143:
```suggestion
#fallbackfee=0.00001
```

azuchi (changes requested):

azuchi (approved):

PR #220, opened 2022-07-27
generatekey in Tapyrus-genesis

Naviabheeman:
Add "`generatekey`" option to Tapyrus-genesis in dev mode

It creates a private key and its pub key. outputs to the command line
`private key:cRzj2yQHHdzwEe78dmcztbZd71YoiR5NgajASSHWsc7nyuzPME5M   `
`public key:02eac86bd83c9f042f56a077f3df719b515f961b597e2c6acc0f37543949a09f8b`

A new key is generated everytime the command executes.


src/tapyrus-genesis.cpp
```diff
@@ -87,8 +87,63 @@ static int AppInit(int argc, char* argv[])
...
+    CKey secret;
+    secret.MakeNewKey(true);//compressed
+
+    //generate its public key
+    CPubKey pubkey = secret.GetPubKey();
+    assert(pubkey.IsFullyValid());
+    assert(secret.VerifyPubKey(pubkey));
+
+    fprintf(stdout, "private key:%s\n", EncodeSecret(secret).c_str());
+    fprintf(stdout, "public key:%s\n", HexStr(pubkey.begin(), pubkey.end()).c_str());
```
azuchi, line 102:
```suggestion
    fprintf(stdout, "public key: %s\n", HexStr(pubkey.begin(), pubkey.end()).c_str());
```

src/tapyrus-genesis.cpp
```diff
@@ -87,8 +87,63 @@ static int AppInit(int argc, char* argv[])
...
+    //generate a secret key
+    CKey secret;
+    secret.MakeNewKey(true);//compressed
+
+    //generate its public key
+    CPubKey pubkey = secret.GetPubKey();
+    assert(pubkey.IsFullyValid());
+    assert(secret.VerifyPubKey(pubkey));
+
+    fprintf(stdout, "private key:%s\n", EncodeSecret(secret).c_str());
```
azuchi, line 101:
```suggestion
    fprintf(stdout, "private key: %s\n", EncodeSecret(secret).c_str());
```

It is more convenient to copy if there is space.

azuchi (changes requested):
`doc/man/tapyrus-genesis.1` also needs to be updated.

azuchi (approved):

PR #223, opened 2022-08-06
Fix segmentation fault in issue token

Naviabheeman:
Issue Reissuable token checks the script for validity. but the checks were incorrect. These have been fixed.
Issue non-reissuable and NFT did not have checks for the outpoint given in input. these have been added.
Functional tests have been added with the new error scenarios.


Naviabheeman:
closing as there are incorrect commit in this

PR #224, opened 2022-08-06
Fix segmentation fault in issuetoken

Naviabheeman:
Issue Reissuable token checks the script for validity. but the checks were incorrect. These have been fixed.
Issue non-reissuable and NFT did not have checks for the outpoint given in input. these have been added.
Functional tests have been added with the new error scenarios.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4221,11 +4221,12 @@ static UniValue IssueReissuableToken(CWallet* const pwallet, const std::string& 
         txnouttype type;
         std::vector<CTxDestination> vDest;
         int nRequired;
-        if (ExtractDestinations(scriptPubKey, type, vDest, nRequired)) {
-            for (const CTxDestination &dest : vDest)
-                if(!IsValidDestination(dest))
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -4423,7 +4423,29 @@ static UniValue issuetoken(const JSONRPCRequest& request)
         return IssueReissuableToken(pwallet, request.params[2].getValStr(), tokenValue, coin_control);
     else
     {
-        COutPoint out(uint256S(request.params[2].get_str()), request.params[3].get_int());
+        uint256 hash = uint256S(request.params[2].get_str());
```
azuchi, line 4426:
```suggestion
        uint256 txid = uint256S(request.params[2].get_str());
```

azuchi (changes requested):
The following help scriptPubkey should also be corrected to the correct one:

```cpp
            + HelpExampleCli("issuetoken", "\"1\" \"100\" 8282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508")
```

The following help is about NFT and should be quantity 1. The following command will result in an error:
```cpp
+ HelpExampleCli("issuetoken", "\"3\" \"500\" 485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954 1")
```

azuchi (approved):

PR #227, opened 2022-08-15
 only_token filter in listunspent rpc

Naviabheeman:
Added only_token option to listunspent RPC to filter only token outputs.
Updated tests


test/functional/wallet_coloredcoin.py
```diff
@@ -641,6 +676,14 @@ def run_test(self):
...
         assert_equal(self.nodes[0].getbalance(), 300)
 
+        total_unspent = len(self.nodes[0].listunspent())
+        if self.options.usecli:
+            token_unspent = len(self.nodes[0].listunspent(query_options="{\"only_token\": true}"))
+        else:
+            token_unspent = len(self.nodes[0].listunspent(query_options={"only_token": True}))
+        assert_equal(total_unspent, 6)
+        assert_equal(token_unspent, 0)
+
```
azuchi, line 686: Please add a case that includes a TPC.

azuchi (changes requested):

test/functional/wallet_coloredcoin.py
Naviabheeman: In` test_only_token_filter() ` there are tokens and TPCs. The test checks different combinations of the options like minimum and maximum confirmations in listunspent.  
Above check at the beginning of the test just verifies that "only_token" filter lists nothing as at this point in the test only TPCs are present.

PR #229, opened 2022-08-15
Move getcolor rpc to blockchain list from wallet

Naviabheeman:
moved getcolor RPC to blockchain list and added validation to its inputs


test/functional/wallet_coloredcoin.py
```diff
@@ -159,6 +159,9 @@ def test_getcolorRPC(self, utxos):
         assert_raises_rpc_error(-8, "Extra parameter for Reissuable token", self.nodes[0].getcolor, 1, utxos[4]['txid'], utxos[4]['vout'])
         assert_raises_rpc_error(-8, "Parameter missing for Non-Reissuable or NFT token", self.nodes[0].getcolor, 2, utxos[4]['txid'])
         assert_raises_rpc_error(-8, "Parameter missing for Non-Reissuable or NFT token", self.nodes[0].getcolor, 3, utxos[4]['txid'])
+        assert_raises_rpc_error(-8, "Invalid Tapyrus script", self.nodes[0].getcolor, 1, "ty")
+        assert_raises_rpc_error(-8, "txid must be hexadecimal string (not 'ty')", self.nodes[0].getcolor, 3, "ty", 0)
+        assert_raises_rpc_error(-8, "Invalid transaction id :0011001100110011001100110011001100110011001100110011001100110011", self.nodes[0].getcolor, 3, "0011001100110011001100110011001100110011001100110011001100110011", 0)
```
azuchi, line 164: With the move to blockchain RPC, this test also needs to be moved.

azuchi (changes requested):

azuchi (changes requested):
Following code in `rpcwallet.cpp` also should be moved.

```cpp
extern UniValue getcolor(const JSONRPCRequest& request);
```

src/rpc/blockchain.cpp
```diff
@@ -2058,6 +2058,96 @@ UniValue scantxoutset(const JSONRPCRequest& request)
...
+            "2. \"scriptpubkey\"     (string, optional) Script pubkey that is used to issue REISSUABLE token\n"
+            "2. \"txid\"             (string, optional) Transaction id from which the NON-REISSUABLE or NFT tokens are issued\n"
+            "3. \"index\"            (numeric, optional) Index in the above transaction id used for issuing token\n"
+            "\nResult:\n"
+            "\"color\"               (string) The color or token.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getcolor", "\"1\" 8282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508")
+            + HelpExampleCli("getcolor", "\"2\" 485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954 0")
+            + HelpExampleCli("getcolor", "\"3\" 485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954 1")
+        );
```
azuchi, line 2081: All errors occur when the example are executed.

```
$ ./src/tapyrus-cli getcolor "1" 8282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508
error code: -8
error message:
Invalid Tapyrus script
$ ./src/tapyrus-cli getcolor "2" 485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954 0
error code: -8
error message:
Invalid transaction id :485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954
$ ./src/tapyrus-cli getcolor "3" 485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954 1
error code: -8
error message:
Invalid transaction id :485273f6703f038a234400edadb543eb44b4af5372e8b207990beebc386e7954
```

azuchi (changes requested):

src/rpc/blockchain.cpp
Naviabheeman: In the first case I'll change the script. 
But the other two can't be fixed. Giving a valid transaction id is not possible, we can only have a hex string of the correct size. It can't be guaranteed to be valid in the environment where it is executed.

azuchi: I think so too. In addition to that same data as in the example should be added to the test case.

azuchi (approved):

PR #233, opened 2022-09-19
Remove block weight from Tapyrus

Naviabheeman:
Removed block weight and its related terms:
 - block weight is replaced by block size
 - witness scaling factor is removed
 - transaction virtual size is removed and transaction size is used in its place
 - sign count is 1 instead of 4 before


src/miner.cpp
```diff
@@ -50,22 +50,21 @@ int64_t UpdateTime(CBlockHeader* pblock, const Consensus::Params& consensusParam
 
 BlockAssembler::Options::Options() {
     blockMinFeeRate = CFeeRate(DEFAULT_BLOCK_MIN_TX_FEE);
-    nBlockMaxWeight = DEFAULT_BLOCK_MAX_WEIGHT;
+    nBlockMaxSize = DEFAULT_BLOCK_MAX_SIZE;
 }
 
 BlockAssembler::BlockAssembler(const CChainParams& params, const Options& options) : chainparams(params)
 {
     blockMinFeeRate = options.blockMinFeeRate;
```
azuchi, line 58: It seems that the argument `-blockmaxsize` could originally be specified. Please check past changes.

src/miner.cpp
```diff
@@ -411,7 +410,8 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
         std::vector<CTxMemPool::txiter> sortedEntries;
         SortForBlock(ancestors, sortedEntries);
 
-        for (size_t i=0; i<sortedEntries.size(); ++i) {
+        for (size_t i=0; i<sortedEntries.
+        size(); ++i) {
```
azuchi, line 414: This change does not necessary.

src/policy/policy.cpp
```diff
@@ -247,17 +239,17 @@ CFeeRate incrementalRelayFee = CFeeRate(DEFAULT_INCREMENTAL_RELAY_FEE);
...
 }
 
-int64_t GetVirtualTransactionSize(const CTransaction& tx, int64_t nSigOpCost)
+int64_t GetTransactionSize(const CTransaction& tx, int64_t nSigOpCost)
 {
-    return GetVirtualTransactionSize(GetTransactionWeight(tx), nSigOpCost);
+    return GetTransactionSize(GetSerializeSize(tx, SER_NETWORK, PROTOCOL_VERSION), nSigOpCost);
 }
 
 int64_t GetVirtualTransactionInputSize(const CTxIn& txin, int64_t nSigOpCost)
```
azuchi, line 252: Isn't the name `GetVirtualTransactionInputSize()` appropriate?

src/policy/policy.h
```diff
@@ -76,8 +76,8 @@ extern CFeeRate dustRelayFee;
 extern unsigned int nBytesPerSigOp;
 
 /** Compute the virtual transaction size (weight reinterpreted as bytes). */
```
azuchi, line 78:
```suggestion
/** Compute the transaction size. */
```

src/policy/policy.h
```diff
@@ -76,8 +76,8 @@ extern CFeeRate dustRelayFee;
...
-int64_t GetVirtualTransactionSize(int64_t nWeight, int64_t nSigOpCost);
-int64_t GetVirtualTransactionSize(const CTransaction& tx, int64_t nSigOpCost = 0);
+int64_t GetTransactionSize(int64_t nWeight, int64_t nSigOpCost);
+int64_t GetTransactionSize(const CTransaction& tx, int64_t nSigOpCost = 0);
 int64_t GetVirtualTransactionInputSize(const CTxIn& tx, int64_t nSigOpCost = 0);
```
azuchi, line 81: Isn't the name `GetVirtualTransactionInputSize()` appropriate?

src/qt/locale/bitcoin_ja.ts
```diff
@@ -2736,7 +2736,7 @@ Note:  Since the fee is calculated on a per-byte basis, a fee of "100 tapyrus pe
         <translation>トランザクションの全体サイズ</translation>
     </message>
     <message>
-        <source>Transaction virtual size</source>
+        <source>Transaction size</source>
         <translation>トランザクションの仮想サイズ</translation>
```
azuchi, line 2740:
```suggestion
        <translation>トランザクションサイズ</translation>
```

And other languages as well.

src/rpc/blockchain.cpp
```diff
@@ -1754,18 +1730,18 @@ static UniValue getblockstats(const JSONRPCRequest& request)
...
+                feerate_array.emplace_back(std::make_pair(feerate, tx_size));
             }
             maxfeerate = std::max(maxfeerate, feerate);
             minfeerate = std::min(minfeerate, feerate);
         }
     }
 
     CAmount feerate_percentiles[NUM_GETBLOCKSTATS_PERCENTILES] = { 0 };
-    CalculatePercentilesByWeight(feerate_percentiles, feerate_array, total_weight);
+    CalculatePercentilesByWeight(feerate_percentiles, feerate_array, total_size);
```
azuchi, line 1744: `CalculatePercentilesByWeight` name contains `weigth`.

src/rpc/blockchain.cpp
```diff
@@ -1774,7 +1750,7 @@ static UniValue getblockstats(const JSONRPCRequest& request)
 
     UniValue ret_all(UniValue::VOBJ);
     ret_all.pushKV("avgfee", (block.vtx.size() > 1) ? totalfee / (block.vtx.size() - 1) : 0);
-    ret_all.pushKV("avgfeerate", total_weight ? (totalfee * WITNESS_SCALE_FACTOR) / total_weight : 0); // Unit: sat/vbyte
+    ret_all.pushKV("avgfeerate", total_size > 0 ? totalfee / total_size : 0); // Unit: tap/byte
```
azuchi, line 1753: Are there any cases where total_size is less than or equal `0`?

src/rpc/mining.cpp
```diff
@@ -524,14 +522,8 @@ static UniValue getblocktemplate(const JSONRPCRequest& request)
...
-    nSigOpLimit /= WITNESS_SCALE_FACTOR;
-    assert(nSizeLimit % WITNESS_SCALE_FACTOR == 0);
-    nSizeLimit /= WITNESS_SCALE_FACTOR;
-    result.pushKV("sigoplimit", nSigOpLimit);
-    result.pushKV("sizelimit", nSizeLimit);
+    result.pushKV("sigoplimit", (int64_t)MAX_BLOCK_SIGOPS_COST);
+    result.pushKV("sizelimit", (int64_t)MAX_BLOCK_BASE_SIZE);
     result.pushKV("curtime", pblock->GetBlockTime());
     result.pushKV("proof", HexStr(pblock->proof));
     result.pushKV("height", (int64_t)(pindexPrev->nHeight+1));
```
azuchi, line 529: In `getblocktemplate` results description still remain `weight` and `weightlimit` fields.

azuchi (changes requested):

src/consensus/consensus.h
```diff
@@ -12,15 +12,16 @@
...
-static const unsigned int MAX_BLOCK_WEIGHT = 4000000;
-/** The maximum allowed number of signature check operations in a block (network rule) */
-static const int64_t MAX_BLOCK_SIGOPS_COST = 80000;
+/**
+  * The maximum allowed size for a block, in bytes (network rule).
+  */
+static const unsigned int MAX_BLOCK_BASE_SIZE = 1000000;
+/** 
+ * The maximum allowed number of signature check operations in a block (network rule) */
+static const int64_t MAX_BLOCK_SIGOPS_COST = 20000;
```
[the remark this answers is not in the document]
Naviabheeman: if this needs to be relative to block size, it can't be a constant. we need to add it to federation parameters and change it when block size changes. In this PR I'll just add the above line. In #234 I'll make the federation param changes.

src/consensus/consensus.h
```diff
@@ -12,15 +12,16 @@
 
 /** The maximum allowed size for a serialized block, in bytes (only for buffer size limits) */
 static const unsigned int MAX_BLOCK_SERIALIZED_SIZE = 4000000;
```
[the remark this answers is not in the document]
Naviabheeman: ok. I didn't change it as this value is used only in buffers sizes. removing it is going to affect #234 also when the block size becomes dynamic. But it makes the node respond better to configuration.  The node will use more memory as the block size increases.

src/qt/locale/bitcoin_ja.ts
Naviabheeman: Done! I didn't think of google translate before!

src/rpc/blockchain.cpp
Naviabheeman: Yes,coinbase transactions sizes are 0 because of lines 1686:1688
`        if (tx->IsCoinBase()) {
            continue;
        }`

src/miner.cpp
Naviabheeman: it is added now

doc/man/tapyrusd.1
```diff
@@ -498,11 +498,11 @@ not relaying transactions (default: 1)
 .PP
 Block creation options:
 .HP
-\fB\-blockmaxweight=\fR<n>
+\fB\-blockmintxfee=\fR<amt>
 .IP
-Set maximum BIP141 block weight (default: 3996000)
+Set maximum block size in bytes (default: 999000)
 .HP
-\fB\-blockmintxfee=\fR<amt>
```
azuchi, line 505: Isn't it unnecessary to change line 505(`blockmintxfee`)?

src/rpc/mining.cpp
azuchi: @Naviabheeman this is not fix yet.

src/rpc/blockchain.cpp
azuchi: @Naviabheeman this is not fix yet.

src/qt/locale/bitcoin_ja.ts
azuchi: Can you replace other `仮想サイズ` to `サイズ` as well.

src/rpc/mining.cpp
azuchi: sorry, my mistake.

src/rpc/blockchain.cpp
azuchi: sorry, my mistake.

src/consensus/consensus.h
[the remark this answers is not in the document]
Naviabheeman: Federation param (or XField history) change is not in PR234. It will be a new PR.

src/consensus/consensus.h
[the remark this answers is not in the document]
Naviabheeman: In PR234 MAX_BLOCK_SERIALIZED_SIZE is replaced with MAX_BLOCK_SIZE. but it's not dynamic according to xfield. I think I'll create a new PR to make this and MAX_BLOCK_SIGOPS dynamic.

PR #232, opened 2022-09-07
New <xfield_type:xfield> parameter in getnewblock RPC

Naviabheeman:
Implemented #230 
This PR adds a new parameter to getnewblock RPC to fill the xfieldType and xfield information in a new block being created. 
Adds new functional test for the same.

Issue #231, opened 2022-08-30
Add new xfield type to change max block size


azuchi:
The block size is currently fixed at 1MB (the implementation defines it as 4MB in weight equivalent), but to allow this size to be changed, the following new xfield type is introduced:

* xfield type: 2
* xfield value: max block size(bytes) with 4-byte hex value.

If this xfield type is specified in the block header, the new maximum block size is applied starting with the next block, if that value is valid.

Naviabheeman:
1. Due to **MAX_PROTOCOL_MESSAGE_LENGTH** set to 4 MB two nodes are not able to synch blocks larger than that.  The p2p message is not split into multiple messages when its size is large.  I changed this limit to 10 MB and ran the test. Nodes are able to create large blocks upto 10 MB and synch with each other. Blocks with non standard sigops and non standard transaction size are also accepted. How do me handle this limitation, can we make it a cmd line option so that we can restart a node with new message length if the block size changes or make this also dependant on block size?

2. When a node restarts after a block changing the block size limit in some cases it needs to be restarted with -reindex option. For example, let B30 be the block setting xfield max block size to 4 MB. Then B31, B32 are blocks of large size. IF the node restarts at this point, there is code to VeridyDB() which validated the last 6 blocks in reverse order. i.e. B32 would be checked first and at that point federation params contains only the old block size limit. So the validation fails. These are the logs from this scenario.

'2022-11-23T12:38:20Z Loaded best chain: hashBestChain=54f69a7f6365a0b8fdbce8f3e53bf98a23bdb131dfd479eb76907b689d86941b height=318 date=2022-11-23T12:11:09Z progress=1.000000
2022-11-23T12:38:20Z init message: Rewinding blocks...
2022-11-23T12:38:20Z init message: Verifying blocks...
2022-11-23T12:39:04Z Verifying last 6 blocks at level 3
2022-11-23T12:39:16Z [0%]...[16%]...[33%]...ERROR: VerifyDB: *** found bad block at 316, hash=221b92c7c660293cd2ae71ef10a56ccadf7d6db4464775f1770aa892f9dacd17 (bad-blk-length, size limits failed (code 16))

2022-11-23T12:48:02Z : Corrupted block database detected.
Please restart with -reindex or -reindex-chainstate to recover.
: Corrupted block database detected.'

Restarting the node with -reindex works fine. The node is able to start and synch to the latest block.

azuchi:
[...]

PR #234, opened 2022-10-16
New Field for Max block size change

Naviabheeman:
#231 is implemented in this PR.

Some notable changes are in this PR:
1.CXField

A new class CXField is added to CBlockHeader to handle xfield i.e XFieldType and xfield. It makes access to the data member safe based on the field type and provides constructors and copy constructors. It replaces the XFieldData class added in PR #232.

2. Federation params map:

A map in federation params keeps track of all the Xfield changes and their block heights. As long as the maxblocksize is kept at the default size i.e 1MB there is no extra data in the federation params. Only when the block size is changed the new entry is added to the map.

3. Reorg safety:

In the case of aggpubkey change we do not allow a reorg to a block which is before the federation block with aggpubkey change. Max block size change need have the same rule as it would not stop the blockchain. Max block size change can be re-orged out. In that case the block size is removed from the federation params map. In case it is brought back by another re-org, the  block size is added back to federation params map. This does carry a risk: max block size validation is not aware of the branches in the chain.


Naviabheeman:
closing this and opening a new PR as the commit history is very long. multiple rebases are adding files unrelated to this change

PR #235, opened 2022-10-27
Add ability to retry functional tests which timeout

Naviabheeman:
This PR makes changes in the functional test framework to add the following features:

1. Retry tests which timeout
      After the initial run with all eligible tests the second run lists only the tests which fail on a timeout. The error code used to identify this has been changed and new status added to identify this.

2. Use a dynamic timeout value depending on the time taken to connect to a node
      When the test framework starts a node, it remembers the time taken to establish the RPC connection to it. Every RPC request to the node uses the time_to_connect*5 as the timeout. As this is very dynamic, according to the test environment we expect that the tests would not fail because of short timeout values.

### DIFF
--- a/test/functional/p2p_unrequested_blocks.py
+++ b/test/functional/p2p_unrequested_blocks.py
@@ -77,10 +77,10 @@ class AcceptBlockTest(BitcoinTestFramework):
     def run_test(self):
         # Setup the p2p connections
         # test_node connects to node0 (not whitelisted)
-        test_node = self.nodes[0].add_p2p_connection(P2PInterface())
+        test_node = self.nodes[0].add_p2p_connection(P2PInterface(self.nodes[0].time_to_connect))
         self.nodes[0].setmocktime(self.mocktime)
         # min_work_node connects to node1 (whitelisted)
-        min_work_node = self.nodes[1].add_p2p_connection(P2PInterface())
+        min_work_node = self.nodes[1].add_p2p_connection(P2PInterface(self.nodes[1].time_to_connect))
         self.nodes[1].setmocktime(self.mocktime)
 
         # 1. Have nodes mine a block (leave IBD)
@@ -203,7 +203,7 @@ class AcceptBlockTest(BitcoinTestFramework):
         self.nodes[0].disconnect_p2ps()
         self.nodes[1].disconnect_p2ps()
 
-        test_node = self.nodes[0].add_p2p_connection(P2PInterface())
+        test_node = self.nodes[0].add_p2p_connection(P2PInterface(self.nodes[0].time_to_connect))
 
         test_node.send_message(msg_block(block_h1f))
 
@@ -288,7 +288,7 @@ class AcceptBlockTest(BitcoinTestFramework):
             test_node.wait_for_disconnect()
 
             self.nodes[0].disconnect_p2ps()
-            test_node = self.nodes[0].add_p2p_connection(P2PInterface())
+            test_node = self.nodes[0].add_p2p_connection(P2PInterface(self.nodes[0].time_to_connect))
 
         # We should have failed reorg and switched back to 290 (but have block 291)
         assert_equal(self.nodes[0].getblockcount(), 290)
--- a/test/functional/p2p_unrequested_blocks.py
+++ b/test/functional/p2p_unrequested_blocks.py
@@ -284,7 +284,7 @@ class AcceptBlockTest(BitcoinTestFramework):
             # Only wait a short while so the test doesn't take forever if we do get
             # disconnected
             test_node.sync_with_ping(timeout=1)
-        except AssertionError:
+        except TimeoutError:
             test_node.wait_for_disconnect()
 
             self.nodes[0].disconnect_p2ps()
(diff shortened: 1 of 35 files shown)


test/functional/test_framework/key.py
```diff
@@ -13,7 +13,7 @@
 import ctypes.util
 import hashlib
 
-ssl = ctypes.cdll.LoadLibrary(ctypes.util.find_library ('ssl') or 'libeay32')
+ssl = ctypes.cdll.LoadLibrary(ctypes.util.find_library ('ssl'))
```
azuchi, line 16: Why is this change necessary?

test/functional/test_framework/mininode.py
```diff
@@ -236,6 +236,8 @@ def __init__(self):
         # The network services received from the peer
         self.nServices = 0
 
+        self.timeout = time_to_connect * 5
```
azuchi, line 239: Why are you multiplying by 5 here?

test/functional/test_framework/test_node.py
```diff
@@ -231,7 +234,14 @@ def is_node_stopped(self):
...
     def wait_until_stopped(self, timeout=BITCOIND_PROC_WAIT_TIMEOUT):
-        wait_until(self.is_node_stopped, timeout=timeout)
+        i = 0
+        while i < 2:
+            try:
+                i = i + 1
+                wait_until(self.is_node_stopped, timeout=timeout)
+            except TimeoutError as e:
+                pass
+
```
azuchi, line 244: What is the intention behind this addition of logic?

test/functional/test_framework/test_node.py
```diff
@@ -245,7 +255,7 @@ def assert_start_raises_init_error(self, extra_args=None, expected_msg=None, mat
              tempfile.NamedTemporaryFile(dir=self.stdout_dir, delete=False) as log_stdout:
             try:
                 self.start(extra_args, stdout=log_stdout, stderr=log_stderr, *args, **kwargs)
-                self.wait_for_rpc_connection()
+                timeout = self.wait_for_rpc_connection()
```
azuchi, line 258: This `timeout` variable used? If not, `wait_for_rpc_connection`  does not need to return this.

test/functional/p2p_unrequested_blocks.py
```diff
@@ -284,11 +284,11 @@ def run_test(self):
             # Only wait a short while so the test doesn't take forever if we do get
             # disconnected
             test_node.sync_with_ping(timeout=1)
-        except AssertionError:
+        except TimeoutError:
```
azuchi, line 287: If it is an AssertionError, does it matter if it is no longer processed?

test/functional/test_framework/mininode.py
Naviabheeman: 5 is just a random number, there is no reason behind it. I want to make the timeout long enough to wait for network and wallet to sync

test/functional/test_framework/test_node.py
Naviabheeman: this was to double the timeout period for stopping a node. Some tests were failing because the node did not stop within the given time.

test/functional/test_framework/test_node.py
Naviabheeman: right! its not used. will fix it.

test/functional/test_framework/key.py
Naviabheeman: this is leftover from my test environment. it usually complaints about library libeay32 not found when I run functional tests. I remove it as a shortcut and continue texting. it shouldn't be in the repo. will fix it.

test/functional/test_framework/mininode.py
azuchi: can you  add this intention as a comment?

test/functional/p2p_unrequested_blocks.py
Naviabheeman: When the test is run with `AssertionError` the test exit code seems to be different and the status is
```
 2023-01-20T11:36:31.303000Z TestFramework (WARNING): Timeout. None
2023-01-20T11:36:31.360000Z TestFramework (INFO): Stopping nodes
2023-01-20T11:36:32.191000Z TestFramework (INFO): Cleaning up /var/folders/0y/rhp0p3vx2rj70m8zxhgpw1c80000gn/T/test3ss97r3u on exit
2023-01-20T11:36:32.192000Z TestFramework (INFO): Test timeout
``` 

with `TimeoutError` it is different:
```
2023-01-20T11:38:29.427000Z TestFramework (INFO): Successfully synced nodes 1 and 0
2023-01-20T11:38:29.483000Z TestFramework (INFO): Stopping nodes
2023-01-20T11:38:30.307000Z TestFramework (INFO): Cleaning up /var/folders/0y/rhp0p3vx2rj70m8zxhgpw1c80000gn/T/testvaocqx1o on exit
2023-01-20T11:38:30.307000Z TestFramework (INFO): Tests successful
```

azuchi (approved):

Issue #237, opened 2022-11-27
Issues in Aggregate public key set by federation block 


Naviabheeman:
The following problems are found while setting  aggregate pub key in federation params using xfield:
1.  When a node restarts with "-reindex" after the chain's aggregate pub key is updated by a federation block, it makes the ferderation params inconsistent. 
before restart
`[{'025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d3': 0}, {'03831a69b8009833ab5b0326012eaf489bfea35a7321b1ca15b11d88131423fafc': 12}, {'02bf2027c8455800c7626542219e6208b5fe787483689f1391d6d443ec85673ecf': 25}`
and after restart it recovers to:
`[{'025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d3': 0}, {'03831a69b8009833ab5b0326012eaf489bfea35a7321b1ca15b11d88131423fafc': 12}, {'02bf2027c8455800c7626542219e6208b5fe787483689f1391d6d443ec85673ecf': 25},
{'03831a69b8009833ab5b0326012eaf489bfea35a7321b1ca15b11d88131423fafc': 12}, {'02bf2027c8455800c7626542219e6208b5fe787483689f1391d6d443ec85673ecf': 25}])`
Genesis block aggpubkey is loaded at startup. After this during the "-reindex" phase the next two aggpubkey changes are added. Then when the tip is  updated until the latest block is reached using ConnectBlock() two more aggpubkey changes are added to the list. 

- To fix this we can make ReadAggregatePubkey() check the list before adding a new entry. 

2. It's not possible to add a new node to a network after the aggpubkey change. I.e. The p2p protocol used HEADERS message to send headers first and then the complete blocks. When a block header with aggpubkey change is present in the list, all headers following this should be checked using the new aggpubkey. But this is not happening now. So any new node would assume that the headers are invalid and disconnect.

- To fix this we need to make the p2p protocol aware of xfield in block headers.  The receiving node should process the xfield in block header during the headers message processing and not wait for the entire block. But federation params should be updated only when the block is accepted to the tip.

3. Restarting a node after an xfield change causes VerifyDB to fail. So that node doesn't start. it should be started using "-reindex"
`2022-11-28T05:47:44.662717Z [0%]...ERROR: ReadBlockFromDisk: ReadBlockFromDisk: bad-proof, Proof verification failed (code 16) nHeight = 23
[...]

PR #238, opened 2022-12-04
Fix issues in aggpubkey during node restart and new node addition in v0.5

Naviabheeman:
This PR fixes all issues reported in #237. These are the major changes made in v0.5 code:

 #### 1. Node restart Fix
 
Aggpubkey change was only in federation params and not persisted. So it was lost on restart. Now it is made persistent using blocks DB in level DB. **Key "1"** represents aggpubkey change. The value is a list of aggpubkey, height and block hash. There is no entry in the DB for genesis block. Only when the first federation block is accepted, it is changed. 
We do not allow reorg to remove a federation block in v0.5. This behaviour is not changed. So the aggpubkey list in levelDB is never removed. New entries maybe added to the list but never removed from the list.

#### 2. New node synch

Headers are validated by **`CheckBlockHeader`** during header verification(_HEADERS_ msg) and block verification(_BLOCK_ msg). The aggpubkey list was always fetched from federation params in both cases. Now this behaviour is changed to correctly process headers with aggpubkey changes. After a header is successfully verified in **`AcceptBlockHeader`**, if it contains a new xfield, it is added to a temp list created in **`ProcessNewBlockHeaders`**. This list is valid only until the headers message is processed. Later, if the block is accepted, the list is added to federation params and level DB following the old route.

### 3. Test

Federation management test case is enhanced to have 4 nodes. One node receives the blocks using submit block RPC. Node1 tests the p2p network synch, Node2 tests headers message at different important points in the test. Node3 simulates a new node in the network.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1514,6 +1514,19 @@ bool AppInitMain()
                         break;
                     }
                 }
+                //********************************************************** Step 11a: Load Xfield data from db
+                std::vector<XFieldEntry> xFieldList;
+                pblocktree->ReadXFieldAggpubkeys(xFieldList);
+                for(auto &XFieldData:xFieldList)
+                {
+                    //verify the aggpubkey from the xfield in the block db and then add to federation params
+                    CBlockIndex* pindex = LookupBlockIndex(XFieldData.blockHash);
+                    if (!pindex)//block might not be in the best chain
+                        continue;
+
+                    if((TAPYRUS_XFIELDTYPES)pindex->xfieldType == TAPYRUS_XFIELDTYPES::AGGPUBKEY && pindex->xfield == XFieldData.aggpubkey)
+                        FederationParams().ReadAggregatePubkey(XFieldData.aggpubkey, XFieldData.height);
+                }
 
                 if (!is_coinsview_empty) {
                     uiInterface.InitMessage(_("Verifying blocks..."));
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -29,6 +29,8 @@ static const char DB_FLAG = 'F';
 static const char DB_REINDEX_FLAG = 'R';
 static const char DB_LAST_BLOCK = 'l';
 
+static const char DB_XFIELD_AGGPUBKEY = '1';
+
 namespace {
 
 struct CoinEntry {
@@ -166,6 +168,17 @@ void CBlockTreeDB::ReadReindexing(bool &fReindexing) {
 
 bool CBlockTreeDB::ReadLastBlockFile(int &nFile) {
     return Read(DB_LAST_BLOCK, nFile);
+}
+
+bool CBlockTreeDB::ReadXFieldAggpubkeys(std::vector<XFieldEntry> & xFieldList) {
+    return Read(DB_XFIELD_AGGPUBKEY, xFieldList);
+}
+
+bool CBlockTreeDB::AddXFieldAggpubkey(XFieldEntry & xFieldEntry) {
+    std::vector<XFieldEntry> xFieldList;
+    Read(DB_XFIELD_AGGPUBKEY, xFieldList);
+    xFieldList.push_back(xFieldEntry);
+    return Write(DB_XFIELD_AGGPUBKEY, xFieldList);
 }
 
 CCoinsViewCursor *CCoinsViewDB::Cursor() const
--- a/src/txdb.h
+++ b/src/txdb.h
@@ -22,6 +22,34 @@ class CBlockIndex;
 class CCoinsViewDBCursor;
 class uint256;
 
+/*
+struct to store and retrieve xfield aggpubkey change information into level db blocks db
+*/ 
+
+struct XFieldEntry {
+    std::vector<unsigned char> aggpubkey;
+    int32_t height;
+    uint256 blockHash;
+
+    explicit XFieldEntry(const std::vector<unsigned char>& paggpubkey, const int32_t inheight, const uint256& inblockHash) : aggpubkey(paggpubkey), height(inheight), blockHash(inblockHash) {}
+    explicit XFieldEntry() : aggpubkey(), height(0), blockHash() {}
+
+    template<typename Stream>
+    void Serialize(Stream &s) const {
+        s << aggpubkey;
+        s << height;
+        s << blockHash;
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s) {
+        s >> aggpubkey;
+        s >> height;
+        s >> blockHash;
+    }
+};
+
+
 //! No need to periodic flush if at least this much space still available.
 static constexpr int MAX_BLOCK_COINSDB_USAGE = 10;
 //! -dbcache default (MiB)
@@ -97,6 +125,8 @@ public:
     bool WriteFlag(const std::string &name, bool fValue);
     bool ReadFlag(const std::string &name, bool &fValue);
     bool LoadBlockIndexGuts(std::function<CBlockIndex*(const uint256&)> insertBlockIndex);
+    bool ReadXFieldAggpubkeys(std::vector<XFieldEntry> & xFieldList);
+    bool AddXFieldAggpubkey(XFieldEntry & xFieldEntry);
 };
 
 #endif // BITCOIN_TXDB_H
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -165,7 +165,7 @@ public:
      * If a block header hasn't already been seen, call CheckBlockHeader on it, ensure
      * that it doesn't descend from an invalid block, and then add it to mapBlockIndex.
      */
-    bool AcceptBlockHeader(const CBlockHeader& block, CValidationState& state, CBlockIndex** ppindex) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
+    bool AcceptBlockHeader(const CBlockHeader& block, CValidationState& state, CBlockIndex** ppindex, std::vector<aggPubkeyAndHeight>* aggPubkeys = nullptr) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
     bool AcceptBlock(const std::shared_ptr<const CBlock>& pblock, CValidationState& state, CBlockIndex** ppindex, bool fRequested, const CDiskBlockPos* dbp, bool* fNewBlock) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
 
     // Block (dis)connection on a given view:
@@ -1173,7 +1173,7 @@ bool GetTransaction(const uint256& hash, CTransactionRef& txOut, const Consensus
 
 //declaration for compilation
 
-static bool CheckBlockHeader(const CBlockHeader& block, CValidationState& state, int nHeight = -1, bool fCheckPOW = true);
+static bool CheckBlockHeader(const CBlockHeader& block, CValidationState& state,  std::vector<aggPubkeyAndHeight>* aggPubkeys = nullptr, int nHeight = -1, bool fCheckPOW = true);
 
 
 
@@ -1222,7 +1222,7 @@ bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, int nHeight)
     }
 
     CValidationState state;
-    if(!CheckBlockHeader(block.GetBlockHeader(), state, nHeight, true))
+    if(!CheckBlockHeader(block.GetBlockHeader(), state, nullptr, nHeight, true))
         return error("%s: ReadBlockFromDisk: %s nHeight = %d", __func__, FormatStateMessage(state), nHeight);
 
     return true;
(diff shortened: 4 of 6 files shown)


src/validation.cpp
```diff
@@ -2366,8 +2366,12 @@ bool CChainState::ConnectTip(CValidationState& state, CBlockIndex* pindexNew, co
 
         // if the block was added successfully and it is a federation block,
         // make sure that the aggregatepubkey from this block is added to CFederationParams
-        if(blockConnecting.xfieldType == 1 && blockConnecting.xfield.size() == CPubKey::COMPRESSED_PUBLIC_KEY_SIZE && (CPubKey(blockConnecting.xfield.begin(), blockConnecting.xfield.end()) != FederationParams().GetLatestAggregatePubkey()))
+        if((TAPYRUS_XFIELDTYPES)blockConnecting.xfieldType == TAPYRUS_XFIELDTYPES::AGGPUBKEY && blockConnecting.xfield.size() == CPubKey::COMPRESSED_PUBLIC_KEY_SIZE && (CPubKey(blockConnecting.xfield.begin(), blockConnecting.xfield.end()) != FederationParams().GetLatestAggregatePubkey()))
```
azuchi, line 2369: Since the xfield type and value checks for aggregate public key are found in several codes, it would be better to provide them as method.

src/validation.cpp
```diff
@@ -3007,7 +3033,8 @@ static bool CheckBlockHeader(const CBlockHeader& block, CValidationState& state,
     if(!proofSize)
         return state.Invalid(false, REJECT_INVALID, "bad-proof", "No Proof in block");
 
-    CPubKey aggregatePubkey = FederationParams().GetAggPubkeyFromHeight(nHeight);
+    CPubKey aggregatePubkey = aggPubkeys ?
+         [nHeight](const std::vector<aggPubkeyAndHeight>& list) { return list.rbegin()->aggpubkey; }(*aggPubkeys) : FederationParams().GetAggPubkeyFromHeight(nHeight);
 
     if(!aggregatePubkey.IsValid())
```
azuchi, line 3039: There are following check in the above code, why is this check necessary?

```cpp
if(!aggregatePubkeyinBlock.IsFullyValid())
```

src/validation.cpp
```diff
@@ -3228,9 +3234,20 @@ bool CChainState::AcceptBlockHeader(const CBlockHeader& block, CValidationState&
...
-        if (!CheckBlockHeader(block, state))
+        if (!CheckBlockHeader(block, state, aggPubkeys))
             return error("%s: Consensus::CheckBlockHeader: %s, %s", __func__, hash.ToString(), FormatStateMessage(state));
 
+        //if this header was valid and has an aggpubkey change remember it until we finish processing the headers message
+        if(aggPubkeys && (TAPYRUS_XFIELDTYPES)block.xfieldType == TAPYRUS_XFIELDTYPES::AGGPUBKEY
+          && block.xfield.size() == CPubKey::COMPRESSED_PUBLIC_KEY_SIZE
+          && CPubKey(block.xfield.begin(), block.xfield.end()) != aggPubkeys->rbegin()->aggpubkey)
+        {
+            aggPubkeyAndHeight x;
```
azuchi, line 3245: Although not directly related to this change, it would be better to rename this struct to `AggPubkeyAndHeight`.

src/validation.cpp
```diff
@@ -3228,9 +3234,20 @@ bool CChainState::AcceptBlockHeader(const CBlockHeader& block, CValidationState&
...
+            x.aggpubkey = CPubKey(block.xfield.begin(), block.xfield.end());
+            x.height = -1;
```
azuchi, line 3247: Why this height is `-1`?

azuchi (changes requested):
Please add description for xfield database structure in `doc/tapyrus/signedblocks.md`.

azuchi:
@Naviabheeman What happens if I switch a v0.5.1 node that has an updated aggpubkey but has not saved it to the DB to a version that merges this PR?

src/txdb.h
```diff
@@ -22,6 +22,34 @@ class CBlockIndex;
 class CCoinsViewDBCursor;
 class uint256;
 
+/*
+struct to store and retrieve xfield aggpubkey change information into level db blocks db
+*/ 
+
+struct XFieldEntry {
+    std::vector<unsigned char> aggpubkey;
+    int32_t height;
```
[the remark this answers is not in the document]
Naviabheeman: Let's change `uint64_t` to `uint32_t` instead. As we expect block height to be present in `prevout.n` in the coinbase transaction it should only be 4 bytes.

src/txdb.cpp
```diff
@@ -168,6 +170,17 @@ bool CBlockTreeDB::ReadLastBlockFile(int &nFile) {
...
 }
 
+bool CBlockTreeDB::ReadXFieldAggpubkeys(std::vector<XFieldEntry> & xFieldList) {
+    return Read(DB_XFIELD_AGGPUBKEY, xFieldList);
+}
+
+bool CBlockTreeDB::AddXFieldAggpubkey(XFieldEntry & xFieldEntry) {
+    std::vector<XFieldEntry> xFieldList;
+    Read(DB_XFIELD_AGGPUBKEY, xFieldList);
+    xFieldList.push_back(xFieldEntry);
```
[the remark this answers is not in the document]
Naviabheeman: There is a check before this call in `validation.cpp` 

`(CPubKey(...) != FederationParams().GetLatestAggregatePubkey()))`

This check verifies that the new aggregate pub key is not the same as the previous one. But older aggregate pub keys may be accepted again.

This scenario is also covered by the functional test where the first two aggregate pub keys are repeated
```
expectedAggPubKeys = [
            { self.aggpubkeys[0] : 0},
            { self.aggpubkeys[1] : 12},
            { self.aggpubkeys[2] : 25},
            { self.aggpubkeys[3] : 33},
            { self.aggpubkeys[4] : 37},
            { self.aggpubkeys[0] : 42},
            { self.aggpubkeys[1] : 43}, ]

File: src/txdb.h (hunk 5 above)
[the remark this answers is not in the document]
Comment (azuchi) [reply src/txdb.h:31]: ok. let's change `uint32_t`.

File: src/validation.cpp (hunk 2 above)
Comment (Naviabheeman) [reply src/validation.cpp:3039]: This is the check used to validate pubkey inputs throughout the code base like `importpubkey` rpc. It checks the pub key validity using `secp256k1` library.

File: src/txdb.cpp (hunk 6 above)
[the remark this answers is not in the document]
Comment (azuchi) [reply src/txdb.cpp:180]: I see, but `XFieldEntry` has block height and hash in addition to aggregated public key. therefore this value do not duplicate. and it would be better to make this check a feature of this method, so that the necessary checks are not dispersed at the time of the call.

File: src/validation.cpp (hunk 4 above)
Comment (Naviabheeman) [reply src/validation.cpp:3247]: "-1" because we do not expect it to be a valid value in the blockchain.  As this is within headers message processing, we do not know the actual block height. Since this is a temporary list and only the last aggpubkey in the list is always used to check the next block proof, height is unused.

File: src/validation.cpp (hunk 2 above)
Comment (azuchi) [reply src/validation.cpp:3039]: yes, I mean

```cpp
    switch((TAPYRUS_XFIELDTYPES)block.xfieldType)
    {
        case TAPYRUS_XFIELDTYPES::AGGPUBKEY:
            if(block.xfield.size() == CPubKey::COMPRESSED_PUBLIC_KEY_SIZE)
            {
                CPubKey aggregatePubkeyinBlock(block.xfield.begin(), block.xfield.end());
                if(!aggregatePubkeyinBlock.IsFullyValid())
                    return state.Invalid(false, REJECT_INVALID, "bad-aggpubkey", "invalid aggregatePubkey");
            }
            else
                return state.Invalid(false, REJECT_INVALID, "bad-xfieldType-xfield", "invalid aggregatePubkey");
            break;
...
```

This code checks aggregated pubkey validity in `aggregatePubkeyinBlock.IsFullyValid()`, but after that

```cpp
    CPubKey aggregatePubkey = aggPubkeys ?
         [nHeight](const std::vector<aggPubkeyAndHeight>& list) { return list.rbegin()->aggpubkey; }(*aggPubkeys) : FederationParams().GetAggPubkeyFromHeight(nHeight);

    if(!aggregatePubkey.IsValid())
        return state.Error("Invalid aggregatePubkey");
```

this code checks ` if(!aggregatePubkey.IsValid())`. This check is necessary?

Naviabheeman:
> What happens if I switch a v0.5.1 node that has an updated aggpubkey but has not saved it to the DB to a version that merges this PR

I think it would only require a "-_reindex_" after the upgrade. The federation params and db aggpubkey entry will be inn synch when the blockchain is traced back upto the tip in `ConnectTip()`. I can test it and confirm.

src/txdb.h
```diff
@@ -22,6 +22,34 @@ class CBlockIndex;
 class CCoinsViewDBCursor;
 class uint256;
 
+/*
+struct to store and retrieve xfield aggpubkey change information into level db blocks db
+*/ 
+
+struct XFieldEntry {
```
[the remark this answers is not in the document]
Naviabheeman: PR #234 contains the better classes that can manage multiple xfields. After this PR I'll try to make those methods templates so that we need not repeat writing similar code for each type.

azuchi:
> I think it would only require a "-_reindex_" after the upgrade. The federation params and db aggpubkey entry will be inn synch when the blockchain is traced back upto the tip in `ConnectTip()`. I can test it and confirm.

ok. please check it.

src/validation.cpp
Naviabheeman: ok! I didn't understand the question... its fixed now

src/validation.cpp
Naviabheeman: in the new commit I moved this code such that the index entry is created and height is a valid value.

src/test/txdb_tests.cpp
```diff
@@ -0,0 +1,101 @@
...
+    try {
+        stream4 >> xFieldAggpubkeyList;
+        BOOST_CHECK_MESSAGE(false, "We should have thrown");
+    } catch (const std::ios_base::failure& e) {
+    }
+
+    // extra data
+    CDataStream stream5(ParseHex("0121025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d30a00000000000000000000000000000000000000000000000000000000000000000000002103b44f1cfcf46aba8bc98e2fd39f137cc43d98ab7792e4848b09c06198b042ca8b00000000d39be6bcedc39831203a6ab531040f9ab30d1e66c1cbcd657c38b590d91e0da700"), SER_NETWORK, PROTOCOL_VERSION);
+    stream5 >> xFieldAggpubkeyList;
+    BOOST_CHECK_EQUAL(xFieldAggpubkeyList.size(), 1);
```
azuchi, line 72: How handled about extra data?

src/test/txdb_tests.cpp
```diff
@@ -0,0 +1,101 @@
...
+    stream1 >> xFieldAggpubkeyList;
+
+    xFieldAggpubkey = xFieldAggpubkeyList[1];
+    BOOST_CHECK_EQUAL(HexStr(xFieldAggpubkey.aggpubkey.begin(), xFieldAggpubkey.aggpubkey.begin()+xFieldAggpubkey.aggpubkey.size()), "03b44f1cfcf46aba8bc98e2fd39f137cc43d98ab7792e4848b09c06198b042ca8b");
+    BOOST_CHECK_EQUAL(xFieldAggpubkey.height, 0);
+    BOOST_CHECK_EQUAL(xFieldAggpubkey.blockHash, uint256S("a70d1ed990b5387c65cdcbc1661e0db39a0f0431b56a3a203198c3edbce69bd3"));
+
+    // 5 items - error
+    CDataStream stream2(ParseHex("0521025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d30a00000000000000000000000000000000000000000000000000000000000000000000002103b44f1cfcf46aba8bc98e2fd39f137cc43d98ab7792e4848b09c06198b042ca8b00000000d39be6bcedc39831203a6ab531040f9ab30d1e66c1cbcd657c38b590d91e0da7"), SER_NETWORK, PROTOCOL_VERSION);
+
```
azuchi, line 36: What happens if the number exceeds `255(0xff)`?

azuchi (changes requested):

src/test/txdb_tests.cpp
```diff
@@ -0,0 +1,101 @@
...
+    XFieldAggpubkey xFieldAggpubkey;
+    std::vector<XFieldAggpubkey> xFieldAggpubkeyList;
+    CDataStream stream(ParseHex("0121025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d30a0000000000000000000000000000000000000000000000000000000000000000000000"), SER_NETWORK, PROTOCOL_VERSION);
+    stream >> xFieldAggpubkeyList;
+
+    // 1 item
+    xFieldAggpubkey = xFieldAggpubkeyList[0];
+    BOOST_CHECK_EQUAL(HexStr(xFieldAggpubkey.aggpubkey.begin(), xFieldAggpubkey.aggpubkey.begin()+xFieldAggpubkey.aggpubkey.size()), "025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d3");
+    BOOST_CHECK_EQUAL(xFieldAggpubkey.height, 10);
+    BOOST_CHECK_EQUAL(xFieldAggpubkey.blockHash, uint256S("0000000000000000000000000000000000000000000000000000000000000000000000"));
```
azuchi, line 23: `0000000000000000000000000000000000000000000000000000000000000000000000` value is 35 bytes, `0000000000000000000000000000000000000000000000000000000000000000` is correct.

doc/tapyrus/signedblocks.md
```diff
@@ -38,7 +38,11 @@ It reads the aggregate public key from the *genesis block file* named `genesis.d
 The aggregate public key is generated by tapyrus-signer network during network initialization. 
 The process involves creation of genesis block using tapyrus-genesis utility and 
 starting the tapyrus-signer network to *sign* the genesis block. 
-Genesis Block header also has the aggregate public key in x_field. Thus the aggregated public key is embedded into the chain.
+Genesis Block header also has the aggregate public key in x_field. Thus the aggregated public key is embedded into the chain. Tapyrus-core allows the aggregate public key to be updated according the change in the federation controlling tapyrus-signer network. This is done using xfield in the block header. In order to persist this aggregate public key change, it is stored in the block tree database in level db
+
+| Key  | Value |
+|------|-------|
+| 1    | <n><agg_pub_key_1><height_1><block_hash_1><agg_pub_key_2><height_2><block_hash_2>...<agg_pub_key_n><height_n><block_hash_n>|
```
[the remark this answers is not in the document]
azuchi: Please define the exact data format with endian for each of `n`, `height_1`, `agg_pub_key_1`, and `block_hash_n`.

src/init.cpp
```diff
@@ -1514,6 +1514,19 @@ bool AppInitMain()
...
+                std::vector<XFieldAggpubkey> xFieldList;
+                pblocktree->ReadXFieldAggpubkeys(xFieldList);
+                for(auto &XFieldData:xFieldList)
+                {
+                    //verify the aggpubkey from the xfield in the block db and then add to federation params
+                    CBlockIndex* pindex = LookupBlockIndex(XFieldData.blockHash);
+                    if (!pindex)//block might not be in the best chain
+                        continue;
+
+                    if((TAPYRUS_XFIELDTYPES)pindex->xfieldType == TAPYRUS_XFIELDTYPES::AGGPUBKEY && pindex->xfield == XFieldData.aggpubkey)
```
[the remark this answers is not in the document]
Naviabheeman: yes! will fix it

src/test/txdb_tests.cpp
Naviabheeman: It uses VarInt to serialise and deserialise these length fields. The resulting integer would be anywhere from one byte to 5 bytes in the stream. I added a serialise test with a list of size 300 to verify this.

src/test/txdb_tests.cpp
Naviabheeman: Sorry! that was copied from the stream. But when the length is more there is no error!

src/test/txdb_tests.cpp
Naviabheeman: It is just ignored. Only the data that can be deserialised id read from the stream.
IF we want to avoid extra data being added to the db itself we have to add some kind of checksum.

azuchi: Can you add length check? Since `n`, the number of data, is added at the beginning, the remaining data should be `n*70` bytes.

azuchi (changes requested):

src/validation.cpp
azuchi: As a result, can type of height of XFieldAggpubkey also be changed from int32_t to uint32_t, along with `AggPubkeyAndHeight` struct?

azuchi (approved):

PR #245, opened 2023-02-12
Dynamic block sigops and max block size

Naviabheeman:
Block size and block sig ops limits are made dynamic depending on max block size which is read from xfield history


src/consensus/consensus.h
```diff
@@ -14,9 +14,8 @@
   * The maximum allowed size for a block, in bytes (network rule).
   */
 static const unsigned int MAX_BLOCK_SIZE = 1000000;
-/** 
- * The maximum allowed number of signature check operations in a block (network rule) */
-static const int64_t MAX_BLOCK_SIGOPS = MAX_BLOCK_SIZE/50;
+/** Default blockmax size, which controls the range of block size the mining code will create **/
+static const unsigned int DEFAULT_BLOCK_MAX_SIZE = MAX_BLOCK_SIZE - 1000;
```
azuchi, line 18: What is the difference between `MAX_BLOCK_SIZE` and `DEFAULT_BLOCK_MAX_SIZE`?

src/consensus/tx_verify.cpp
```diff
@@ -135,9 +136,9 @@ unsigned int GetP2SHSigOpCount(const CTransaction& tx, const CCoinsViewCache& in
     return nSigOps;
 }
 
-int64_t GetTransactionSigOps(const CTransaction& tx, const CCoinsViewCache& inputs, int flags)
+int32_t GetTransactionSigOps(const CTransaction& tx, const CCoinsViewCache& inputs, int flags)
```
azuchi, line 139: What are the reasons for this type of change?

src/xfieldhistory.cpp
```diff
@@ -82,4 +82,14 @@ bool IsXFieldNew(const CXField& xfield, CXFieldHistoryMap* pxfieldHistory)
     return xfield.IsValid()
           && std::find(XFIELDTYPES_INIT_LIST.begin(), XFIELDTYPES_INIT_LIST.end(), xfield.xfieldType) != XFIELDTYPES_INIT_LIST.end()
           && !boost::apply_visitor(checkVisitor, xfield.xfieldValue);
-}
\ No newline at end of file
+}
+
+uint32_t CURRENT_MAXBLOCKSIZE() {
```
azuchi, line 87: These upper snake cases are easily mistaken for constants, so the function name should be camel case as well as others.

test/functional/feature_xfield_maxblocksize.py
```diff
@@ -32,7 +32,7 @@
 from test_framework.script import CScript, OP_CHECKSIG, OP_TRUE, SignatureHash, SIGHASH_ALL
 from test_framework.messages import CTransaction, MAX_BLOCK_BASE_SIZE, CTxOut, CTxIn, COutPoint, uint256_from_str, ser_compact_size, msg_headers, CBlockHeader
 
-MAX_BLOCK_SIGOPS = 20000
+MAX_BLOCK_SIGOPS = int((100000 - 1000) / 50)
```
azuchi, line 35: What is the basis for this formula?

azuchi (changes requested):

test/functional/feature_xfield_maxblocksize.py
Naviabheeman: 20000 would be the value of  MAX_BLOCK_SIZE/50. we now need it to be DEFAULT_BLOCK_MAX_SIZE/50 as DEFAULT_BLOCK_MAX_SIZE is the initial value in XFieldHistory for MaxBlockSize

src/consensus/tx_verify.cpp
Naviabheeman: Our max block size xfield is only 4 bytes.  Transaction Sigops, block signups are all derived from it. So I made them all 4 bytes.

azuchi (changes requested):

azuchi:
@Naviabheeman In the following code, MAX_BLOCK_SIZE is referenced, is this a problem?

in `coins.cpp`:

```cpp
static const size_t MAX_OUTPUTS_PER_BLOCK = MAX_BLOCK_SIZE / MIN_TRANSACTION_OUTPUT_SIZE;

const Coin& AccessByTxid(const CCoinsViewCache& view, const uint256& txid)
{
    COutPoint iter(txid, 0);
    while (iter.n < MAX_OUTPUTS_PER_BLOCK) {
        const Coin& alternate = view.AccessCoin(iter);
        if (!alternate.IsSpent()) return alternate;
        ++iter.n;
    }
    return coinEmpty;
}
```

in `tapyrus-tx.cpp`:

```cpp
static const unsigned int maxVout = MAX_BLOCK_SIZE / minTxOutSz;
```

Naviabheeman:
> ```c++
> static const size_t MAX_OUTPUTS_PER_BLOCK = MAX_BLOCK_SIZE / MIN_TRANSACTION_OUTPUT_SIZE;
> ```

These were intentionally left as they were. `MAX_OUTPUTS_PER_BLOCK` is guessing the maximum possible vout and using it to get any unspent output from coins db. If we make this dynamic the range will be huge and make the worse case scenario i.e. when transaction does not have unspent outputs even worse.
On the other hand if we leave this as it is, only two places are affected:   `GetTransaction` RPC and retrieving transaction undo data during a reorg. In case a the block size becomes bigger, say 3MB and there is a huge transaction with more than MAX_OUTPUTS_PER_BLOCK outputs, then gettransaction RPC on that transaction will fail. And in case this huge transaction is reversed in a reorg then the undo input may not be available. 
Since these two scenarios are rare I think we can leave MAX_OUTPUTS_PER_BLOCK limit as it is. Adding a new API to level db wrapper which can get all the outputs of a transaction would be a better solution.

PR #250, opened 2023-03-14
upgrade boost to 1.81

Naviabheeman:
upgrade boost to 1.81 with b2 as the build tool.

update all boost m4 scripts from http://www.randspringer.de/boost/index.html

fix windows linking: -pthread flag was added incorrectly to windows linking in lib tool. Fixed it my adding -DPIC flag

missing header causing build errors have been fixed.


configure.ac
```diff
@@ -614,7 +620,9 @@ dnl Check for endianness
 AC_C_BIGENDIAN
 
 dnl Check for pthread compile/link requirements
-AX_PTHREAD
+if test $host !=  *mingw*; then
+   AX_PTHREAD
+fi
```
azuchi, line 625: @Naviabheeman Bitcoin Core does not appear to have made this change. Why is this necessary?

Naviabheeman: this was an attempt to fix the pthread flag. I think it's not necessary. I'll wait for the CI to finish and verify.

azuchi (approved):

PR #258, opened 2023-05-10
Upgrade Qt to 5.15.5

Naviabheeman:
tapyrus Qt and its dependencies are upgraded to qt5.15.5.
Qt 5.15.5 compiles only on c++17. So tapyrus c++ version is upgraded to c++17.

bitcoin builds qt as a static plugin and adds more dependencies to support it. This is not adapted in tapyrus to avoid more package dependencies. Builds with static(without additional dependencies) caused errors complaining about missing xcb, qpa modules.


Naviabheeman:
macOS build failure is fixed by #260.

Naviabheeman:
closing as #259 is merged. it includes this upgrade.

PR #259, opened 2023-05-11
Upgrade depends libraries

Naviabheeman:
Tapyrus depends libraries are now equivalent to bitcoin. 
- libraries like zeromq, qrencode, Libevent have been upgraded

- dbus and native_libdmg-hfsplus are removed

- bdb is patched to compile with c++17

PR #260, opened 2023-05-14
Fix Darwin univalve build

Naviabheeman:
Upgrade UniValue source files from upstream
Change makefiles such that the same relative path to macOS toolchain is used in src and in UniValue directories. That fixes the archiver missing error.


azuchi:
duplicate #259

Issue #241, opened 2022-12-23
Make uncompressed aggregate pub keys uniform across all tapyrus binaries


Naviabheeman:
**generate** and **generatetoaddress** RPCs use the following check to verify whether the private key parameter corresponds to the aggregate public key. 

`
    if(cPrivKey.GetPubKey() != FederationParams().GetLatestAggregatePubkey())
        throw JSONRPCError(RPC_WALLET_INVALID_AGGREGATE_KEY, "Given private key doesn't correspond to the Aggregate Key.");
`

This is not correct. When the aggregate pub key is not an exact match to the randomly generated pub key blocks are not generated. But the private key is indeed correct. It should be

`  if(!cPrivKey.VerifyPubKey(FederationParams().GetLatestAggregatePubkey())`

Naviabheeman:
The real issue here was aggpubkey format. Compressed and uncompressed private keys are not handled uniformly by 'tapyrus-genesis' and Tapyrus-core RPCs. 
**To avoid confusion like this, in future versions, tapyrus-genesis should check that the WIF format is compressed, and the generate command should output a dedicated error if the WIF is uncompressed.**

PR #262, opened 2023-05-14
Check uncompresses key in tapyrus-genesis.

Naviabheeman:
Accepting uncompresses keys causes confusion as reported in #241 i.e Tapyrus-core can start with the genesis block created with uncompresses pub key. But new blocks cannot be created as proof verification expected only compresses pub key. So the chain cannot progress.

### DIFF
--- a/src/tapyrus-genesis.cpp
+++ b/src/tapyrus-genesis.cpp
@@ -153,7 +153,7 @@ static int CommandLine()
             }
             if(!aggpubkey.IsCompressed())
             {
-                fprintf(stderr, "Error: Uncompressed Aggregate Public Keys are not supported.\n");
+                fprintf(stderr, "Error: Uncompressed Aggregate Public Key are not supported.\n");
                 return EXIT_FAILURE;
             }
         }
--- a/src/tapyrus-genesis.cpp
+++ b/src/tapyrus-genesis.cpp
@@ -151,6 +151,11 @@ static int CommandLine()
                 fprintf(stderr, "Error: Aggregate Public Key was invalid.\n");
                 return EXIT_FAILURE;
             }
+            if(!aggpubkey.IsCompressed())
+            {
+                fprintf(stderr, "Error: Uncompressed Aggregate Public Keys are not supported.\n");
+                return EXIT_FAILURE;
+            }
         }
     } catch (const std::exception& e) {
         fprintf(stderr, "Error: %s\n", e.what());
--- a/src/tapyrus-genesis.cpp
+++ b/src/tapyrus-genesis.cpp
@@ -153,7 +153,7 @@ static int CommandLine()
             }
             if(!aggpubkey.IsCompressed())
             {
-                fprintf(stderr, "Error: Uncompressed Aggregate Public Key are not supported.\n");
+                fprintf(stderr, "Error: Uncompressed Aggregate Public Keys are not supported.\n");
                 return EXIT_FAILURE;
             }
         }
--- a/src/tapyrus-genesis.cpp
+++ b/src/tapyrus-genesis.cpp
@@ -153,7 +153,7 @@ static int CommandLine()
             }
             if(!aggpubkey.IsCompressed())
             {
-                fprintf(stderr, "Error: Uncompressed Aggregate Public Keys are not supported.\n");
+                fprintf(stderr, "Error: Uncompressed Aggregate Public Key is not supported.\n");
                 return EXIT_FAILURE;
             }
         }
--- a/test/util/data/tapyrus-util-test.json
+++ b/test/util/data/tapyrus-util-test.json
@@ -775,7 +775,7 @@
     ],
     "return_code": 1,
     "error_txt": "Error: Aggregate private key does not correspond to given Aggregate public key",
-    "description": "Ensure aggregate pubkey and aggregate private correspond to each other."
+    "description": "Ensure aggregate pubkey and aggregate private key correspond to each other."
   },
   {
     "exec": "./tapyrus-genesis",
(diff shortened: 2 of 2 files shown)


test/util/data/tapyrus-util-test.json
```diff
@@ -733,5 +733,59 @@
...
+    "exec": "./tapyrus-genesis",
+    "args": [
+      "-dev",
+      "-time=1563342688",
+      "-signblockpubkey=04473757a955a23f75379820f3071abf5b3343b78eb54e52373d06259ffa6c550bd0acc38ff1b182a5de0c795aa7109e8348ca5436d65a267ef134af595ec6c75e",
+      "-signblockprivatekey=L55noj9NZSxyX9E6M2ijiUQSDZeXK8ga5YRmHpeFcERmBnvc7bS3"
+    ],
+    "return_code": 1,
+    "error_txt": "Error: Uncompressed Aggregate Public Keys are not supported",
+    "description": "Ensure uncompressed pubkeys are not accepted as aggregate pubkey in dev mode."
```
azuchi, line 758: > Ensure uncompressed pubkeys are not accepted as aggregate pubkey in dev mode.

but,, `L55noj9NZSxyX9E6M2ijiUQSDZeXK8ga5YRmHpeFcERmBnvc7bS3` dose not dev mode format.

test/util/data/tapyrus-util-test.json
```diff
@@ -733,5 +733,59 @@
...
+  },
+  {
+    "exec": "./tapyrus-genesis",
+    "args": [
+      "-signblockpubkey=04b0da749730dc9b4b1f4a14d6902877a92541f5368778853d9c4a0cb7802dcfb25e01fc8fde47c96c98a4f3a8123e33a38a50cf9025cc8c4494a518f991792bb7",
+      "-signblockprivatekey=5KSSJQ7UNfFGwVgpCZDSHm5rVNhMFcFtvWM3zQ8mW4qNDEN7LFd"
+    ],
+    "return_code": 1,
+    "error_txt": "Error: Uncompressed Aggregate Public Keys are not supported",
+    "description": "Ensure uncompressed pubkeys are not accepted as aggregate pubkey in dev mode."
```
azuchi, line 768: What is the difference between this case and the one before it?

test/util/data/tapyrus-util-test.json
```diff
@@ -733,5 +733,59 @@
...
+  {
+    "exec": "./tapyrus-genesis",
+    "args": [
+      "-dev",
+      "-signblockpubkey=025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d3",
+      "-signblockprivatekey=cUJN5RVzYWFoeY8rUztd47jzXCu1p57Ay8V7pqCzsBD3PEXN7Dd4"
+    ],
+    "return_code": 1,
+    "error_txt": "Error: Aggregate private key does not correspond to given Aggregate public key",
+    "description": "Ensure aggregate pubkey and aggregate private correspond to each other in dev mode."
```
azuchi, line 789:
```suggestion
    "description": "Ensure aggregate pubkey and aggregate private key correspond to each other in dev mode."
```

test/util/data/tapyrus-util-test.json
```diff
@@ -733,5 +733,59 @@
...
+  },
+  {
+    "exec": "./tapyrus-genesis",
+    "args": [
+      "-signblockpubkey=025700236c2890233592fcef262f4520d22af9160e3d9705855140eb2aa06c35d3",
+      "-signblockprivatekey=5KSSJQ7UNfFGwVgpCZDSHm5rVNhMFcFtvWM3zQ8mW4qNDEN7LFd"
+    ],
+    "return_code": 1,
+    "error_txt": "Error: Aggregate private key does not correspond to given Aggregate public key",
+    "description": "Ensure aggregate pubkey and aggregate private correspond to each other."
```
azuchi, line 778:
```suggestion
    "description": "Ensure aggregate pubkey and aggregate private key correspond to each other."
```

azuchi (changes requested):

test/util/data/tapyrus-util-test.json
Naviabheeman: These keys were generated  from an online tool. sorry, I hadn't cross checked it. 
But the error here is generated from the pub key. Private key is not yet validated when the error happens.

test/util/data/tapyrus-util-test.json
Naviabheeman: this shouldn't be dev mode. there is no "-dev" parameter. Fixed it. Other than that it's the same test. just different pub key.

test/util/data/tapyrus-util-test.json
azuchi: > But the error here is generated from the pub key. Private key is not yet validated when the error happens.

I see. However, considering the intent of the test, the key should be in the correct format.

src/tapyrus-genesis.cpp
```diff
@@ -151,6 +151,11 @@ static int CommandLine()
                 fprintf(stderr, "Error: Aggregate Public Key was invalid.\n");
                 return EXIT_FAILURE;
             }
+            if(!aggpubkey.IsCompressed())
+            {
+                fprintf(stderr, "Error: Uncompressed Aggregate Public Keys are not supported.\n");
```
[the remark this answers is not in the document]
azuchi: I think `key` is correct, not `keys`, because an aggregate public key is a single key that aggregates multiple public keys. Therefore, this should be changed, including testing code.

azuchi (approved):

PR #263, opened 2023-05-14
Rename BitcoinGUI to tapyrus GUI

Naviabheeman:
Rename Bitcoin GUI to Tapyrus GUI


src/qt/README.md
```diff
@@ -30,15 +30,15 @@ Resources such as the icon.
 
 Tests.
 
-### bitcoingui.(h/cpp)
+### TapyrusGUI.(h/cpp)
 
 Represents the main window of the Bitcoin UI.
```
azuchi, line 35:
```suggestion
Represents the main window of the Tapyrus UI.
```

src/qt/tapyrusgui.h
```diff
@@ -49,15 +49,15 @@ QT_END_NAMESPACE
   Bitcoin GUI main class. This class represents the main window of the Bitcoin UI. It communicates with both the client and
```
azuchi, line 49:
```suggestion
  Tapyrus GUI main class. This class represents the main window of the Tapyrus UI. It communicates with both the client and
```

azuchi (changes requested):
There still seems to be a BitcoinGUI within the documentation and comments.

src/qt/tapyrus.cpp
```diff
@@ -543,7 +543,7 @@ static void SetupUIArgs()
     gArgs.AddArg("-resetguisettings", "Reset all settings changed in the GUI", false, OptionsCategory::GUI);
     gArgs.AddArg("-rootcertificates=<file>", "Set SSL root certificates for payment request (default: -system-)", false, OptionsCategory::GUI);
     gArgs.AddArg("-splash", strprintf("Show splash screen on startup (default: %u)", DEFAULT_SPLASHSCREEN), false, OptionsCategory::GUI);
-    gArgs.AddArg("-uiplatform", strprintf("Select platform to customize UI for (one of windows, macosx, other; default: %s)", BitcoinGUI::DEFAULT_UIPLATFORM), true, OptionsCategory::GUI);
+    gArgs.AddArg("-uiplatform", strprintf("Select platform to customize UI for (one of windows, macosx, other; default: %s)", TapyrusGUI::DEFAULT_UIPLATFORM), true, OptionsCategory::GUI);
 }
 
 #ifndef BITCOIN_QT_TEST
```
azuchi, line 549: This is also the target?

Naviabheeman: Yes, this is intentional.  `DEFAULT_UIPLATFORM` is a static string defined inside the class `TapyrusGUI` . We changed the class name.

`static const std::string DEFAULT_UIPLATFORM;`

azuchi: oh, I'm talking about

```cpp
#ifndef BITCOIN_QT_TEST
```

Naviabheeman: sorry! you mean change BITCOIN_QT_TEST to TAPYRUS_QT_TEST? Most of the include guards and conditional compilation are still starting with BITCOIN_ Maybe I'll make a separate PR for these and the their makefile changes?

azuchi: ok.  Let's make a separate PR for this.

azuchi (approved):

PR #268, opened 2023-06-03
Change boost classes to std classes

Naviabheeman:
change boost::bind and boost::variant to std::bind and std::variant

boost::chrono change in not in this PR. it is part of #269


src/validation.cpp
```diff
@@ -2293,15 +2293,13 @@ class ConnectTrace {
 private:
     std::vector<PerBlockConnectTrace> blocksConnected;
     CTxMemPool &pool;
+    boost::signals2::scoped_connection m_connNotifyEntryRemoved;
```
azuchi, line 2296: Why was this variable added?

Naviabheeman: This variable now manages releasing the connection when the object goes out of scope instead of doing it in the destructor.  

There are similar changes in `validationinterface.cpp `which are in PR#269. There I had compilation errors when replacing `boost::bind` with `std::bind`. The signal methods were not resolved in the destructor.  using these  connection variables help to remember the signal and slot to destroy in the end.  This file is changes to keep the same style even though there is no compilation error.

azuchi (approved):

PR #272, opened 2023-07-07
Upgrade leveldb to 1.23 release version.

Naviabheeman:
Update leveldb to its latest release version 1.23. bitcoin is at version 1.22. 
makefiles are updated to build the new version. The source code in downloaded as is from [leveldb](https://github.com/google/leveldb) repository.

Data format compatibility was tested by creating a wallet in tapyrus core version 0.5.2 and upgrading it. i.e the binaries built after this change was used to load the old wallet. wallet balance, utxo list were the same as before the upgrade.


azuchi (changes requested):
Is it also necessary to change the configure.ac below?

https://github.com/bitcoin/bitcoin/pull/23282/files

Also, `Makefile.leveldb.include` seems to be different from the bitcoin's one.

Naviabheeman:
> Is it also necessary to change the configure.ac below?
> 
> https://github.com/bitcoin/bitcoin/pull/23282/files

Yes, we can port this.  It moves all leveldb settings into one file. 

> 
> Also, `Makefile.leveldb.include` seems to be different from the bitcoin's one.

Bitcoin is using static compilation, a new library for crc232 and  some new flags for building on other operating systems. Our builds don't need them.

azuchi (approved):

PR #276, opened 2023-08-02
Updates in  synchronization mutex for Shared data

Naviabheeman:
Changed CCriticalSections to Mutex and RecursiveMutex as in Bitcoin. 

Add GUARDED_BY and EXCLUSIVE_LOCK_REQUIRED annotations to some shared data only. 
  Adding them to all shared data and functions caused lots of build errors on macOS. Clang supports static analysis for thread safety with option -thread-safety-analysis. But this does not seem to be strong enough to identify scope. i.e. when lock is outside a class or a function where the data is accessed. As most of our data and locks are global, and locking is in the outer scope, it causes lots of false positives. Adding more LOCKS in the same scope to pass the compilation warning causes deadlock in our tests.

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -373,8 +373,8 @@ private:
     static bool NodeFullyConnected(const CNode* pnode);
 
     // Network usage totals
-    CCriticalSection cs_totalBytesRecv;
-    CCriticalSection cs_totalBytesSent;
+    Mutex cs_totalBytesRecv;
+    Mutex cs_totalBytesSent;
     uint64_t nTotalBytesRecv GUARDED_BY(cs_totalBytesRecv);
     uint64_t nTotalBytesSent GUARDED_BY(cs_totalBytesSent);
 
@@ -394,17 +394,17 @@ private:
     std::vector<ListenSocket> vhListenSocket;
     std::atomic<bool> fNetworkActive;
     banmap_t setBanned;
-    CCriticalSection cs_setBanned;
+    Mutex cs_setBanned;
     bool setBannedIsDirty;
     bool fAddressesInitialized;
     CAddrMan addrman;
     std::deque<std::string> vOneShots;
-    CCriticalSection cs_vOneShots;
+    Mutex cs_vOneShots;
     std::vector<std::string> vAddedNodes GUARDED_BY(cs_vAddedNodes);
-    CCriticalSection cs_vAddedNodes;
+    Mutex cs_vAddedNodes;
     std::vector<CNode*> vNodes;
     std::list<CNode*> vNodesDisconnected;
-    mutable CCriticalSection cs_vNodes;
+    mutable RecursiveMutex cs_vNodes;
     std::atomic<NodeId> nLastNodeId;
 
     /** Services this instance offers */
@@ -427,7 +427,7 @@ private:
     bool fMsgProcWake;
 
     std::condition_variable condMsgProc;
-    std::mutex mutexMsgProc;
+    Mutex mutexMsgProc;
     std::atomic<bool> flagInterruptMsgProc;
 
     CThreadInterrupt interruptNet;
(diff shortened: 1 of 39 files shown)


src/net.h
```diff
@@ -373,8 +373,8 @@ class CConnman
     static bool NodeFullyConnected(const CNode* pnode);
 
     // Network usage totals
-    CCriticalSection cs_totalBytesRecv;
-    CCriticalSection cs_totalBytesSent;
+    Mutex cs_totalBytesRecv;
+    Mutex cs_totalBytesSent;
```
azuchi, line 377: In Bitcoin, RecursiveMutex is used.

src/net.h
```diff
@@ -625,15 +625,15 @@ class CNode
     size_t nSendOffset; // offset inside the first vSendMsg already sent
     uint64_t nSendBytes;
     std::deque<std::vector<unsigned char>> vSendMsg;
-    CCriticalSection cs_vSend;
-    CCriticalSection cs_hSocket;
-    CCriticalSection cs_vRecv;
+    RecursiveMutex cs_vSend;
+    RecursiveMutex cs_hSocket;
+    RecursiveMutex cs_vRecv;
```
azuchi, line 630: In Bitcoin, Mutex is used.

src/net.h
```diff
@@ -394,17 +394,17 @@ class CConnman
...
+    RecursiveMutex cs_setBanned;
     bool setBannedIsDirty;
     bool fAddressesInitialized;
     CAddrMan addrman;
     std::deque<std::string> vOneShots;
-    CCriticalSection cs_vOneShots;
+    Mutex cs_vOneShots;
     std::vector<std::string> vAddedNodes GUARDED_BY(cs_vAddedNodes);
-    CCriticalSection cs_vAddedNodes;
+    Mutex cs_vAddedNodes;
```
azuchi, line 404: In Bitcoin, RecursiveMutex is used.

src/net.h
```diff
@@ -394,17 +394,17 @@ class CConnman
...
     std::atomic<bool> fNetworkActive;
     banmap_t setBanned;
-    CCriticalSection cs_setBanned;
+    RecursiveMutex cs_setBanned;
     bool setBannedIsDirty;
     bool fAddressesInitialized;
     CAddrMan addrman;
     std::deque<std::string> vOneShots;
-    CCriticalSection cs_vOneShots;
+    Mutex cs_vOneShots;
```
azuchi, line 402: In Bitcoin, RecursiveMutex is used.

src/net.h
```diff
@@ -530,7 +530,7 @@ struct LocalServiceInfo {
     int nPort;
 };
 
-extern CCriticalSection cs_mapLocalHost;
+extern Mutex cs_mapLocalHost;
```
azuchi, line 533: In Bitcoin, RecursiveMutex is used.

src/net.h
Naviabheeman: they have only one mutex. other is changed to atomic 
```    
    // Network usage totals
    mutable Mutex m_total_bytes_sent_mutex;
    std::atomic<uint64_t> nTotalBytesRecv{0};
    ```

File: src/net.h (hunk 4 above)
Comment (Naviabheeman) [reply src/net.h:402]: This is completely removed actually. Bitcoin has reimplemented the peer management layer.

File: src/net.h (hunk 3 above)
Comment (Naviabheeman) [reply src/net.h:404]: this remains the same:
```
mutable Mutex m_added_nodes_mutex;
```

File: src/net.h (hunk 5 above)
Comment (Naviabheeman) [reply src/net.h:533]:
```
extern GlobalMutex g_maplocalhost_mutex;
```
you mean they say globalmutex which is actually
```
class GlobalMutex : public Mutex { };
```

azuchi (approved):

PR #281, opened 2023-08-20
Tracepoints for mempool in tapyrus

Naviabheeman:
Port mempool add, remove, replace and rejected UDST tracepoints from bitcoin

### DIFF
--- a/contrib/tracing/mempool_monitor.py
+++ b/contrib/tracing/mempool_monitor.py
@@ -27,7 +27,7 @@ PROGRAM = """
 struct added_event
 {
   u8    hash[HASH_LENGTH];
-  u64   vsize;
+  s32   vsize;
   s64   fee;
 };
 
@@ -35,7 +35,7 @@ struct removed_event
 {
   u8    hash[HASH_LENGTH];
   char  reason[MAX_REMOVAL_REASON_LENGTH];
-  u64   vsize;
+  s32   vsize;
   s64   fee;
   u64   entry_time;
 };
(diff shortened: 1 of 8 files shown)


contrib/tracing/mempool_monitor.py
```diff
@@ -0,0 +1,372 @@
...
+// strprintf("mandatory-script-verify-flag-failed (%s)", ScriptErrorString(check.GetScriptError()))
+#define MAX_REJECT_REASON_LENGTH        118
+// The longest string returned by RemovalReasonToString() is 'sizelimit'
+#define MAX_REMOVAL_REASON_LENGTH       9
+#define HASH_LENGTH                     32
+
+struct added_event
+{
+  u8    hash[HASH_LENGTH];
+  u64   vsize;
```
azuchi, line 30: It would be better to also port the following PR:
https://github.com/bitcoin/bitcoin/pull/23962

azuchi (changes requested):
Similar to Bitcoin, the following documentation needs to be updated:

* contrib/tracing/README.md
* doc/tracing.md

contrib/tracing/mempool_monitor.py
Naviabheeman: Done!
We removed tx weight related code from tapyrus. So only the size calculation within mempool needed to be updated.

Naviabheeman:
> Similar to Bitcoin, the following documentation needs to be updated:
> 
> * contrib/tracing/README.md
> * doc/tracing.md

done!

azuchi (approved):

PR #291, opened 2023-12-31
Refactor only - simplify AcceptToMempool argument list

Naviabheeman:
refactoring: 
remove argument mempool as it is a global
add test and bypass limit flags to enum MempoolAcceptanceFlags
add state and all other outputs in the same struct CTxMempoolAcceptanceOptions
make smaller function calls for some of the checks in AcceptToMempoolWorker

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -438,9 +438,6 @@ bool CheckSequenceLocks(const CTransaction &tx, int flags, LockPoints* lp, bool 
     return EvaluateSequenceLocks(index, lockPair);
 }
 
-// Returns the script flags which should be checked for a given block
-static unsigned int GetBlockScriptFlags(const CBlockIndex* pindex);
-
 static void LimitMempoolSize(CTxMemPool& pool, size_t limit, unsigned long age) {
     int expired = pool.Expire(GetTime() - age);
     if (expired != 0) {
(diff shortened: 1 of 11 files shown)


src/validation.cpp
```diff
@@ -1009,64 +1075,17 @@ static bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state,
         // There is a similar check in CreateNewBlock() to prevent creating
         // invalid blocks (using TestBlockValidity), however allowing such
         // transactions into the mempool can be exploited as a DoS attack.
-        unsigned int currentBlockScriptVerifyFlags = GetBlockScriptFlags(chainActive.Tip());
-        if (!CheckInputsFromMempoolAndCache(tx, state, view, pool, currentBlockScriptVerifyFlags, true, txdata)) {
+        if (!CheckInputsFromMempoolAndCache(tx, state, view, pool, SCRIPT_VERIFY_NONE, true, txdata)) {
```
azuchi, line 1078: Why did you remove `currentBlockScriptVerifyFlags`?

src/validation.cpp
```diff
@@ -1088,8 +1107,8 @@ static bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state,
                 tx.GetTotalSize(),
                 nConflictingFees
             );
-            if (plTxnReplaced)
```
azuchi, line 1091: Why this condition removed?

src/validation.cpp
Naviabheeman: It looked redundant because there are no real flags now. I'm undoing this as you advice.

src/validation.cpp
Naviabheeman: Earlier it was a pointer and optional. So the check ensured that null pointer is not dereferenced. Now it is a vector inside 'opt' which is passed around as a reference. So we can safely access its contents as they are always initialized.

azuchi (approved):

Issue #289, opened 2023-12-18
Fix hash_serialized_2 calculation


azuchi:
Porting the following Bitcoin Core fixe:

https://github.com/bitcoin/bitcoin/pull/28685

Naviabheeman:
I logged the output of a stream after pushing these expressions:
```
VARINT(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase ? 1u : 0u);
VARINT(outputs.begin()->second.nHeight * 2 + (outputs.begin()->second.fCoinBase ? 1u : 0u));
VARINT((outputs.begin()->second.nHeight << 1) + (outputs.begin()->second.fCoinBase ? 1u : 0u));
VARINT((outputs.begin()->second.nHeight << 1)  + outputs.begin()->second.fCoinBase ? 1u : 0u);
uint32_t(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase ? 1u : 0u);
uint32_t(outputs.begin()->second.nHeight * 2 + (outputs.begin()->second.fCoinBase ? 1u : 0u));
 ```
The log also has the height. For the same height, each one produces a different output.   **Varint[856b]** is the expected output for VARINT encoding.

```
ApplyStats: [437], Varint[01] 
ApplyStats: [437], Varint[856b] 
ApplyStats: [437], Varint[856b] 
ApplyStats: [437], Varint[01] 
ApplyStats: [437], Varint[31]
ApplyStats: [437], Varint[383735]
 ```

This bug is fixed by using a parenthesis around _'second.fCoinBase'_ expression like in _TxInUndoSerializer_. It does not matter whether we use ```* 2``` or ```<< 1 ```. 
Bitcoin moved away from VARINT and used ```static_cast<uint32_t>```. I think we can keep VARINT as it gives the most compact encoding.

Naviabheeman:
`feature_dbcrash.py` with output while logging all the above VARINT hashes:

```
2024-01-29T12:28:16.013000Z TestFramework (INFO): Verifying utxo hash matches for all nodes: 9e64e045679e77ce153a58a446e55651bf34b0efaa4659ae5722694fcf42cb3c
2024-01-29T12:28:16.588000Z TestFramework (INFO): Verifying utxo hash matches for all nodes: eea17067712df5369cbd5f5308232797a57667a9a4d522ee176a2f06c2f31f0d
[...]

PR #293, opened 2024-01-29
Fix hash serialized calculation 

Naviabheeman:
Fix issue #289. 

This bug is fixed by using a parenthesis around _'second.fCoinBase'_ expression like in _TxInUndoSerializer_. It does not matter whether we use `* 2` or `<< 1 `. 

- Bitcoin moved away from VARINT and used `static_cast<uint32_t>`. I think we can keep VARINT as it gives the most compact encoding.
- Like Bitcoin `hash_serialized_2` is changed to `hash_serialized_3` to differentiate from the old hash output.

 For a coin with height **437**, the stream contents after pushing each expressions wasps follows:
 
| expression | output |
|--|--|
| VARINT(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase ? 1u : 0u); |01 |
| VARINT(outputs.begin()->second.nHeight * 2 + (outputs.begin()->second.fCoinBase ? 1u : 0u)); | 856b|
| VARINT((outputs.begin()->second.nHeight << 1) + (outputs.begin()->second.fCoinBase ? 1u : 0u)); | 856b |
| VARINT((outputs.begin()->second.nHeight << 1)  + outputs.begin()->second.fCoinBase ? 1u : 0u); | 01 |
| uint32_t(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase ? 1u : 0u); | 31 |
| uint32_t(outputs.begin()->second.nHeight * 2 + (outputs.begin()->second.fCoinBase ? 1u : 0u)); | 383735 |

### DIFF
--- a/test/functional/feature_dbcrash.py
+++ b/test/functional/feature_dbcrash.py
@@ -81,7 +81,7 @@ class ChainstateWriteCrashTest(BitcoinTestFramework):
                 # Any of these RPC calls could throw due to node crash
                 self.start_node(node_index)
                 self.nodes[node_index].waitforblock(expected_tip)
-                utxo_hash = self.nodes[node_index].gettxoutsetinfo()['hash_serialized_2']
+                utxo_hash = self.nodes[node_index].gettxoutsetinfo()['hash_serialized_3']
                 return utxo_hash
             except:
                 # An exception here should mean the node is about to crash.
@@ -133,7 +133,7 @@ class ChainstateWriteCrashTest(BitcoinTestFramework):
         If any nodes crash while updating, we'll compare utxo hashes to
         ensure recovery was successful."""
 
-        node3_utxo_hash = self.nodes[3].gettxoutsetinfo()['hash_serialized_2']
+        node3_utxo_hash = self.nodes[3].gettxoutsetinfo()['hash_serialized_3']
 
         # Retrieve all the blocks from node3
         blocks = []
@@ -175,12 +175,12 @@ class ChainstateWriteCrashTest(BitcoinTestFramework):
         """Verify that the utxo hash of each node matches node3.
 
         Restart any nodes that crash while querying."""
-        node3_utxo_hash = self.nodes[3].gettxoutsetinfo()['hash_serialized_2']
-        self.log.info("Verifying utxo hash matches for all nodes")
+        node3_utxo_hash = self.nodes[3].gettxoutsetinfo()['hash_serialized_3']
+        self.log.info("Verifying utxo hash matches for all nodes: " +  node3_utxo_hash)
 
         for i in range(3):
             try:
-                nodei_utxo_hash = self.nodes[i].gettxoutsetinfo()['hash_serialized_2']
+                nodei_utxo_hash = self.nodes[i].gettxoutsetinfo()['hash_serialized_3']
             except OSError:
                 # probably a crash on db flushing
                 nodei_utxo_hash = self.restart_node(i, self.nodes[3].getbestblockhash())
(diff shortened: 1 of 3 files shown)


test/functional/feature_dbcrash.py
```diff
@@ -61,7 +61,7 @@ def set_test_params(self):
         self.node2_args = ["-dbcrashratio=24", "-dbcache=16"] + self.base_args
 
         # Node3 is a normal node with default args, except will mine full blocks
-        self.node3_args = ["-blockmaxweight=4000000"]
+        self.node3_args = ["-blockmaxsize=1000000"]
```
azuchi, line 64: Why this change needs?

test/functional/feature_fee_estimation.py
```diff
@@ -134,8 +134,8 @@ def setup_network(self):
         which we will use to generate our transactions.
         """
         self.add_nodes(3, extra_args=[["-maxorphantx=1000", "-whitelist=127.0.0.1"],
-                                      ["-blockmaxweight=68000", "-maxorphantx=1000"],
-                                      ["-blockmaxweight=32000", "-maxorphantx=1000"]])
+                                      ["-blockmaxsize=17000", "-maxorphantx=1000"],
+                                      ["-blockmaxsize=8000", "-maxorphantx=1000"]])
         # Use node0 to mine blocks for input splitting
```
azuchi, line 139: Why this change needs?

test/functional/feature_pruning.py
```diff
@@ -127,7 +127,7 @@ def reorg_test(self):
         # Reboot node 1 to clear its mempool (hopefully make the invalidate faster)
         # Lower the block max size so we don't keep mining all our big mempool transactions (from disconnected blocks)
         self.stop_node(1)
-        self.start_node(1, extra_args=["-maxreceivebuffer=20000","-blockmaxweight=20000", "-checkblocks=5"])
+        self.start_node(1, extra_args=["-maxreceivebuffer=20000","-blockmaxsize=5000", "-checkblocks=5"])
```
azuchi, line 130: Why this change needs?

test/functional/feature_dbcrash.py
```diff
@@ -218,7 +218,7 @@ def run_test(self):
 
         # Start by creating a lot of utxos on node3
         initial_height = self.nodes[3].getblockcount()
-        utxo_list = create_confirmed_utxos(self.nodes[3].getnetworkinfo()['relayfee'], self.nodes[3], 5000, self.signblockprivkey)
+        utxo_list = create_confirmed_utxos(self.nodes[3].getnetworkinfo()['relayfee'], self.nodes[3], 5000, self.signblockprivkey_wif)
```
azuchi, line 221: Why this change needs?

test/functional/feature_pruning.py
Naviabheeman: We removed ```blockmaxweight``` argument from tapyrus. But these tests hadn't been updated. They are some long running scripts listed as ```EXTENDED_SCRIPTS```. These are not run by default by the CI. It has to be enabled in ```configure``` So I did not notice them until now. 
I wanted this change to be a separate PR after checking the running time. But had added them to PR unintentionally.

test/functional/feature_dbcrash.py
Naviabheeman: This change should have been done earlier when all 'generate' rpcs were changed.  
 They are some long running scripts listed as EXTENDED_SCRIPTS. These are not run by default by the CI. It has to be enabled in configure So I did not notice them until now.
I wanted this change to be a separate PR after checking the running time. But I had to add them to this PR as this test needs it to execute correctly.

test/functional/feature_pruning.py
azuchi: Since multiple changes should not be included in one PR, please delete it from this PR and create new PR.

test/functional/feature_dbcrash.py
azuchi: Create another PR as above.

Naviabheeman: This PR is now clean i.e. only hash serialized bug fix and tis tests are present. Test case 'feature_dbcrash.py' is not executed in the CI now.
I am in the process of fixing the extended functional tests. This will be in a separate PR. After that one is merged all python tests will run in the CI every time.

azuchi (approved):

PR #296, opened 2024-02-28
Upgrade CI node in v0.5.2

Naviabheeman:
_Problem:_
CI running on ubuntu 16 is not supported in github anymore. So ubuntu20 was needed. With that upgrade some of the dependencies from `depends` directory were not available/failed to build. They needed upgraded. MacOS SDK needed upgrade. 
_Fix:_
So the complete build environment is upgraded to match the master branch. IT includes:
  - ci.yml
  - build-aux/m4 - all configuration files
  - depends/* - all packages, patches and base scripts
  - configure script and Makefiles - for link errors
  - c++ files which failed to build in the new environment
  - python tests


depends/packages/packages.mk
```diff
@@ -1,18 +1,23 @@
...
 qt_darwin_packages=qt
+
 qt_mingw32_packages=qt
 
 wallet_packages=bdb
 
 upnp_packages=miniupnpc
 
-darwin_native_packages = native_biplist native_ds_store native_mac_alias
+usdt_linux_packages=systemtap
```
azuchi, line 15: This is necessary for USDT support and is not required for the 0.5 branch.

azuchi (changes requested):

Naviabheeman:
dropping this PR as #299 can achieve this with minimum changes

PR #298, opened 2024-03-01
Refactor files

Naviabheeman:
This PR is only refactoring of validation.cpp file. There is no code modification, it is only code move. This PR does not change behaviour. No new tests are added. 

New files:

1. chainstate.h and chainstate.cpp
 - these two files now contain all the chain state class and global chain state, g_chainstate.
2.  cs_main.h and cs_main.cpp
- This is similar to bitcoin. These files contains global mutex cs_main and cs_LastBlockFile. This makes accessing. these two critical sections from multiple files possible.
3. file_io.h and file_io.cpp
- All disk operations are moved to these new files
4. scriptcheck.h
- contains CScriptCheck class
5. verifydb.h and verifydb.cpp
- contains CVerifyDB class
6. blockprune.h and blockprune.cpp
- global functions that are related to block pruning are in these files


azuchi (changes requested):

azuchi (changes requested):
Please fix following warnings:

```
blockprune.cpp:108:36: warning: comparison of integer expressions of different signedness: ‘int’ and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
  108 |     if (chainActive.Tip()->nHeight <= nPruneAfterHeight) {
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
...

chainstate.cpp:611:22: warning: unused variable ‘flags’ [-Wunused-variable]
  611 |         unsigned int flags = GetBlockScriptFlags(pindex);
      |                      ^~~~~
...

file_io.cpp:371:18: warning: unused variable ‘coins_count’ [-Wunused-variable]
  371 |     const size_t coins_count = pcoinsTip->GetCacheSize();
      |                  ^~~~~~~~~~~
file_io.cpp:372:18: warning: unused variable ‘coins_mem_usage’ [-Wunused-variable]
  372 |     const size_t coins_mem_usage = pcoinsTip->DynamicMemoryUsage();
      |                  ^~~~~~~~~~~~~~~
...

 2365 |         CValidationState& state = opt.state;
      |                           ^~~~~
...

```

cmake/modules/FindUPNP.cmake
```diff
@@ -2,13 +2,15 @@
 # UPNP_FOUND - system has ZeroMQ lib
 # UPNP_INCLUDE_DIR - the ZeroMQ include directory
 # UPNP_LIBRARY - Libraries needed to use ZeroMQ
+include(BrewHelper)
+find_brew_prefix(BREW_HINT_UPNP miniupnpc)
```
azuchi, line 6: Write homebrew-related code in a code block that has been confirmed to be for macOS like bellow:

https://github.com/bitcoin/bitcoin/blob/master/cmake/module/FindBerkeleyDB.cmake#L31-L33

Please modify FindBerkeleyDB.cmake in the same way.

test/functional/test_framework/script.py
```diff
@@ -18,6 +18,7 @@
 
 MAX_SCRIPT_SIZE = 10000
 MAX_SCRIPT_ELEMENT_SIZE = 520
+MAX_SCRIPT_SIZE = 10000
```
azuchi, line 21: This change unnecessary, because duplicate L19.

src/consensus/validation.h
```diff
@@ -93,6 +93,20 @@ class CValidationState {
     unsigned int GetRejectCode() const { return chRejectCode; }
     std::string GetRejectReason() const { return strRejectReason; }
     std::string GetDebugMessage() const { return strDebugMessage; }
+    std::string Describe() const {
```
azuchi, line 96: This method is not used?

src/leveldb/db/db_impl.cc
```diff
@@ -8,6 +8,7 @@
 #include <atomic>
 #include <cstdint>
 #include <cstdio>
+#include <deque>
```
azuchi, line 11: Why this include need?

src/policy/packages.cpp
```diff
@@ -102,7 +102,7 @@ bool SubmitPackageToMempool(const Package& package,
                                   PackageValidationState& results,
                                   CTxMempoolAcceptanceOptions& opt)
 {
-    if(!CheckPackage(package, state))
+   if(!CheckPackage(package, state))
```
azuchi, line 105: This change not necessary.

src/cs_main.h
```diff
@@ -0,0 +1,24 @@
...
+ * or changing the chainstate.
+ *
+ * This may also need to be locked when updating the transaction pool, e.g. on
+ * AcceptToMemoryPool. See CTxMemPool::cs comment for details.
+ *
+ * The transaction pool has a separate lock to allow reading from it and the
+ * chainstate at the same time.
+ */
+extern RecursiveMutex cs_main;
+extern RecursiveMutex cs_LastBlockFile;
```
azuchi, line 22: In Bitcoin, it seems there is only cs_main, but what is cs_LastBlockFile for?

src/file_io.cpp
```diff
@@ -0,0 +1,620 @@
...
+#include <shutdown.h>
+#include <trace.h>
+#include <blockprune.h>
+#include <validation.h>
+#include <file_io.h>
+#include <deque>
+
+static const uint64_t MEMPOOL_DUMP_VERSION = 1;
+
+bool LoadMempool(void)
```
azuchi, line 15:
```suggestion
bool LoadMempool()
```

src/cs_main.h
Naviabheeman: This is another global mutex which is shared between multiple files. So it is needed in a common place. In bitcoin, this is moved inside a new class called block manager.

cmake/modules/FindBerkeleyDB.cmake
```diff
@@ -1,37 +1,110 @@
...
-# BDB_FOUND - system has Berkeley DB lib
-# BDB_INCLUDE_DIR - the Berkeley DB include directory
-# BDB_LIBRARY - Library needed to use Berkeley DB
-# BDBXX_INCLUDE_DIR - the Berkeley DB include directory for C++
-# BDBXX_LIBRARY - Library needed to use Berkeley DB C++ API
+# BerkeleyDB_FOUND - system has Berkeley DB lib
+# BerkeleyDB_INCLUDE_DIR - the Berkeley DB include directory
+# BerkeleyDB_LIBRARY - Library needed to use Berkeley DB
 
 include(BrewHelper)
```
azuchi, line 6: Shouldn't this include also be executed only when CMAKE_HOST_APPLE is specified?

azuchi (approved):

Issue #290, opened 2023-12-18
Add UTXO snapshot creation


azuchi:
Introducing the dumptxoutset RPC added to Bitcoin Core.

https://github.com/bitcoin/bitcoin/pull/16899

azuchi:
> mode - prod/dev

This would be unnecessary. Even in Bitcoin, snapshots do not include data that distinguishes between mainnet and testnet.

> snapshot_time

`Since `base_blockhash` is present, this is also unnecessary.

Naviabheeman:
In PR #29612 from bitcoin-core ( Optimize serialization and enhance metadata of dumptxoutset output) utxo snapshot metadata is enhanced to include
1.  version
2.  network magic bytes
3.  block height


In Tapyrus version is added.
Network magic bytes depend on network id. So we use network mode instead.
Block height can be deduced from block hash. This is not added.

PR #308, opened 2024-06-12
UTXO snapshot creation in Tapyrus

Naviabheeman:
This PR ports '`dumptxoutset`' rpc from bitcoin core. It is renamed as `utxosnapshot` in Tapyrus.
`SnapshotMetadata` class in Tapyrus includes network Id and mode. Other data such as version, Bloch hash and coin count are the same as bitcoin. But block height is not included as it can be deduced from the hash.

Functional test `rpc_dumptxoutset.py` is ported as `rpc_utxosnapshot.py` and enhanced to verify the snapshot metadata


test/functional/rpc_utxosnapshot.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
```
azuchi, line 1: Let's name the file `rpc_dumptxoutset.py` to match Bitcoin.

azuchi (changes requested):
`contrib/devtools/utxo_snapsho.sh` also need.

test/functional/rpc_utxosnapshot.py
Naviabheeman: Do you want to rename the rpc also? This file is named `rpc_utxosnapshot.py` to match the rpc name, `utxosnapshot`. In bitcoin both are `dumptxoutset`.

Naviabheeman:
> `contrib/devtools/utxo_snapsho.sh` also need.

This script allows us to generate the snapshot at any height. It does this by invalidating all blocks after the given height. This looks risky to me in Tapyrus as we do not support multiple chain states. We may rewind to an old block and lose everything after that point just to take the snapshot. Then we'll redownload everything all over again. 
Should we consider implementing one of these :
 1. prepare a copy of the utxoset only and rewind it. 
 2. follow bitcoin and create a copy chain state to rewind the blockchain.

test/functional/rpc_utxosnapshot.py
azuchi: Yes, let's make them both the same.

azuchi:
> This looks risky to me in Tapyrus as we do not support multiple chain states. 

Why is this rewinding a risk in Tapyrus?

> prepare a copy of the utxoset only and rewind it.

Please tell me the specific process for this.

Naviabheeman:
> Why is this rewinding a risk in Tapyrus?

I was talking about the time lost in creating the snapshot as the node is not available on the network for that long. We can disrupt the network by doing it on multiple nodes. When the node is online again it's like restarting the blockchain at an older point. New transactions and blocks may be created. We do not have an alternate the 'live' chainstate db like bitcoin to verify whether the catchup is to the correct blockchain tip.

> Please tell me the specific process for this.

It can be an independent thread like reindex. Get only the utxos from level db and put them into a temporary db. Use the undo blocks to remove utxos until we reach the needed block height. This is just a crude idea. It may be more complex to implement.

azuchi:
@Naviabheeman This tool and the `invalidateblock` RPC are intended for developers and are not intended for production use. Rather than writing special purpose logic, let's take the same approach as bitcoin.

src/rpc/blockchain.cpp
```diff
@@ -2126,6 +2128,150 @@ static UniValue getcolor(const JSONRPCRequest& request)
...
+    fs::rename(temppath, path);
+    const CBlockIndex* tip = LookupBlockIndex(maybe_stats.hashBlock);
+
+    UniValue result(UniValue::VOBJ);
+    result.pushKV("coins_written", maybe_stats.nTransactionOutputs);
+    result.pushKV("base_hash", tip->GetBlockHash().ToString());
+    result.pushKV("base_height", tip->nHeight);
+    result.pushKV("path", path.c_str());
+    result.pushKV("txoutset_hash", maybe_stats.hashSerialized.ToString());
+    result.pushKV("nchaintx", tip->nChainTx);
```
azuchi: The `versionO and `network_id` mentioned in the [issue](https://github.com/chaintope/tapyrus-core/issues/290) are missing.

src/rpc/blockchain.cpp
```diff
@@ -2126,6 +2128,152 @@ static UniValue getcolor(const JSONRPCRequest& request)
...
+        afile << last_hash;
+        WriteCompactSize(afile, coins.size());
+        for (const auto& [n, coin] : coins) {
+            WriteCompactSize(afile, n);
+            afile << coin;
+            ++written_coins_count;
+        }
+    };
+    while (pcursor->Valid()) {
+        if (iter % 5000 == 0) ShutdownRequested();
```
azuchi, line 2199: Why shut it down?

Naviabheeman: This is the older version of checking for interrupts. In the bitcoin implementation there is a newer technique, `RpcInterruptionPoint`. In Tapyrus, we are checking whether to shutdown using `ShutdownRequested` at regular intervals.

src/rpc/blockchain.cpp
Naviabheeman: These are part of the shapshot metadata written to the snapshot file. It is not in the RPC output. Test case `test/functional/rpc_dumptxoutset.py` verifies that these are written to the file by the RPC. IT then reads them from the file, deserialises and verifies that they match the expected output.

azuchi (approved):

PR #315, opened 2024-09-24
Security fix 4 - Simplify node time and network-adjusted time warning logic

Naviabheeman:
Port fix of vulnerability fix of [Disclosure of netsplit due to timestamp adjustment](https://bitcoincore.org/en/2024/07/03/disclose-timestamp-overflow/) which was changed later by PR 29623(Simplify network-adjusted time warning logic)

As this PR removes the `timeoffset` previously defined in each node, the notion of 'adjusted time' is also removed. So a unified `std::chrono::system_clock` is used to get the time in GetTime, GetAdjustedTime, GetSystemTimeInSeconds, GetTimeMillis and GetTimeMicros. 

- `GetAdjustedTime` and `GetTime` are now the same. They return the mockable system time. 
- `GetSystemTimeInSeconds`  does not allow mocktime. This is essential as the p2p layer depends heavily on the connection and message time. Any change to this time due to mocking results in unpredictable behaviour in functional tests.


src/CMakeLists.txt
```diff
@@ -203,7 +203,6 @@ add_library(server
         script/descriptor.cpp
         script/sigcache.cpp
         script/ismine.cpp
-        timedata.cpp
```
azuchi, line 206: Do you need to add `timeoffsets.cpp`?

azuchi (changes requested):
Isn't the `nTimeOffset` field in classes `CNode` and `CNodeStats` also unnecessary?

src/net_processing.cpp
```diff
@@ -23,6 +23,7 @@
 #include <random.h>
 #include <reverse_iterator.h>
 #include <scheduler.h>
+#include <timeoffsets.h>
```
azuchi, line 26: Are the following changes to Bitcoin unnecessary?

https://github.com/bitcoin/bitcoin/pull/29623/files#diff-6875de769e90cec84d2e8a9c1b962cdbcda44d870d42e4215827e599e11e90e3R1801

src/rpc/net.cpp
```diff
@@ -426,7 +429,6 @@ static UniValue getnetworkinfo(const JSONRPCRequest& request)
             "  \"protocolversion\": xxxxx,              (numeric) the protocol version\n"
             "  \"localservices\": \"xxxxxxxxxxxxxxxx\", (string) the services we offer to the network\n"
             "  \"localrelay\": true|false,              (bool) true if transaction relay is requested from peers\n"
-            "  \"timeoffset\": xxxxx,                   (numeric) the time offset\n"
```
azuchi, line 429: It doesn't seem to have been removed in Bitcoin.

src/test/timedata_tests.cpp
```diff
@@ -3,7 +3,6 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 //
-#include <timedata.h>
```
azuchi, line 6: Bitcoin has removed this test file itself. Why keep this?

test/functional/test_framework/mininode.py
```diff
@@ -299,6 +299,7 @@ def on_reject(self, message): pass
     def on_sendcmpct(self, message): pass
     def on_sendheaders(self, message): pass
     def on_tx(self, message): pass
+    def on_notfound(self, message): pass
```
azuchi, line 302: Is this change related to this PR?

azuchi (changes requested):

Naviabheeman:
> Isn't the `nTimeOffset` field in classes `CNode` and `CNodeStats` also unnecessary?

`nTimeOffset` field  in `CNode` class has the time offset for each peer. I was not able to remove it.

src/net_processing.cpp
```diff
@@ -1863,9 +1864,13 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
                   pfrom->nStartingHeight, addrMe.ToString(), pfrom->GetId(),
                   remoteAddr);
 
-        int64_t nTimeOffset = nTime - GetTime();
-        pfrom->nTimeOffset = nTimeOffset;
-        AddTimeData(pfrom->addr, nTimeOffset);
+        pfrom->nTimeOffset = nTime - GetTime();
```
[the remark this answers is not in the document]
Naviabheeman: In Bitcoin there are lots of readable user friendly names. Each is actually just a typedef
`using NodeSeconds = std::chrono::time_point<NodeClock, std::chrono::seconds>;`
In our code the long qualified names are used. And the interface is not changed i.e. GetTime(), GetSystemTime() etc are retained.

src/qt/rpcconsole.cpp
```diff
@@ -1120,7 +1120,6 @@ void RPCConsole::updateNodeDetail(const CNodeCombinedStats *stats)
     ui->peerPingTime->setText(GUIUtil::formatPingTime(stats->nodeStats.dPingTime));
     ui->peerPingWait->setText(GUIUtil::formatPingTime(stats->nodeStats.dPingWait));
     ui->peerMinPing->setText(GUIUtil::formatPingTime(stats->nodeStats.dMinPing));
-    ui->timeoffset->setText(GUIUtil::formatTimeOffset(stats->nodeStats.nTimeOffset));
     ui->peerVersion->setText(QString("%1").arg(QString::number(stats->nodeStats.nVersion)));
```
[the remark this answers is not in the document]
Naviabheeman: It should be there. I have missed it.

src/net_processing.cpp
Naviabheeman: 'CNodeStateStats' has more information in Bitcoin than in Tapyrus. It was changed while refactoring for Peer manager class. In out code time offset is still in 'CNodeStats'

src/rpc/net.cpp
Naviabheeman: Yes, this is a little different now. Earlier, each node maintains its own time offset from to the median block time. So this information was available globally and reported in network info. Now it is maintained by the node for each peer and reported only in peerinfo rpc.

test/functional/test_framework/mininode.py
Naviabheeman: no, this code is not necessary. will remove it.

azuchi:
> > Isn't the `nTimeOffset` field in classes `CNode` and `CNodeStats` also unnecessary?
> 
> `nTimeOffset` field in `CNode` class has the time offset for each peer. I was not able to remove it.

but, `nTimeOffset` in `CNodeStats` seems not referenced anywhere.

Naviabheeman:
> but, `nTimeOffset` in `CNodeStats` seems not referenced anywhere.

oh! yes I'm removing it.

azuchi (approved):

PR #318, opened 2024-11-03
Security fix6 getblocktxn

Naviabheeman:
Port fix for hindered block propagation due to mutated blocks, stalling peers and crash in compact block CVE-2024-35202.
PRs:
 #27626 - Parallel compact block downloads
 #26898 - Add PartiallyDownloadedBlock target 
 #27608 - Avoid prematurely clearing download state for other peer

- 'mapBlocksInFlight' is changed into a multi map( like bitcoin) to remember the order of compact block requests made to peers.
- global variables used to support witness and multiple compact block version are removed
- 'nBlocksInFlight' and 'nBlocksInFlightValidHeaders' are removed from NodeState(like bitcoin) as both are the same and can be deciphered from 'vBlocksInFlight'

 in bitcoin
- 'MarkBlockAsReceived'  is renamed to  'RemoveBlockRequest'. 
- 'MarkBlockAsInFlight' is renamed to 'BlockREquested'

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -444,7 +444,7 @@ static bool MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const CBlock
     auto itInFlight = mapBlocksInFlight.insert(std::make_pair(hash, std::make_pair(nodeid, it)));
     if (pit)
         *pit = &itInFlight->second.second;
-     return true;
+    return true;
 }
 
 /** Check whether the last unknown block a peer advertised is not yet known. */
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -157,7 +157,8 @@ namespace {
         bool fValidatedHeaders;                                  //!< Whether this block has validated headers at the time of request.
         std::unique_ptr<PartiallyDownloadedBlock> partialBlock;  //!< Optional, used for CMPCTBLOCK downloads
     };
-    std::map<uint256, std::pair<NodeId, std::list<QueuedBlock>::iterator> > mapBlocksInFlight GUARDED_BY(cs_main);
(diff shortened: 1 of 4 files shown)


src/net_processing.cpp
```diff
@@ -395,33 +418,33 @@ static bool MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const CBlock
...
-    }
-    if (state->nBlocksInFlightValidHeaders == 1 && pindex != nullptr) {
         nPeersWithValidatedDownloads++;
     }
-    itInFlight = mapBlocksInFlight.insert(std::make_pair(hash, std::make_pair(nodeid, it))).first;
+    auto itInFlight = mapBlocksInFlight.insert(std::make_pair(hash, std::make_pair(nodeid, it)));
     if (pit)
         *pit = &itInFlight->second.second;
-    return true;
+     return true;
```
azuchi, line 447: This change is not necessary.

azuchi (changes requested):

azuchi:
Is there any testing for the fix in #26898?

Naviabheeman:
> Is there any testing for the fix in #26898?

The changes in  _#26898_ were made to fix the crash I mentioned. Before this change, when the same block was received from multiple peers, '_tapyrusd_' would exit at the assertions as the header was a `nullptr`. The null pointer is fixed by _#27626_. It makes sure that the header is always initialised. 
So I think the tests in `p2p_compactblocks.py`should be enough. Old tests verify compact block reconstruction with missing transactions. The new scenario verifies the full flow by receiving compactblocks (for a new block) from more than 3 peers . Block reconstruction from compact block and node accepting the compact block/txn only from the peer which was requested are both verified.

azuchi (approved):

PR #321, opened 2024-11-24
Fix compiler warnings for 0.6.0 release

Naviabheeman:
Major changes include:

- remove C_89 requirement from autogen.sh
- configure: make Qt compilation off by default
-  move seed node list from federationparams to params and this list should be different for prod and test 
- fix boost::filesystem deprecated warnings in many functions and make it compatible with fs::filesystem change 
- port LockResult and  LocKDirectory for directory locking result from bitcoin, remove IsDirWritable and update its tests 
- remove memset usage in  colored coin and other functions that gave warning 
- change BadXFieldException to store the exception message and return a pointer to that string to fix a warning. 
- remove unused variables and functions across files


configure.ac
```diff
@@ -1,8 +1,8 @@
 dnl require autoconf 2.60 (AS_ECHO/AS_ECHO_N)
 AC_PREREQ([2.60])
 define(_CLIENT_VERSION_MAJOR, 0)
-define(_CLIENT_VERSION_MINOR, 5)
-define(_CLIENT_VERSION_REVISION, 2)
+define(_CLIENT_VERSION_MINOR, 6)
+define(_CLIENT_VERSION_REVISION, 0)
```
azuchi, line 5: Please include version changes in PR322 instead of here.

configure.ac
```diff
@@ -140,8 +140,8 @@ AC_ARG_ENABLE(tests,
 
 AC_ARG_ENABLE(gui-tests,
     AS_HELP_STRING([--disable-gui-tests],[do not compile GUI tests (default is to compile if GUI and tests enabled)]),
-    [use_gui_tests=$enableval],
-    [use_gui_tests=$use_tests])
+    [use_gui_tests=no],
+    [use_gui_tests=no])
```
azuchi, line 144: Why did you change it from a variable?

azuchi (changes requested):

configure.ac
Naviabheeman: This must have been removed sorry. I tried to disable GUI builds in configure script initially, then realised that CI had  the option.

build-aux/m4/tapyrus_qt.m4
```diff
@@ -54,7 +54,7 @@ AC_DEFUN([BITCOIN_QT_INIT],[
   dnl enable qt support
   AC_ARG_WITH([gui],
     [AS_HELP_STRING([--with-gui@<:@=no|qt5|auto@:>@],
-    [build tapyrus_qt GUI (default=auto)])],
+    [build tapyrus_qt GUI (default=no)])],
```
azuchi, line 57: Why change this?

configure.ac
```diff
@@ -1143,6 +1115,7 @@ else
       MINIUPNPC_CPPFLAGS="-DSTATICLIB -DMINIUPNP_STATICLIB"
     fi
   else
+    use_upnp=no
```
azuchi, line 1118: Why change this?

Naviabheeman: without this the final value of `use_upnp` was `auto`, the default value. But it started happening only after removing libc related changes. I didn't investigate further, just fixed it.

azuchi (approved):

PR #331, opened 2025-02-18
Cmake build Part 4 -  porting cmake scripts from bitcoin core

Naviabheeman:
Port all cmake scripts and configuration from bitcoin core. Only paths and file names are changed to tapyrus. All configuration are the same. This PR depends on #335, #336 and #337 and scep256k1 PP #10
Add CI using cmake for Ubuntu24 and macOS 14 compilation and tests.


.github/workflows/ci_cmake.yml
```diff
@@ -0,0 +1,130 @@
+# Copyright (c) 2023-present The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+name: CI_CMAKE
+on: [push, pull_request]
+
```
azuchi, line 7: Why did you create a separate file instead of ci.yml?

.github/workflows/ci_cmake.yml
```diff
@@ -0,0 +1,130 @@
...
+          ${{ github.workspace }}/test/functional/test_runner.py -j $(( $(nproc) * 2 )) --combinedlogslen=99999999 --coverage --failfast --quiet 
+          echo "---------------- DONE ---------------------"
+
+      - name: Debug Core Dump (if exists)
+        if: failure()
+        run: |
+          if [ -f core ]; then
+            gdb -batch -ex "bt" -ex "quit" ./test_binary core
+          fi
+
```
azuchi, line 58: For Linux, just like with current CI, testing is required on both ARM and x86 architectures.

.github/workflows/ci_cmake.yml
```diff
@@ -0,0 +1,130 @@
...
+name: CI_CMAKE
+on: [push, pull_request]
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CI_FAILFAST_TEST_LEAVE_DANGLING: 1  # GHA does not care about dangling processes and setting this variable avoids killing the CI script itself on error
+  MAKEJOBS: '-j10'
```
azuchi, line 14: The script below uses $(nproc), this variable is necessary?

cmake/modules/FindUPNP.cmake
```diff
@@ -1,14 +1,16 @@
...
+      find_brew_prefix(BREW_HINT_UPNP miniupnpc)
 
-if(UPNP_INCLUDE_DIR AND UPNP_LIBRARY)
-	# Already in cache, be silent
-    set(UPNP_FIND_QUIETLY TRUE)
-endif()
+      if(UPNP_INCLUDE_DIR AND UPNP_LIBRARY)
+          # Already in cache, be silent
+          set(UPNP_FIND_QUIETLY TRUE)
+      endif()
```
azuchi, line 11: Is the indentation above messed up?

.github/workflows/ci_cmake.yml
```diff
@@ -0,0 +1,130 @@
...
+          if [ -f core ]; then
+            gdb -batch -ex "bt" -ex "quit" ./test_binary core
+          fi
+
+      - name: Save Ccache cache
+        uses: actions/cache/save@v4
+        if: steps.ccache-cache.outputs.cache-hit != 'true'
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ${{ github.job }}-ccache-${{ github.run_id }}
```
azuchi, line 130: What about existing WIN64 tests?

.github/workflows/ci_cmake.yml
Naviabheeman: I wanted to keep the gnu build CI and cmake based ci in different files. The idea was that this one can be a basic smoke test. If this passes it can trigger the other tests which we'll configure with a matrix combination of dependencies, compilers and os.

.github/workflows/ci_cmake.yml
Naviabheeman: I haven't attempted the cross compilation for windows using make yet. Let me fix linux and macOS with the new sdk first. 
This would still run using the gnu autotools ci.

src/xfieldhistory.cpp
```diff
@@ -20,6 +22,7 @@ bool CXFieldHistoryMap::IsNew(TAPYRUS_XFIELDTYPES type, const XFieldChange& xFie
 }
 
 void CXFieldHistoryMap::Add(TAPYRUS_XFIELDTYPES type, const XFieldChange& xFieldChange) {
+    LOCK(xfieldHistoryMutex);
```
azuchi, line 25: Why is this modification, including xfieldhistory.h, necessary for CMake?

Naviabheeman: When I was running `cmake ci ` `ctest` command it ran running `test_tapyrus` with multiple threads. There was a risk of the global data being corrupted when `xfieldhistory` run in parallel with other tests. Suspecting this as the cause of test failures I added this code to make the access synchronous.
With `ctest` I couldn't find how to forward arguments to `test_tapyrus`. So I changed the CI to use `test_tayrus` command directly with all the arguments.

src/random.cpp
```diff
@@ -17,7 +17,7 @@
 #include <sync.h>     // for WAIT_LOCK
 #include <utiltime.h> // for GetTime()
 
-#include <stdlib.h>
+#include <unistd.h>
```
azuchi, line 20: Why is this change necessary?
There is a conditional include in `L37-L39`.

Naviabheeman: Right, the new include is not necessary.

CMakeLists.txt
```diff
@@ -1,20 +1,37 @@
...
+  # `codegen` is a reserved target name.
+  # See: https://cmake.org/cmake/help/latest/policy/CMP0171.html
+  cmake_policy(SET CMP0171 NEW)
 endif()
 
 #=============================
 # Project / Package metadata
 #=============================
 set(CLIENT_NAME "Tapyrus Core")
+set(PACKAGE_NAME "Tapyrus Core")
```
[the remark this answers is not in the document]
Naviabheeman: Yes, bitcoin renamed all PACKAGE_NAME to CLIENT_NAME in the makefiles. Since I didn't want to change the old old makefiles unnecessarily, I just added the missing definition to fix the compilation. in the recent commit I used 'PACKAGE_NAME' everywhere. In case you prefer using the name as bitcoin I'll update it to CLIENT_NAME

PR #333, opened 2025-03-17
Fix macOS cross compilation with Qt5 upgrade to 5.15.16 

Naviabheeman:
This PR fixes the Mac OS cross compilation on linux. 
Additional CI job is added  to verify tapyrus-gui build with qt from depends and with qt from system installation.

Depends:
Qt and zeromq are upgraded.
Boost is changed to header only dependency

Mac OS deployment is changed from `dmg` to `zip` porting `macdeployqtplus` script from bitcoin.


.github/workflows/ci.yml
```diff
@@ -91,17 +104,17 @@ jobs:
...
           GOAL="install" \
-          BITCOIN_CONFIG="--enable-zmq --with-gui=qt5 --enable-reduce-exports --enable-debug" \
+          BITCOIN_CONFIG="--enable-zmq --with-gui=qt5 --enable-reduce-exports" \
           CXXFLAGS="-g0 -O2" \
           bash -xe ./build-test.sh
 
-  linux-with-qt-and-system-lib:
-    name: "x86_64 Linux (Qt5 & system libs)"
+  linux_x86_with_only_system_lib:
+    name: x86_64 Linux, with GUI, system libs"
```
azuchi, line 117:
```suggestion
    name: x86_64 Linux, with GUI, system libs
```

.github/workflows/ci.yml
```diff
@@ -73,16 +68,34 @@ jobs:
...
+          DEP_OPTS="V=1" \
+          RUN_TESTS=true \
+          RUN_BENCH=true \
+          GOAL="install" \
+          BITCOIN_CONFIG="--enable-zmq --with-gui=qt5 --enable-reduce-exports" \
+          CXXFLAGS="-g0 -O2" \
+          bash -xe ./build-test.sh
+
+  linux-x86_linux_qt5_release:
+    name: x86_64 Linux, qt5 dev, depends, Release"
```
azuchi, line 98:
```suggestion
    name: x86_64 Linux, qt5 dev, depends, Release
```

src/net_processing.cpp
```diff
@@ -2362,6 +2362,7 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
         mapAlreadyAskedFor.erase(inv.hash);
 
         CTxMempoolAcceptanceOptions opt;
+        CValidationState& state = opt.state;
```
azuchi, line 2365: Why this change necessary?

src/policy/packages.cpp
```diff
@@ -102,7 +102,7 @@ bool SubmitPackageToMempool(const Package& package,
                                   PackageValidationState& results,
                                   CTxMempoolAcceptanceOptions& opt)
 {
-    if(!CheckPackage(package, state))
+   if(!CheckPackage(package, state))
```
azuchi, line 105: Please revert this change.

src/consensus/validation.h
```diff
@@ -93,6 +93,20 @@ class CValidationState {
...
+        if(chRejectCode)
+            ss << "reject-reason:" << chRejectCode << " :  "  << strRejectReason << "; " ;
+        if(!strDebugMessage.empty())
+            ss << strDebugMessage << " ";
+        if(corruptionPossible)
+            ss << "corruption possible";
+        if(missingInputs)
+            ss << "missing inputs";
+        return ss.str();
+    }
```
azuchi, line 109: I think, this method does not used anywhere.

contrib/gitian-descriptors/gitian-osx.yml
```diff
@@ -35,7 +35,7 @@ files:
 - "MacOSX10.15.sdk.tar.gz"
```
azuchi, line 35: You should be change `python` and `pythond-dev` to `python3` and `python3-dev` to fix compile error in gitian build.

```
Package python-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python2-dev python2 python-dev-is-python3

Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  2to3 python2-minimal python2 dh-python python-is-python3

E: Package 'python' has no installation candidate
E: Package 'python-dev' has no installation candidate
```

azuchi (changes requested):

contrib/gitian-descriptors/gitian-osx.yml
```diff
@@ -6,40 +6,37 @@ suites:
...
 - "amd64"
 packages:
+- "bsdmainutils"
 - "ca-certificates"
 - "curl"
-- "g++"
+- "clang"
+- "lld"
+- "llvm"
+- "llvm-dev"
```
azuchi, line 15: In `depends/README.md`, requires `g++`:

https://github.com/chaintope/tapyrus-core/blob/master/depends/README.md#for-macos-cross-compilation

Which one correct?

azuchi (changes requested):

contrib/gitian-descriptors/gitian-osx.yml
Naviabheeman: It is the documentation that is not updated. I'll update it according the list in the CI. CI library list is the most recent and accurate.

azuchi (approved):

PR #336, opened 2025-05-26
[WIP] add cmake based build support to dependencies

Naviabheeman:


depends/funcs.mk
```diff
@@ -88,6 +88,7 @@ $(1)_extract_cmds ?= mkdir -p $$($(1)_extract_dir) && echo "$$($(1)_sha256_hash)
 $(1)_preprocess_cmds ?= true
 $(1)_build_cmds ?= true
 $(1)_config_cmds ?= true
+$(1)_cmake_config_cmds ?= true
```
azuchi, line 91: Need this configuration  to make a cmake build?

Naviabheeman: I added this in order to define the cmake configurations in all the package files without removing auto tools. This worked on macOS and ubuntu. But windows build setup was messed, some extra LDFLAGS started appearing. So I reverted back the package files. But this still remains.

PR #337, opened 2025-05-26
Cmake build Part 3 -  Change tests

Naviabheeman:
This PR changes the header generated for json test data according to bitcoin. The same format is used in cmake in bitcoin. This is done prior to cmake file changes to decrease the number of files in the final PR.


src/test/base32_tests.cpp
```diff
@@ -5,8 +5,8 @@
 
 #include <utilstrencodings.h>
 #include <test/test_tapyrus.h>
-
 #include <boost/test/unit_test.hpp>
+#include <bech32.h>
```
azuchi, line 9: Why is bech32 relevant to testing base32?

src/test/timeoffsets_tests.cpp
```diff
@@ -13,6 +13,8 @@
 
 using namespace std::chrono_literals;
 
+
+
```
azuchi, line 17: This doesn't need to be changed, right?

Naviabheeman: Yes, some of the changes in header includes were not necessary. I cleaned up those and kept only what was needed.

azuchi (approved):

PR #338, opened 2025-06-10
Cmake unit tests are failing because of very subtle errors

Naviabheeman:
All unit tests pass in the automate CI. but in cmake there are some failures:
1. The problem is caused by the global xfieldHistory map not being reset between test runs in the CMake CI. So data added by one test  persists and affects subsequent tests.
2. Low fee making IsStandard test fail counting the output as dust
3. Block proof size was 0 instead of 64 in the constructor of TestingChainSetup making test_tapyrus exit

### DIFF
--- a/src/primitives/block.cpp
+++ b/src/primitives/block.cpp
@@ -73,8 +73,7 @@ bool CBlockHeader::AbsorbBlockProof(const std::vector<unsigned char>& blockproof
 
     //add signatures to block
     proof = std::move(blockproof);
-    if(proof.size() != CPubKey::SCHNORR_SIGNATURE_SIZE)
-        return false;
+
     return true;
 
 }
--- a/src/primitives/block.cpp
+++ b/src/primitives/block.cpp
@@ -73,6 +73,8 @@ bool CBlockHeader::AbsorbBlockProof(const std::vector<unsigned char>& blockproof
 
     //add signatures to block
     proof = std::move(blockproof);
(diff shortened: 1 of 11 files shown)


src/util.cpp
```diff
@@ -113,18 +113,15 @@ LockResult LockDirectory(const fs::path& directory, const std::string lockfile_n
...
+    auto lock = std::make_unique<fsbridge::FileLock>(pathLockFile.string().c_str());
+    if (!lock->TryLock()) {
         return LockResult::ErrorLock;
     }
+    if (!probe_only) {
+        // Lock successful and we're not just probing, put it into the map
+        dir_locks.emplace(pathLockFile.string(), std::move(lock));
+    }
+
     return LockResult::Success;;
```
azuchi, line 125:
```suggestion
    return LockResult::Success;
```

src/primitives/block.cpp
```diff
@@ -73,6 +73,8 @@ bool CBlockHeader::AbsorbBlockProof(const std::vector<unsigned char>& blockproof
 
     //add signatures to block
     proof = std::move(blockproof);
+    if(proof.size() != CPubKey::SCHNORR_SIGNATURE_SIZE)
+        return false;
```
azuchi, line 77: Why is this check necessary when L64-L65 does the following check:

```cpp
if(blockproof.size() != CPubKey::SCHNORR_SIGNATURE_SIZE)
        return false;
```

PR #341, opened 2025-07-16
[WIP] Fix lock directory

Naviabheeman:


src/test/util_tests.cpp
```diff
@@ -1131,14 +1131,20 @@ enum : char {
...
 };
 
-[[noreturn]] static void TestOtherProcess(fs::path dirname, fs::path lockname, int fd)
+/*
+ * Use _exit in the child process instead of exit to avoid destructors.
+ * All memory allocated is done by parent process and released by parent process.
+ * When destructors and invoked by child process, it causes double memory release,
+ * invalid pointers and non 0 exit code. Unit test failed because of this.
+ */
+static void TestOtherProcess(fs::path dirname, fs::path lockname, int fd)
```
azuchi, line 1140: The `[[noreturn]]` attribute has been removed, but the TestOtherProcess function never actually returns , so shouldn't the attribute be left as it is?

PR #342, opened 2025-07-23
Extended Cmake ci

Naviabheeman:
This PR Splits the Cmake CI into two:
1. Smoke test which runs the cmake build using installed dependencies on linux and macOS, x86_64 and arm64. It runs unit testa, a short benchmark test and fast functional tests. Clang compiler is used in both linux and macOS(similar to bitcoin)
2. Daily test which builds depends directory using cmake. Then builds tapyrus using these libraries. Linux builds use gcc and macOS use clang. Multiple combinations like GUI, USDT, debug mode, no wallet etc. are built and tested.


azuchi (changes requested):
* Please include an explanation of smoke-test and daily-test in the README.
* The Qt-related fixes overlap with PR#344, so please exclude them from this PR.

azuchi (approved):

PR #343, opened 2025-07-28
Fix check queue deadlock

Naviabheeman:
The unit test check queue often hangs when using cmake. This seems to be a race condition as it doesn't happen in every run and restarting fixes it. Creating this PR to check if the issue can be fixed consistently.

sample log:
https://github.com/chaintope/tapyrus-core/actions/runs/16612039425/job/46996844684

### DIFF
--- a/src/test/checkqueue_tests.cpp
+++ b/src/test/checkqueue_tests.cpp
@@ -96,6 +96,8 @@ struct FrozenCleanupCheck {
     static std::atomic<uint64_t> nFrozen;
     static std::condition_variable cv;
     static std::mutex m;
+    static std::atomic<bool> should_stop;
+    static std::atomic<bool> exited_via_stop_flag;
     // Freezing can't be the default initialized behavior given how the queue
     // swaps in default initialized Checks.
     bool should_freeze {true};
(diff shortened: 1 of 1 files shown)


src/checkqueue.h
```diff
@@ -171,10 +172,13 @@ class CCheckQueue
...
 
-        if (vChecks.size() == 1)
+        if (vChecks.size() == 1) {
             condWorker.notify_one();
-        else if (vChecks.size() > 1)
+        } else if (vChecks.size() > 1) {
             condWorker.notify_all();
+        }
+        // Also notify master in case it's waiting
+        condMaster.notify_one();
```
azuchi, line 181: From AI review:

 The addition of condMaster.notify_one() in the Add() method is unnecessary and may negatively impact performance.

  Reasons:
  1. The Add() method is called to add new tasks to the queue for worker threads to process
  2. The master thread only waits on condMaster at line 90 when nTodo == 0 (no pending work)
  3. Since Add() increments nTodo (line 171), the master thread cannot be in a waiting state when Add() is called
  4. The proper notification to condMaster already exists at line 83, which correctly notifies when the last task is completed (nTodo == 0)

Naviabheeman: Right! This turned out to be a bad example of trusting the AI completely. In fact I had noted in the test case that it hanged sometimes at a particular `wait()` and this issue is limited to the test case only. So I had tried to debug the issue earlier but couldn't find the right fix. it's done now. There is an additional condition that triggers the exit.

azuchi (approved):

PR #349, opened 2025-09-07
Fix check queue deadlock

Naviabheeman:
Multiple deadlocks have been detected in the CI recently,
https://github.com/chaintope/tapyrus-core/actions/runs/17559929460/job/49873526564
https://github.com/chaintope/tapyrus-core/actions/runs/17559929460/job/49873526541
https://github.com/chaintope/tapyrus-core/actions/runs/17500977895/job/49713610774

The primary issue was that the master thread was waiting on a condition variable with predicate !queue.empty() || 
m_request_stop, but it needed to also wake up when nTodo == 0 (work completion). Without this, the master could
finish processing work but remain stuck waiting, while workers were also waiting for more work that would never
come.

### DIFF
--- a/src/checkqueue.h
+++ b/src/checkqueue.h
@@ -201,7 +201,6 @@ public:
             m_request_stop = true;
         }
         condWorker.notify_all();
-        //condMaster.notify_all();  // Also notify master thread
         for (std::thread& t : m_worker_threads) {
             t.join();
         }
--- a/src/checkqueue.h
+++ b/src/checkqueue.h
@@ -97,7 +97,12 @@ private:
                         return fRet;
                     }
                     nIdle++;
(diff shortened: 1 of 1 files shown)


src/checkqueue.h
```diff
@@ -180,8 +196,12 @@ class CCheckQueue
     //! Stop all of the worker threads.
     ~CCheckQueue()
     {
-        m_request_stop = true;
+        {
+            WaitableLock lock(mutex);
+            m_request_stop = true;
+        }
         condWorker.notify_all();
+        //condMaster.notify_all();  // Also notify master thread
```
azuchi, line 204: Isn't this commented out line unnecessary?

azuchi (approved):

PR #354, opened 2025-10-08
Change config name of aarch64-apple-dawrin to arm64-apple-darwin

Naviabheeman:
Rename aarch64 to arm64 in macOS cross compilation 
Use cmake to build zeromq and libevent in depends
Use python 3 interpreter instead of python3_executable in cmake configuration files


depends/config.guess
```diff
@@ -1176,249 +1344,414 @@ EOF
...
-	exit ;;
+	GUESS=powerpc-apple-rhapsody$UNAME_RELEASE
+	;;
     *:Rhapsody:*:*)
-	echo "$UNAME_MACHINE"-apple-rhapsody"$UNAME_RELEASE"
-	exit ;;
+	GUESS=$UNAME_MACHINE-apple-rhapsody$UNAME_RELEASE
+	;;
+    arm64:Darwin:*:*)
+	GUESS=aarch64-apple-darwin$UNAME_RELEASE
```
azuchi: `aarch64-apple` remains.

Naviabheeman: [config.guess](https://gitweb.git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD) and [`config.sub`](https://gitweb.git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD) scripts are the latest scripts from GNU. Do we need to edit it?
I also verified that both `arm64-apple-darwin` and `aarch64-apple-darwin` configurations work. i.e. I'm able to build depends and tapyrusd using both these configurations on macOS native build.

azuchi: I see. It's ok as it is.

azuchi (approved):

PR #356, opened 2025-10-28
port bitcoin PR24858

Naviabheeman:
incorrect blk file size calculation during reindex results in recoverable blk file corruption. This was the root cause of the "out of order" blocks reported in all failures of feature_xfield_maxblocksize test

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -51,6 +51,7 @@ BITCOIN_TESTS =\
   test/cuckoocache_tests.cpp \
   test/denialofservice_tests.cpp \
   test/descriptor_tests.cpp \
+  test/file_io_tests.cpp \
   test/getarg_tests.cpp \
   test/hash_tests.cpp \
   test/key_io_tests.cpp \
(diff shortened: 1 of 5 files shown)


azuchi (changes requested):
Add regression tests like Bitcoin's PR24858 `blockmanager_tests.cpp`.

azuchi (commented):
Also, does this mean that the tests in `feature_xfield_maxblocksize.py` need to be changed?

Naviabheeman:
> Also, does this mean that the tests in `feature_xfield_maxblocksize.py` need to be changed?

The test should work without an additional node after reindex. That is not related to this fix, there is a bug in the size lookup during reindex causing block validation to fail for the large block. This bug has been there since I implemented it. I'm still validating the fix.

src/Makefile.test.include
```diff
@@ -51,6 +51,7 @@ BITCOIN_TESTS =\
   test/cuckoocache_tests.cpp \
   test/denialofservice_tests.cpp \
   test/descriptor_tests.cpp \
+  test/file_io_tests.cpp \
```
azuchi: Do you need to add this to CMakeLists.txt as well?

azuchi (approved):

PR #366, opened 2025-11-20
Fix guix builds

Naviabheeman:
Qt 5 can only be built by c++17. But master branch is updated to build with c++20. So some dependencies of Qt were built with c++20 standard while Qt was built with c++17 standard. This caused segmentation fault in the compiler. This is fixed by hardcoding the c++17 standard in qt while using one common standard definition for the other dependencies.

### DIFF
--- a/contrib/guix/manifest.scm
+++ b/contrib/guix/manifest.scm
@@ -494,9 +494,9 @@ inspecting signatures in Mach-O binaries.")
         pkg-config
         bison
         cmake-minimal
-        ;; Native GCC 10 toolchain
-        gcc-toolchain-10
-        (list gcc-toolchain-10 "static")
+        ;; Native GCC 11 toolchain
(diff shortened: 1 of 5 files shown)


azuchi (changes requested):
The error still persists. 

```
...
/distsrc-base/distsrc-05c5b959216f-x86_64-linux-gnu/src/qt/guiutil.cpp:831:48: error: cannot convert 'std::u8string' {aka 'std::__cxx11::basic_string<char8_t>'} to 'const string&' {aka 'const std::__cxx11::basic_string<char>&'}
  831 |     return QString::fromStdString(path.u8string());
      |                                   ~~~~~~~~~~~~~^~
      |                                                |
      |                                                std::u8string {aka std::__cxx11::basic_string<char8_t>}
...
```

Does it work in your environment?

azuchi (approved):

PR #368, opened 2025-12-05
Upgrade tapyrus-qt from Qt5 to Qt6 phase 1

Naviabheeman:
This update contains the minimal compilable changes to upgrade Qt5 to Qt6. This is entirely generated by Claude code agent.


depends/packages/qt.mk
```diff
@@ -214,75 +160,33 @@ define $(package)_extract_cmds
...
+  export PKG_CONFIG_LIBDIR=$(host_prefix)/lib/pkgconfig && \
+  export PKG_CONFIG_PATH=$(host_prefix)/share/pkgconfig && \
   cd qtbase && \
-  ./configure  -top-level $($(package)_config_opts)
+  ./configure $($(package)_config_opts) -- $($(package)_cmake_opts)
 endef
 
 define $(package)_build_cmds
-  $(MAKE)
+  cd qtbase && cmake --build .
```
azuchi: Before the change, qt.pro would also build qttools and qttranslations, but this change means that they are no longer built.

azuchi (changes requested):

depends/packages/qt.mk
Naviabheeman: oh yes! sorry I didn't catch this during the build. I'll check how to build qttools and qttranslations too,

azuchi (approved):

PR #373, opened 2025-12-17
Lint test in ci

Naviabheeman:
1. Enable lint tests in the CI
2. Rename pending-* scripts to enable executing all the scripts
3. Fix listing errors int he codebase identified by the current scripts
4. Add new python scripts like bitcoin for python syntax checks


contrib/seeds/makeseeds.py
```diff
@@ -28,7 +28,7 @@
 }
 
 PATTERN_IPV4 = re.compile(r"^((\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})):(\d+)$")
-PATTERN_IPV6 = re.compile(r"^\[([0-9a-z:]+)\]:(\d+)$")
+PATTERN_IPV6 = re.compile(r"^[([0-9a-z:]+)\]:(\d+)$")
```
azuchi: ^\[ becomes ^[(, which breaks parsing of IPv6 addresses. It should be reverted.

azuchi (changes requested):

entrypoint.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -eu
 
+export LC_ALL=C
```
azuchi: This change related with this PR?

azuchi (approved):

PR #378, opened 2026-01-13
Qt6 upgrade phase2

Naviabheeman:


azuchi (changes requested):
The procedure in the doc is still for Qt5, so please correct this as well.

PR #381, opened 2026-01-21
Qt6 upgrade phase2

Naviabheeman:
Update Qt5 style code to Qt6:
1. Update all SIGNAL and SLOT to use Qt6 function-pointer
2. Remove Qt namespace macros : QT_BEGIN_NAMESPACE and QT_END_NAMESPACE


azuchi (changes requested):
As with PR#378, procedure in the doc is still for Qt5, so please correct this as well.

azuchi (approved):

PR #394, opened 2026-03-13
Fix failure in out of order block processing during reindex

Naviabheeman:
- remove height as a parameter of `ReadBlockFromDisk`
- pass xfield history pointer to `CheckBlockHeader` from `ReadBlockFromDisk` to get the correct aggregatepubkey at the given height during a reindex.
-  add new test case for out of order block with reindex.
 - remove changes from federation management test made in the previous commit(which caused CI failure because of out of order block reindex failure)


azuchi (changes requested):
You probably need to add the newly added `feature_reindex_outoforder_block.py` to `test_runner.py`, right?

azuchi (approved):

PR #397, opened 2026-03-20
Add Label text to issue token/burn token tab

Naviabheeman:


azuchi (changes requested):
@Naviabheeman wallet.cpp does not appear to be related to the content of this pull request.

azuchi (approved):

PR #398, opened 2026-03-24
fix macos native build 

Naviabheeman:
Mac OS native build fails because std::view::reverse is missing. It is replaced with `begin()` and `rend()` vector iterators.


azuchi (changes requested):
No changes are needed to qt/* or interface/*.

azuchi (approved):

PR #401, opened 2026-04-08
Port Bitcoin Core PR#25325: Pool-based memory resource for CCoinsMap

Naviabheeman:
Adds a PoolAllocator backed by PoolResource for the UTXO cache map (CCoinsMap). Pre-allocates memory in 256KB chunks and serves node allocations from freelists, eliminating per-entry malloc/free overhead during IBD.

Changes:
- src/support/allocators/pool.h: new PoolResource + PoolAllocator
- src/coins.h: replace CCoinsMap typedef with PoolAllocator variant; add CCoinsMapMemoryResource member; declare ReallocateCache()
- src/coins.cpp: wire resource into CCoinsMap constructor; add ReallocateCache() to return chunk memory after each Flush()
- src/memusage.h: add DynamicUsage overload for PoolAllocator maps

Tapyrus adaptation notes:
- No m_deterministic flag; SaltedOutpointHasher{} used directly
- COutPoint uses hashMalFix (same sizeof as Bitcoin's hash field)
- No validation.cpp change needed (Tapyrus has no ResizeCoinsCaches)

Bitcoin Core reference: https://github.com/bitcoin/bitcoin/commit/5aa0c82ccd6ceb4a141686fc8658f679de75a787

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -221,10 +221,12 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlockIn
 
 bool CCoinsViewCache::Flush() {
     bool fOk = base->BatchWrite(cacheCoins, hashBlock);
-    if (fOk) {
-        cacheCoins.clear();
-        ReallocateCache();
-    }
+    // BatchWrite (both CCoinsViewDB and CCoinsViewCache) erases entries from
+    // cacheCoins unconditionally during iteration, so the map is always empty
+    // clear() here is defensive; ReallocateCache() returns the pool backing memory.
+    // All three are safe to call on failure and success.
+    cacheCoins.clear();
+    ReallocateCache();
     cachedCoinsUsage = 0;
     return fOk;
 }
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -33,7 +33,9 @@ size_t CCoinsViewBacked::EstimateSize() const { return base->EstimateSize(); }
 
 SaltedOutpointHasher::SaltedOutpointHasher() : k0(GetRand(std::numeric_limits<uint64_t>::max())), k1(GetRand(std::numeric_limits<uint64_t>::max())) {}
 
-CCoinsViewCache::CCoinsViewCache(CCoinsView *baseIn) : CCoinsViewBacked(baseIn), cachedCoinsUsage(0) {}
+CCoinsViewCache::CCoinsViewCache(CCoinsView *baseIn) : CCoinsViewBacked(baseIn),
+    cacheCoins(0, SaltedOutpointHasher{}, CCoinsMap::key_equal{}, &m_cache_coins_memory_resource),
+    cachedCoinsUsage(0) {}
 
 size_t CCoinsViewCache::DynamicMemoryUsage() const {
     return memusage::DynamicUsage(cacheCoins) + cachedCoinsUsage;
@@ -219,9 +221,22 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlockIn
 
 bool CCoinsViewCache::Flush() {
     bool fOk = base->BatchWrite(cacheCoins, hashBlock);
-    cacheCoins.clear();
+    if (fOk) {
+        cacheCoins.clear();
+        ReallocateCache();
+    }
     cachedCoinsUsage = 0;
     return fOk;
+}
+
+void CCoinsViewCache::ReallocateCache()
+{
+    // Cache should be empty when we're calling this.
+    assert(cacheCoins.size() == 0);
+    cacheCoins.~CCoinsMap();
+    m_cache_coins_memory_resource.~CCoinsMapMemoryResource();
+    ::new (&m_cache_coins_memory_resource) CCoinsMapMemoryResource{};
+    ::new (&cacheCoins) CCoinsMap{0, SaltedOutpointHasher{}, CCoinsMap::key_equal{}, &m_cache_coins_memory_resource};
 }
 
 void CCoinsViewCache::Uncache(const COutPoint& hash)
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -221,19 +221,31 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlockIn
 
 bool CCoinsViewCache::Flush() {
     bool fOk = base->BatchWrite(cacheCoins, hashBlock);
-    // BatchWrite (both CCoinsViewDB and CCoinsViewCache) erases entries from
-    // cacheCoins unconditionally during iteration, so the map is always empty
-    // clear() here is defensive; ReallocateCache() returns the pool backing memory.
-    // All three are safe to call on failure and success.
-    cacheCoins.clear();
-    ReallocateCache();
+    if (fOk) {
+        // BatchWrite (both CCoinsViewDB and CCoinsViewCache) erases entries from
+        // cacheCoins inside the iteration loop, so the map must be empty here.
+        // The assertion catches any future refactor that breaks that invariant.
+        if (!cacheCoins.empty()) {
+            throw std::logic_error("Not all cached coins were erased");
+        }
+        ReallocateCache();
+    }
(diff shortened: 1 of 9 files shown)


src/coins.cpp
```diff
@@ -219,11 +221,24 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlockIn
 
 bool CCoinsViewCache::Flush() {
     bool fOk = base->BatchWrite(cacheCoins, hashBlock);
-    cacheCoins.clear();
+    if (fOk) {
+        cacheCoins.clear();
+        ReallocateCache();
+    }
     cachedCoinsUsage = 0;
```
azuchi, line 233: I have a question about the Flush() implementation.                                                                                                                                                                          
                                                                                                                                                                                                                               
  In both the current Tapyrus and Bitcoin Core, `cacheCoins.clear()` is always executed.                                                                                                                                       
  However, in this PR, it is only executed when `fOk` is true:                                                                                                                                                                 
                                                                                                                                                                                                                               
  ```cpp                                                                                                                                                                                                                       
  if (fOk) {                                                                                                                                                                                                                   
      cacheCoins.clear();                                                                                                                                                                                                      
      ReallocateCache();                                                                                                                                                                                                       
  }                                                                                                                                                                                                                            
  cachedCoinsUsage = 0;  // <- This is always executed                                                                                                                                                                         
  ```                                                                                                                                                                                                                          
                                                                                                                                                                                                                               
  Is this change intentional? If so, could you explain the reasoning?                                                                                                                                                          
                                                                                                                                                                                                                               
  Also, when `fOk` is false, the cache is preserved but `cachedCoinsUsage` is reset to 0,                                                                                                                                      
  which seems inconsistent. Could you clarify this behavior as well?

Naviabheeman: `if (fOk)` was added  under the assumption that on a failed flush the cache still had entries and needed to be preserved.  That assumption was wrong because `BatchWrite` erases entries unconditionally regardless of outcome.

azuchi: Thanks for the analysis — that matches what I see when I trace through `CCoinsViewDB::BatchWrite` (txdb.cpp:108) and `CCoinsViewCache::BatchWrite` (coins.cpp:165): both erase entries from `mapCoins` inside the iteration loop, so `cacheCoins` is empty by the time `BatchWrite` returns regardless of `fOk`.

Given that, the `cacheCoins.clear()` here is a no-op. Bitcoin Core's PR #25325 handles the same situation by replacing the redundant clear with an invariant assertion:

```cpp
if (fOk) {
    if (!cacheCoins.empty()) {
        throw std::logic_error("Not all cached coins were erased");
    }
    ReallocateCache();
}
```

This surfaces a programming error if `BatchWrite` ever stops erasing entries (e.g., a future refactor that breaks the in-loop erase). I'd suggest mirroring that here — it costs nothing at runtime and turns a silent regression into a loud one.

If you'd rather keep the current `clear()`, please add a short comment explaining that BatchWrite has already emptied the map and the call is defensive only.

azuchi (commented):
Additional review notes — the port itself looks faithful and the open thread on `Flush()` has a separate reply. Comments below cover testing and documentation gaps that would be worth addressing before merge.

src/support/allocators/pool.h
```diff
@@ -0,0 +1,350 @@
...
+ */
+template <std::size_t MAX_BLOCK_SIZE_BYTES, std::size_t ALIGN_BYTES>
+class PoolResource final
+{
+    static_assert(ALIGN_BYTES > 0, "ALIGN_BYTES must be nonzero");
+    static_assert((ALIGN_BYTES & (ALIGN_BYTES - 1)) == 0, "ALIGN_BYTES must be a power of two");
+
+    /**
+     * In-place linked list of the allocations, used for the freelist.
+     */
```
azuchi, line 78: **M1 — `pool_tests.cpp` is not ported (medium)**

Bitcoin Core PR #25325 includes ~500 lines of unit tests for `PoolResource` covering allocation/deallocation, freelist behavior, chunk reuse across destruct/reconstruct cycles, alignment edge cases, and oversized allocations falling back to `operator new`. None of those tests are in this PR.

This is a 350-line bespoke allocator landing in production with no direct tests. The existing `coins_tests.cpp` exercises it indirectly through `CCoinsMap`, but that doesn't cover the freelist edge cases (e.g., chunk exhaustion mid-allocation, alignment > `ELEM_ALIGN_BYTES` falling through to `::operator new`, `Deallocate` of pool-allocated vs new-allocated blocks).

**Suggested fix:** port [`src/test/pool_tests.cpp`](https://github.com/bitcoin/bitcoin/blob/master/src/test/pool_tests.cpp) from Bitcoin Core. The `PoolResourceTester` friend class declared at line 116 is already in place expecting it.

src/coins.cpp
```diff
@@ -219,11 +221,24 @@ bool CCoinsViewCache::BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlockIn
...
 
+void CCoinsViewCache::ReallocateCache()
+{
+    // Cache should be empty when we're calling this.
+    assert(cacheCoins.size() == 0);
+    cacheCoins.~CCoinsMap();
+    m_cache_coins_memory_resource.~CCoinsMapMemoryResource();
+    ::new (&m_cache_coins_memory_resource) CCoinsMapMemoryResource{};
+    ::new (&cacheCoins) CCoinsMap{0, SaltedOutpointHasher{}, CCoinsMap::key_equal{}, &m_cache_coins_memory_resource};
+}
```
azuchi, line 254: **M2 — Placement-new pattern needs an explanatory comment (minor)**

Destroy-and-reconstruct-in-place on member objects is unusual and easy to misread as undefined behavior. A reader unfamiliar with Bitcoin Core PR #25325 has to reverse-engineer the intent.

Suggested doc comment above the function:

```cpp
// After Flush(), the pool resource still holds all chunks that were used
// during this flush cycle. We want to return that memory to the OS so the
// next flush cycle starts clean. We can't simply reset the resource because
// cacheCoins's allocator holds a pointer to it — so we tear down both objects
// and reconstruct them in place. The member addresses are preserved, so any
// references held elsewhere (there should be none, but defensively) remain
// valid.
//
// Order matters: cacheCoins must be destroyed before the resource it
// allocates from, and reconstructed after the new resource is in place.
```

No code change required.

src/coins.h
```diff
@@ -208,6 +226,7 @@ class CCoinsViewCache : public CCoinsViewBacked
      * declared as "const".
      */
     mutable uint256 hashBlock;
+    mutable CCoinsMapMemoryResource m_cache_coins_memory_resource{};
```
azuchi, line 233: **M3 — Make the declaration-order dependency explicit (minor)**

`m_cache_coins_memory_resource` must be declared before `cacheCoins` because the latter's constructor takes `&m_cache_coins_memory_resource`. C++ initializes members in declaration order, so swapping these two lines would pass a pointer to an uninitialized object → UB.

A short comment makes this hard to break:

```cpp
// Note: declaration order is load-bearing. m_cache_coins_memory_resource
// must precede cacheCoins because cacheCoins's allocator stores a pointer
// to it, which is dereferenced from cacheCoins's destructor (and from
// every allocate/deallocate). Reordering these is undefined behavior.
mutable CCoinsMapMemoryResource m_cache_coins_memory_resource{};
mutable CCoinsMap cacheCoins;
```

src/coins.h
```diff
@@ -208,6 +226,7 @@ class CCoinsViewCache : public CCoinsViewBacked
...
     mutable CCoinsMap cacheCoins;
```
azuchi, line 237: **M4 — Document the thread-safety contract (minor)**

`pool.h` line 35 explicitly states *"PoolResource is not thread-safe."* In Tapyrus, `cacheCoins` (and therefore the resource) is protected by `cs_main` for all consensus paths, but that contract isn't stated at the declaration site.

Suggested:

```cpp
// cacheCoins (and the underlying memory resource) must only be accessed
// while cs_main is held. PoolResource is not thread-safe — concurrent
// access from multiple threads is undefined behavior.
mutable CCoinsMap cacheCoins;
```

This is purely documentation, no behavior change.

azuchi (approved):

PR #402, opened 2026-04-09
Fix block file truncation during reindex in macOS

Naviabheeman:
Problem
                                                                                                                                           
  During ```-reindex```, ```LoadGenesisBlock()``` was called before the block file scan in ```ThreadImport```. Inside ```LoadGenesisBlock()```,   ```SaveBlockToDisk(genesis, 0, nullptr)``` is called with ```dbp=nullptr```, which causes ```FindBlockPos``` to call ```AllocateFileRange(file, 0, 16MB)``` to   pre-allocate space for blk00000.dat.                                                              
  **_On macOS (and Windows)_**, AllocateFileRange uses ```ftruncate(fd, 16MB)```.  ftruncate shrinks files — so if  blk00000.dat already contains more than 16MB of blocks from a previous run, it gets truncated to exactly 16MB, permanently destroying all block data beyond that offset.                                                                                                                     
  **_On Linux_**, AllocateFileRange uses ```posix_fallocate```, which only allocates space and never shrinks an existing file, so the bug does not manifest.                                                                                                             
  
Root Cause                                                                                                                        
  The call to LoadGenesisBlock() before the scan was intentional **(commit d75310473)** — it initializes CXFieldHistory (Tapyrus xfieldhistory), which must be populated before any block validation runs during the scan. 
                                                                                                                                           
  Fix (src/chainstate.cpp)
  Guard the ```SaveBlockToDisk``` write with a condition: if we are reindexing and blk00000.dat already exists, skip the disk write entirely. The genesis block will be read from the existing block file by LoadExternalBlockFile during the scan.
The CXFieldHistory initialization is kept as a fallthrough outside the guard, so xfield history is always initialized regardless of the  reindex state.

Test
```feature_reindex.py``` test is enhanced with xfield history verification across two aggpubkey changes, block file size assertions before/after  each reindex, and a --longchain option that explicitly mines large blocks until blk00000.dat exceeds 16MB then verifies sizes are preserved after reindex.

### DIFF
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -4,6 +4,7 @@
 
 #include <chainstate.h>
 #include <chainparams.h>
+#include <util.h>
 
 #include <consensus/tx_verify.h>
 #include <index/txindex.h>
@@ -22,6 +23,7 @@
 // Defined in validation.cpp; declared here to avoid pulling in validation.h
 // (which includes chainstate.h, creating an indirect self-inclusion).
 void RefreshChainTxDataFromTip(const CBlockIndex* pindexNew);
+extern std::atomic_bool fReindex;
 
 static int64_t nTimeCheck = 0;
 static int64_t nTimeForks = 0;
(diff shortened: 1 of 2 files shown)


azuchi:
Some smoke test failed.

The truncation bug in `-reindex` can also occur with `-reloadxfield`, right? That also needs fixing.

Naviabheeman:
> The truncation bug in `-reindex` can also occur with `-reloadxfield`, right? That also needs fixing.

It turns out that this is already handled by this code. ```-reindex``` and ```-reloadxfield``` are different flags for a short while during initialization. When the genesis block is loaded ```LoadGenesisBlock``` and then ```ThreadImport``` loads the index from db in ```LoadBlockIndexDB``` both flags are equivalent. They set and read the same level db entry called ```fReindex```. I tested it and updated the test to check with '-reloadxfield' parameter also.

azuchi (approved):

PR #406, opened 2026-04-21
Fix coloured coin vulnerabilities

Naviabheeman:
**Vuln 1: Consensus bypass: token balance verification missing in consensus layer**
_Exploit Scenario_
  A federation signer (malicious, key-compromised, or buggy implementation) crafts a block containing:
  - vin: a single TPC UTXO of their own (only for fee) 
  - vout: 100 colored outputs with arbitrary colorId and arbitrary amounts
  If delivered through any path that bypasses mempool acceptance (IBD / reindex / submitblock RPC / compact-block propagation), the receiving node has no way to detect the token balance violation, and "tokens minted from thin air" become persistently part of the UTXO set.

**Fix:**
  Inside the per-tx loop of `ConnectBlock`, do `CheckColorIdentifierValidity` and ` VerifyTokenBalances` checks.

**Vuln 2: NFT uniqueness violation: duplicate NFT colorId outputs allowed within a single transaction** 
 _Description_ 
CheckColorIdentifierValidity iterates outputs independently and only enforces a per-output txout.nValue == 1 check. It does not enforce the invariant "at most one output per NFT colorId". The nValue == 1 check in the issuetoken RPC is a client-side wallet check only, bypassable by constructing a raw transaction directly. The OP_COLOR handler only validates the colorId encoding and does not detect duplicates either.
  Combined with Vuln 1 — where this function isn't even called during ConnectBlock — bypassing the mempool is sufficient for the bad block to land on chain.
_Exploit Scenario_ 
  - vin[0]: any TPC UTXO (prevout = X)
  - vout[0..99]: 100 outputs with the same NFT colorId = SHA256(X || NFT), each with nValue=1
  → An NFT that is supposed to be unique now circulates as 100 copies under the same colorId. Any application relying on NFT uniqueness (tickets, certificates, digital art, ownership proofs) is completely broken.

**Fix:** 
  Track NFT and Non-Reissuable colorIds in a `std::set<ColorIdentifier>`,  `g_issued_colorids`. Reject the transaction if a duplicate is found.
  
**Bug 1:  tapyrus-genesis.cpp -address flow embeds the address string itself as hash160.**
Fix:
use `GetScriptForDestination` to get the correct address from the decoded address.

**Note:**
The specification of [OP_COLOR](https://github.com/chaintope/tapyrus-core/blob/master/doc/tapyrus/colored_coin.md#concerns) has been updated to restrict OP_COLOR to be used only with CP2PKH and CP2SH. Other than these, only colored coin burn scripts may have OP_COLOR. Hence this PR removes the support of custom OP_COLOR scripts. Check script_tests.cpp for some exploit scenarios.


doc/tapyrus/colored_coin.md
```diff
@@ -24,25 +24,31 @@ Script which does not include `OP_COLOR` represents to the amount of the default
...
+But operators like `OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` may be present in the P2SH redeem script without `OP_COLOR`.
 
 As a result, we support the following types of scriptPubkey that coloring the existing P2PKH and P2SH.
 
 * `CP2PKH`(Colored P2PKH)：  
 `<COLOR identifier> OP_COLOR OP_DUP OP_HASH160 <H(pubkey)> OP_EQUALVERIFY OP_CHECKSIG`
 * `CP2SH`(Colored P2SH)：  
 `<COLOR identifier> OP_COLOR OP_HASH160 <H(redeem script)> OP_EQUAL`
+* `Burn`:
+`<COLOR identifier> OP_COLOR OP_TRUE`
```
azuchi: This is anyone can spend token due to OP_TRUE, not burned. Please delete it.

test/functional/feature_colored_coin_vuln.py
```diff
@@ -0,0 +1,843 @@
+#!/usr/bin/env python3
```
azuchi: This filename is inappropriate for future maintenance. Since it's a specification of Colored Coin, please integrate it into feature_coloredcoin.py.

azuchi (commented):
Follow-up review (the burn-template point is in a separate thread). Inline comments below cover the remaining medium and minor issues.

src/chainstate.cpp
```diff
@@ -630,12 +661,27 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+            if (!CheckInputs(tx, state, view, fScriptChecks, SCRIPT_VERIFY_NONE, fCacheResults, fCacheResults, txdata[i], nScriptCheckThreads ? &vChecks : nullptr))
                 return error("ConnectBlock(): CheckInputs on %s failed with %s",
                     tx.GetHashMalFix().ToString(), FormatStateMessage(state));
             control.Add(std::move(vChecks));
         }
 
+        //if there are colored coins in the output verify their colorids
+        if(!CheckColorIdentifierValidity(tx, state, view)) {
+            if (state.IsValid())
+                state.DoS(0, false, REJECT_COLORID, "invalid-colorid");
```
azuchi: **M1 — DoS score inconsistency**

This wrapper (and its twin in `src/validation.cpp:716`) assigns `DoS=0` whenever `CheckColorIdentifierValidity` returns `false` without setting `state`. The function does set `DoS(100, ...)` for the structural / duplicate-issuance / NFT-output-count failures, but three failure paths return `false` silently:

- `validation.cpp:478` — `txout.nValue <= 0`
- `validation.cpp:530` — NFT with `nValue != 1`
- `validation.cpp:503` — `matchFound == false` (no input matches the output colorId)

All three are consensus-rule violations. In the `ConnectBlock` path a peer that delivers such a block should be banned (DoS=100), but currently they aren't — the block is rejected, but the peer keeps retrying. This partially undercuts the Vuln 1 fix.

**Suggested fix:** set `state.DoS(100, false, REJECT_INVALID, "<reason>")` at every failure path inside `CheckColorIdentifierValidity` and remove the caller-side fallback. If mempool acceptance ever needs a softer score for a specific case, handle it explicitly inside the function.

src/validation.cpp
```diff
@@ -491,22 +521,56 @@ bool CheckColorIdentifierValidity(const CTransaction& tx, CValidationState& stat
...
 {
+    // Build input balance map from the UTXO view.
+    // Simultaneously collect TPC input outpoints for issuance detection.
+    TxColoredCoinBalancesMap inColoredCoinBalances;
+    std::vector<COutPoint> tpcInputOutpoints;
+    for (const auto& txin : tx.vin) {
+        const Coin& coin = inputs.AccessCoin(txin.prevout);
+        ColorIdentifier cid = GetColorIdFromScript(coin.out.scriptPubKey);
+        auto it = inColoredCoinBalances.find(cid);
+        if (it == inColoredCoinBalances.end())
```
azuchi, line 571: **M2 — Legacy custom OP_COLOR UTXOs convert tokens → TPC at spend time**

For a legacy custom-OP_COLOR UTXO (created before 0.7.2), `GetColorIdFromScript` now returns `NONE`. Its `nValue` — a raw token amount — is added here to the **TPC** input balance. A spender of such a UTXO then has more "TPC satoshis" to allocate to outputs than they actually own.

`CheckColorIdentifierValidity` only catches the `IsColoredScript() && type==NONE` input pattern when the spending tx *also has colored outputs*. A transaction that consumes a legacy custom-OP_COLOR UTXO and produces only TPC outputs skips the check entirely.

testnet and prod don't carry such UTXOs (confirmed during the strict-template impact study), so production is safe. But private chains running pre-0.7.2 are at risk.

**Suggested fix:** in this loop (or in `CheckColorIdentifierValidity` regardless of whether the tx has colored outputs), reject any input where `IsColoredScript() && colorId.type == NONE`. Reusing `bad-txns-nonstandard-opcolor` as the error string is fine.

doc/tapyrus/colored_coin.md
```diff
@@ -24,25 +24,31 @@ Script which does not include `OP_COLOR` represents to the amount of the default
 In a transaction, the total amount of tokens in the input color identifier and the total amount of tokens in the output color identifier shall be the same.
```
azuchi: **M3 — Conflicts with the new L93 statement**

L24 says input and output amounts "shall be the same", but the new L93 says output "must not exceed" input. The implementation matches L93 (input ≥ output is allowed; the surplus is silently destroyed — the existing implicit burn primitive). L24 should be rewritten to match, e.g.

> The sum of output `nValue` for any color identifier shall not exceed the sum of input `nValue` for that identifier. Any unspent surplus is permanently destroyed.

doc/tapyrus/colored_coin.md
```diff
@@ -24,25 +24,31 @@ Script which does not include `OP_COLOR` represents to the amount of the default
...
 That is, it maintains the balance of each token in the transaction. Please note that Signer does not receive any tokens (except default token) as a fee for block generation at this stage.
 
+For outputs whose scriptPubKey contains `OP_COLOR`, the `nValue` field carries the quantity of the named token as a raw integer, not satoshis. For outputs whose scriptPubKey does not contain `OP_COLOR`, `nValue` carries TPC in satoshis. Token `nValue` is not counted toward TPC transaction fees. Accordingly, every token transaction is required to include at least one TPC-denominated input to cover the fee; a transaction that contains only colored inputs is rejected with `bad-txns-token-without-fee`.
+
 If you use `OP_COLOR` in Script, the following restrictions are applied.
 
 * Only one `OP_COLOR` can be included in a script.
-* A `OP_COLOR` cannot be written in a control opcode such as `OP_IF`.
+* A `OP_COLOR` cannot be written in a control opcode such as `OP_IF`, `OP_CHECKLOCKTIMEVERIFY`, `OP_CHECKSEQUENCEVERIFY` etc.
```
azuchi: **M4 — Wording issue: CLTV/CSV are not control opcodes**

`OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` don't contain other opcodes, so "OP_COLOR written in CLTV" has no well-defined meaning. The intent (from the surrounding lines and the new template restriction) is: **the scriptPubKey itself must match exactly one of CP2PKH / CP2SH / Burn**, and CLTV/CSV are forbidden in the scriptPubKey precisely because they break that template match.

Suggested rewrite:

```markdown
* Only one `OP_COLOR` is allowed in a scriptPubKey.
* The scriptPubKey of a colored output must match exactly one of the templates
  listed below (CP2PKH / CP2SH / Burn). Other opcodes (`OP_IF`,
  `OP_CHECKLOCKTIMEVERIFY`, `OP_CHECKSEQUENCEVERIFY`, `OP_SWAP`, `OP_DROP`,
  etc.) are not permitted in the scriptPubKey of a colored output.
* These opcodes may freely appear inside a CP2SH redeem script, which is
  evaluated only at spend time and does not contain `OP_COLOR`.
```

src/chainstate.cpp
```diff
@@ -23,6 +23,8 @@
 // Defined in validation.cpp; declared here to avoid pulling in validation.h
 // (which includes chainstate.h, creating an indirect self-inclusion).
 void RefreshChainTxDataFromTip(const CBlockIndex* pindexNew);
+extern Mutex cs_issued_colorids;
+extern std::set<ColorIdentifier> g_issued_colorids;
```
azuchi, line 27: **S1 — Duplicate extern declarations**

These are also declared in `src/validation.h:154-158`. The comment at L22-23 explains the duplication is intentional to avoid the `validation.h` → `chainstate.h` self-inclusion. If that constraint is still real, keep as-is; otherwise consolidate via `#include <validation.h>`.

src/txdb.cpp
```diff
@@ -430,3 +431,27 @@ bool CCoinsViewDB::Upgrade() {
...
+    pcursor->Seek(std::make_pair(DB_ISSUED_COLORID, std::vector<unsigned char>{}));
+    while (pcursor->Valid()) {
+        std::pair<char, std::vector<unsigned char>> key;
+        if (!pcursor->GetKey(key) || key.first != DB_ISSUED_COLORID)
+            break;
+        colorIds.insert(ColorIdentifier(key.second));
+        pcursor->Next();
+    }
+    return true;
+}
```
azuchi, line 482: **S2 — Iterator errors are silently swallowed**

This function always returns `true`. If the LevelDB iterator hits a corruption error mid-scan, the partial set is silently accepted and the node starts with an incomplete `g_issued_colorids`, which would let a previously-seen NFT colorId be re-issued.

Suggested:
```cpp
while (pcursor->Valid()) { ... }
return !pcursor->GetError();
```

The `init.cpp` call site already treats a `false` return as fatal, so propagating the iterator error would surface as a clean startup failure rather than silent corruption.

test/functional/feature_colored_coin_vuln.py
```diff
@@ -0,0 +1,843 @@
...
+        # waiting for a pong that will never arrive.  If the peer was NOT
+        # disconnected, fall back to sync_with_ping to let processing finish.
+        if not self.peer.is_connected:
+            pass  # already disconnected
+        else:
+            try:
+                self.peer.wait_for_disconnect(timeout=5)
+            except Exception:
+                # Not disconnected within 5 s — peer is still alive, use ping.
+                self.peer.sync_with_ping()
```
azuchi: **S3 — Fallback masks low-DoS rejections**

Falling back to `sync_with_ping` when the peer is not disconnected within 5 s hides the difference between "rejected with DoS=100" (peer banned) and "rejected with DoS=0" (peer kept alive).

Once M1 is fixed, every block-level rejection in this test should produce a disconnect — so logging a warning when this fallback path is taken would catch regressions where the DoS score is unintentionally lowered.

src/chainstate.cpp
Naviabheeman: Yes, this duplication still needs to be there because validation.h and chainstate.h share a lot of global data.

doc/tapyrus/colored_coin.md
Naviabheeman: This was picked up from a dummy script that burn token RPC uses . Later layers identify the transaction as a color coin transaction only when the color id is seen. So this is used temporarily until the colored is validated. Sorry, I overlooked this documentation change.

azuchi (commented):
Follow-up review for the `Review feedback` commit. Most prior review items are correctly addressed (TokenTypes underlying type, DoS-score wiring inside `CheckColorIdentifierValidity`, removal of `REJECT_COLORID`, doc/spec alignment, vuln tests merged into `feature_coloredcoin.py`).

However this commit introduces a hard regression in `BurnToken`: the new validation rejects every colored output that is not CP2PKH/CP2SH, but `BurnToken` still emits a 36-byte `<colorId> OP_COLOR OP_FALSE` script, so every `burntoken` RPC and the GUI burn flow are now unconditionally rejected with `bad-txns-nonstandard-opcolor`. This is the cause of the failing CI jobs (`L-*-GUI+*Wallet`, `M-*-GUI+Wallet`, etc.) — `wallettests.cpp:392` and `:504` fail because the burn tx never reaches the wallet as a `TokenBurn` record.

Details inline.

src/wallet/rpcwallet.cpp
```diff
@@ -4585,7 +4585,7 @@ UniValue transfertoken(const JSONRPCRequest& request)
 CTransactionRef BurnToken(CWallet * const pwallet, const ColorIdentifier& colorId, CAmount nValue)
 {
     //this is to make create transaction recognize this as colored transaction
-    CScript scriptPubKey = CScript() << colorId.toVector() << OP_COLOR << OP_TRUE;
+    CScript scriptPubKey = CScript() << colorId.toVector() << OP_COLOR << OP_FALSE;
```
azuchi: **Critical — `burntoken` is broken; this is what is failing CI.**

With the new spec only CP2PKH (60 B) and CP2SH (58 B) colored scripts are accepted; `CheckColorIdentifierValidity` rejects anything else with DoS=100 / `bad-txns-nonstandard-opcolor`. The script built here is 36 B (`0x21 <33B colorId> OP_COLOR OP_FALSE`) and matches neither template, so every burn tx produced by this function is now unconditionally rejected by both mempool and `ConnectBlock`.

Flipping `OP_TRUE` → `OP_FALSE` doesn't help — the issue is the template, not the final opcode. `transactionrecord.cpp` already aligns with the new spec (`isBurn = debitAmt > colorOutputTotal`), i.e. burn = *no colored output*. `BurnToken` must do the same: spend the colored input and only emit TPC change (no colored output for that colorId). Alternatively, deprecate the `burntoken` RPC entirely since users can achieve the burn by simply omitting the colored output in a regular `transfertoken`-like flow.

CI evidence (job `L-ARM-Deb-GUI+USDT+Wallet`, run 25014198947):
```
FAIL!  : WalletTests::walletTests() No record of type 8 found for colorId c2fbad15…
   Loc: …/src/qt/test/wallettests.cpp(392)
FAIL!  : WalletTests::walletTests() Compared values are not the same
   Loc: …/src/qt/test/wallettests.cpp(504)
```
(type 8 = `TransactionRecord::TokenBurn`; line 504 is the `combo->findData(cid) == -1` check that assumes the burn went through and the zero-balance token disappeared.)

src/script/script.h
```diff
@@ -542,6 +542,8 @@ class CScript : public CScriptBase
 
     bool IsColoredScript() const;
     bool IsColoredPayToScriptHash() const;
+    bool IsColoredPayToPubkeyHash(std::vector<unsigned char>& pubkeyhash, std::vector<unsigned char>& colorid) const;
+    bool IsColoredBurnScript() const;
```
azuchi: Orphaned declaration: `IsColoredBurnScript()`'s implementation was removed from `script.cpp` in this commit, but the declaration is still here. Nothing references it (verified with `git grep IsColoredBurnScript`), so no link error today, but the dangling declaration will mislead future readers. Please remove this line.

src/dbwrapper.h
```diff
@@ -171,6 +171,10 @@ class CDBIterator
         return piter->value().size();
     }
 
+    bool GetError() {
```
azuchi: Minor naming nit: `GetError()` returns `true` when there *is* an error, but the `Get*` prefix usually implies a value-fetching operation that returns `true` on success. The only caller — `CBlockTreeDB::LoadIssuedColorIds` — ends up writing `return !pcursor->GetError();` which is a double negative. `HasError()` (or `Status()` returning `status().ok()`) would read more naturally on both sides.

src/script/script.h
Naviabheeman: This is a rebase confusion, this declaration is removed in my local repository. I'll fix it.

src/chainstate.cpp
```diff
@@ -657,6 +700,20 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
...
+    // global set and to LevelDB — no further failure can occur that would
+    // require rolling these back within this call.
+    if (!fJustCheck) {
+        for (const auto& colorId : allNewIssuances) {
+            {
+                LOCK(cs_issued_colorids);
+                g_issued_colorids.insert(colorId);
+            }
+            if (!pblocktree->WriteIssuedColorId(colorId))
+                return error("ConnectBlock(): WriteIssuedColorId failed");
```
azuchi: **RAM/disk inconsistency on `WriteIssuedColorId` failure**

```cpp
{
    LOCK(cs_issued_colorids);
    g_issued_colorids.insert(colorId);
}
if (!pblocktree->WriteIssuedColorId(colorId))
    return error("ConnectBlock(): WriteIssuedColorId failed");
```

If `WriteIssuedColorId` fails after some colorIds in `allNewIssuances` have already been inserted into `g_issued_colorids`, the in-memory set holds entries that are not persisted on disk. The block connect fails and is unwound, but the in-memory set is not rolled back — subsequent legitimate issuances of those colorIds will then be rejected with `bad-txns-colorid-already-issued` for the remainder of the process lifetime.

`DisconnectBlock` (lines 308-310) has the symmetric issue: in-memory erasure happens immediately, before the disk-side erase.

**Suggested fix:** stage all writes in a `CDBBatch`, commit it, then mutate `g_issued_colorids` only after the batch succeeds. This way the in-memory set is updated iff the disk state is, and any partial-failure path leaves both stores consistent.

src/init.cpp
```diff
@@ -1554,6 +1554,17 @@ bool AppInitMain()
...
 
+                // Step 7b: Load issued NON_REISSUABLE/NFT colorId set from db
+                {
+                    std::set<ColorIdentifier> issuedColorIds;
+                    if (!pblocktree->LoadIssuedColorIds(issuedColorIds)) {
+                        strLoadError = _("Failed to load issued colorId set from block database");
+                        break;
+                    }
+                    LOCK(cs_issued_colorids);
+                    g_issued_colorids = std::move(issuedColorIds);
```
azuchi: **Reindex: `g_issued_colorids` may be stale-loaded before blocks are re-connected**

`-reindex-chainstate` preserves the block tree DB, so on startup `LoadIssuedColorIds` populates `g_issued_colorids` with every NON_REISSUABLE/NFT colorId from the previous run. The chainstate is then rebuilt by re-connecting blocks. When a block that originally issued one of those colorIds is reconnected, `CheckColorIdentifierValidity` will see the colorId already in the global set and reject the block with `bad-txns-colorid-already-issued` — even though the block is the legitimate origin of that colorId.

The same hazard applies if a reorg replays an old issuance that was previously in `g_issued_colorids` but has since been disconnected — `DisconnectBlock` does erase entries, so a clean reorg should be fine, but reindex bypasses that path.

**Suggested fix (one of):**
1. Clear `g_issued_colorids` (and the on-disk `DB_ISSUED_COLORID` entries) when `fReindex || fReindexChainState` is set, before this load runs.
2. Skip the uniqueness check inside `CheckColorIdentifierValidity` while the chainstate is being rebuilt (e.g. via the same flag used to suppress duplicate-mempool checks during IBD).

Option 1 is simpler and matches how `pblocktree->ReadReindexing()` is handled elsewhere.

src/validation.cpp
```diff
@@ -439,45 +441,80 @@ static bool DoAllInputsExist(const CTransaction& tx, CValidationState& state, CT
...
+            // Burn script: 0x21 <33B colorId> OP_COLOR <1B opcode> — the
+            // 36-byte placeholder used by BurnToken (terminal opcode not
+            // compared since it may vary).  Balance is enforced by
+            // VerifyTokenBalances; skip structural validation here.
+            bool isBurnForm = (s.size() == 36 && s[0] == 0x21 && s[34] == OP_COLOR &&
+                               (s[1] == TokenToUint(TokenTypes::REISSUABLE) ||
+                                s[1] == TokenToUint(TokenTypes::NON_REISSUABLE) ||
+                                s[1] == TokenToUint(TokenTypes::NFT)));
+            if (isBurnForm)
+                continue;
```
azuchi: **The 36-byte burn placeholder script is unnecessary at the consensus layer.**

`doc/tapyrus/colored_coin.md:86-92` defines burn as:

> create a transaction with a UTXO with the token to be burned and a UTXO with a TPC for the fee as inputs, and add an output that receives TPC change.

i.e. the spec-defined burn is **"colored input present, colored output absent for that colorId"** — the burn amount is just the implicit input/output difference. Consensus already enforces only `output ≤ input` per colorId in `VerifyTokenBalances` (`out.second > iter->second` check around line 622), not equality, so the difference (`input − output > 0`) is allowed and is the burn.

The `<colorId> OP_COLOR OP_TRUE` placeholder originated as an internal wallet artifact so that `CreateTransaction` could discover which colorId to spend from the recipient list. With this PR it leaks into the on-chain tx, and to keep that valid the consensus layer needs:

1. The 36-byte structural exception in `CheckColorIdentifierValidity` (lines 462-471) — which also bypasses `nValue > 0`, the `matchFound` colorId-derivation check, the NFT `nValue == 1` check, and the `g_issued_colorids` re-issuance check (validation.cpp:484-552)
2. The mirror burn-script colorId extraction in `VerifyTokenBalances` (lines 587-595)
3. A redundant fallback path in `CWallet::CreateTransaction` (wallet.cpp:2643-2646)

All three exist only to make placeholder outputs valid. Consequences of (1) in particular:

- The 33-byte payload after the type byte is unchecked → an attacker with TPC can craft `0x21 || 0xc2 || <arbitrary 33B> || OP_COLOR || <any byte>` and consensus accepts it as a NON_REISSUABLE issuance funded by TPC sacrifice. No token inflation results (`g_issued_colorids` stays clean because the issuance map uses `GetColorIdFromScript`, which returns NONE for burn scripts), but the consensus rule is admitting arbitrary-payload outputs at the cost of TPC + fee.
- The terminal opcode (`s[35]`) is unconstrained — the comment says "may vary" but `BurnToken` always emits `OP_TRUE`. As written `OP_FALSE`, `OP_RESERVED`, `OP_INVALIDOPCODE` (0xff), any `OP_NOP*` are all accepted.
- (1) and (2) are inconsistent with each other (the type byte 0xc1/0xc2/0xc3 is checked in the former but not the latter); they only happen to behave correctly because `CheckColorIdentifierValidity` is called first.

**Suggested direction (eliminates all of the above)**:

- Drop the placeholder recipient in `BurnToken` (rpcwallet.cpp:4585-4588) and construct the burn tx as the spec describes: colored input + TPC input → TPC change output only, no colored output.
- Drive colored-input selection in `CreateTransaction` from `coin_control.m_colorTxType == BURN` + `coin_control.m_colorId` directly (the fallback you already added in this commit becomes the primary path; the `vecSend` placeholder loop can be skipped for the burn case).
- Remove the `isBurnForm` block in `CheckColorIdentifierValidity` (validation.cpp:462-471).
- Remove the burn-script colorId extraction in `VerifyTokenBalances` (validation.cpp:587-595).
- `doc/tapyrus/colored_coin.md` needs no change — the existing "Burn Token" section already describes this construction.

Net effect: spec, consensus, and wallet all align on "burn = input/output difference", and there is no third script form for re-implementations to special-case.

azuchi (commented):
Latest commit `d283d73` addresses every item from the previous review:

- `BurnToken` regression fixed (revert of `OP_FALSE` → `OP_TRUE`, plus burn-script bypass in `CheckColorIdentifierValidity` and `coin_control.m_colorTxType == BURN` fallback in `CWallet::CreateTransaction`)
- `IsColoredBurnScript()` orphan declaration removed
- `GetError()` renamed to `HasError()` (call site updated)

CI is fully green. Thank you for the quick turnaround.

One larger follow-up below — I think the new burn-script consensus exception is unnecessary and should be removed in favor of the spec-defined behavior.

PR #411, opened 2026-05-11
Fix cp2sh vulnerabilities identified in security audit

Naviabheeman:
1: P0 (HIGH): CP2SH redeem script is never executed

  CScript::IsPayToScriptHash() is hard-coded to the bare 23-byte form (OP_HASH160 <20B> OP_EQUAL). A CP2SH scriptPubKey is 58 bytes (<colorId(33)> OP_COLOR OP_HASH160 <20B> OP_EQUAL), so this guard is never satisfied for CP2SH. There is no parallel IsColoredPayToScriptHash() branch in VerifyScript. The redeem script is therefore never executed for any CP2SH spend.
  The wallet exhibits the symmetric bug: ProduceSignature only invokes the second SignStep (which signs the inner script) when whichType == TX_SCRIPTHASH. For TX_COLOR_SCRIPTHASH, the inner sign step is skipped, and the wallet emits a scriptSig that consists solely of a push of the redeem-script bytes. The wallet then verifies its own output via VerifyScript, which (per the bug above) accepts it.
  
  2. P1 (MEDIUM): xfield history is not rewound on DisconnectTip         

  ConnectTip appends xfield rotations to both the in-memory g_chainstate.xfieldHistory map and to the on-disk pblocktree via WriteXField. DisconnectTip updates the UTXO view, mempool, and chainActive, but leaves xfield history untouched — there is an explicit TODO at line 808: //TODO: if xfield information change is being discarded during this reorg remove it from xFieldHistory here.
  Get(type, height) (xfieldhistory.cpp:40) iterates the history list linearly and returns the first entry whose height field matches. There is no de-duplication by height and no active-chain filter.

3. P2 (LOW): CheckBlock reads xfield-related parameters using attacker-controlled block.GetHeight()

  block.GetHeight() returns block.vtx[0].vin[0].prevout.n — the coinbase author chooses this value. isBlockHeightInCoinbase (validation.cpp:1795-1819) only enforces equality with the actual height when the block is adjacent to the current tip (same hash, builds on tip, or tip-1). For "unrelated" blocks the function returns true unconditionally, with the explicit comment // if the two blocks are unrelated, we assume the block height is valid. 
  CheckBlock then feeds this attacker-controllable height into: 
  - CXFieldHistory::Get(MAXBLOCKSIZE, height) — picks the maximum-block-size policy
  - CheckBlockHeader(..., height, fCheckPOW) — internally calls Get(AGGPUBKEY, height) to pick the aggregate public key for signature verification


azuchi (changes requested):
Thanks for the comprehensive fix — the P0 (`VerifyScript` / `ProduceSignature`) and P1 (`DisconnectTip` xfield rewind) changes look correct, and the new unit tests for CLTV / CSV / OP_COLOR-rejection / OP_IF inside CP2SH are excellent coverage.

However the P2 change (the `int nHeight = -1` parameter on `CheckBlock`) introduces a regression that breaks node startup on any chain that has had an xfield rotation. This is also the cause of the failing `feature_federation_management.py` CI job.

## Showstopper: `VerifyDB` regression (`src/verifydb.cpp:61`)

```cpp
if (nCheckLevel >= 1 && !CheckBlock(block, state))
```

This call relies on the new `nHeight = -1` default. Per the new `CheckBlock` body:

```cpp
uint32_t height = (nHeight >= 0) ? static_cast<uint32_t>(nHeight) : UINT32_MAX;
```

so `Get(AGGPUBKEY, UINT32_MAX)` returns the *latest* aggregate pubkey. `VerifyDB` walks the tail of the chain backwards and must validate each historical block under the aggregate pubkey that was active at *that block's* height, not the latest one. After any xfield rotation, every block below the most recent rotation fails with `bad-proof`.

The failing CI job demonstrates this:

```
node2 ERROR: VerifyDB: *** found bad block at 36, hash=4fa5014a...
       (bad-proof, Proof verification failed at height [-1] (code 16))
```

Default `-checklevel=3 -checkblocks=6` exercises this on every node restart, so any production node upgrading to this version will fail to start once it has connected even one block past a rotation.

`pindex->nHeight` is in scope at the call site, so the fix is one line:

```cpp
if (nCheckLevel >= 1 && !CheckBlock(block, state, true, true, nullptr, pindex->nHeight))
```

## Same issue in compact block reconstruction (`src/blockencodings.cpp:216`)

```cpp
if (!CheckBlock(block, state, true, true)) {
```

This runs after a compact block has been reassembled. The block's height is known from the header that was already accepted (the corresponding `CBlockIndex` exists in `mapBlockIndex` at this point — the compact block path is gated by header-first acceptance). With `nHeight = -1` it falls back to UINT32_MAX and will reject any compact block at a height before the most recent rotation if the `Get(MAXBLOCKSIZE, ...)` or `Get(AGGPUBKEY, ...)` lookups disagree with the latest entry.

Fix: thread the height through `PartiallyDownloadedBlock::FillBlock()` so the caller can pass `pindex->nHeight`, e.g.

```cpp
if (!CheckBlock(block, state, true, true, nullptr, pindex->nHeight)) {
```

(or pass `nHeight` as an explicit parameter to `FillBlock`, since `pindex` is held by the caller in `net_processing.cpp`).

## Defensible but worth tightening: `ConnectBlock(fJustCheck=true)` (`src/chainstate.cpp:547`)

```cpp
if (!CheckBlock(block, state, !fJustCheck, !fJustCheck)) {
```

This call is inside `ConnectBlock`, which already has `pindex` (the block being connected). Passing `pindex->nHeight` here is a small defensive improvement and makes the height-flow self-consistent across all call sites.

## Other CheckBlock call sites — already correct or intentional

| Call site | nHeight | Verdict |
|---|---|---|
| `chainstate.cpp:1457` (`AcceptBlock`) | `pindex->pprev ? pindex->pprev->nHeight + 1 : 0` | ✓ |
| `validation.cpp:1585` (`TestBlockValidity`) | `pindexPrev->nHeight + 1` | ✓ |
| `validation.cpp:1548` (`ProcessNewBlock` pre-lock) | default `-1` | ✓ — new blocks should validate under the latest key, this is the documented intent |
| `tapyrus-genesis.cpp:116` | default `-1` | ✓ — genesis tool, special context |
| `bench/checkblock.cpp:51` | default `-1` | ✓ — benchmark only |

## Minor: removed JSON test vectors (`src/test/data/script_tests.json`)

The two `CP2SH(P2PKH) ECDSA` / `CP2SH(P2PKH) SCHNORR` vectors that were removed presumably no longer pass under the strict redeem-script execution (their pre-computed sigs would not satisfy the inner P2PKH script under the corrected sighash). Recomputing the signatures and keeping these vectors as positive examples would be preferable to deletion — they exercise the JSON-driven path which the new C++ test (`colored_coin_cp2sh_redeem_scripts`) does not. Not a blocker.

## Summary

- ✅ P0 (VerifyScript + ProduceSignature)
- ✅ P1 (DisconnectTip xfield rewind, Get(height) reverse-scan, init.cpp ordering)
- ❌ P2 (CheckBlock nHeight) — `VerifyDB` and `blockencodings::FillBlock` are missing the new parameter and break on rotation-experienced chains

Please address `verifydb.cpp:61` and `blockencodings.cpp:216` (and ideally `chainstate.cpp:547`) before merge. The CI failure should clear automatically once the VerifyDB call passes the height through.

azuchi (approved):

PR #412, opened 2026-05-13
Fix security vulnerabilities identified in audit

Naviabheeman:
K1 — Crash-consistency split between pblocktree and chainstate for issued colorIds

pblocktree (block-tree LevelDB at blocks/index/) and pcoinsdbview (chainstate LevelDB at chainstate/) are independent LevelDB instances with independent crash recovery. ConnectBlock writes new issued colorIds to pblocktree on every block, while pcoinsdbview is only flushed on FlushStateToDisk (PERIODIC / IF_NEEDED / shutdown).
  On crash, pblocktree typically retains the colorId(s) for block N, while pcoinsdbview is at N-1. After restart: 
  1. LoadIssuedColorIds populates g_issued_colorids from pblocktree → contains X. 
  2. ReplayBlocks only reconciles UTXO state, not g_issued_colorids.
  3. ActivateBestChain re-connects N → CheckColorIdentifierValidity → g_issued_colorids.count(X) > 0 → bad-txns-colorid-already-issued.
  4. Block N is permanently marked BLOCK_FAILED_VALID. Recovery requires -reindex-chainstate. 
  Not actively attacker-triggerable, but a real operational hazard. feature_dbcrash does not exercise this path because it uses TPC-only transactions.

K2 — VerifyDB level 3 mutates global state through what is intended to be a view-only sandbox 

VerifyDB level 3 (the default — DEFAULT_CHECKLEVEL = 3) walks back from tip calling g_chainstate.DisconnectBlock(block, pindex, coins), where coins is a local CCoinsViewCache intended as a throwaway sandbox. But the colorId-erase code added inside DisconnectBlock writes to pblocktree->EraseIssuedColorId (global LevelDB) and g_issued_colorids.erase (global memory), ignoring the view parameter.
  After level 3 finishes, the global colorId state is missing the last nCheckDepth blocks' issuances. chainActive is unchanged (no real disconnect happened). Level 4's reconnect would restore the state, but level 4 only runs at -checklevel=4, not at the default 3.
  Token inflation is not possible because NON_REISSUABLE/NFT colorIds derive from Hash(prevout) and the prevouts are spent. But the drift accumulates across restarts and may surface to any consumer that reads g_issued_colorids.

K3 — ConnectBlock commits issued colorIds before WriteUndoDataForBlock, so partial-write failures leave colorIds permanently in pblocktree

ConnectBlock writes the new colorId set to pblocktree and g_issued_colorids immediately after control.Wait() succeeds, then attempts WriteUndoDataForBlock. If the undo write fails (disk full, EIO, fd exhaustion, -prune race that just unlinked the target rev*.dat), ConnectBlock returns false without setting state.IsInvalid().
  ConnectTip only marks the block BLOCK_FAILED_VALID when state.IsInvalid() is true, so the block is rolled back from the in-memory chain but the colorIds remain on disk and in memory. On the next ActivateBestChain retry, CheckColorIdentifierValidity rejects the same block with bad-txns-colorid-already-issued, and this time state.IsInvalid() is set, permanently marking the block failed.
  The author's comment on the colorId commit ("no further failure can occur that would require rolling these back within this call") is incorrect — WriteUndoDataForBlock, pindex->RaiseValidity, and setDirtyBlockIndex.insert all run after.

H1 — getcolor RPC: signed-narrowing OOB read on vout parameter 

get_int() returns int32_t and is silently narrowed to int8_t (range -128..127). The bounds check casts tx->vout.size() to int8_t and compares signed-vs-signed, so -1 >= 1 is false and the check is bypassed. tx->vout[vout] then indexes std::vector with a negative value (implicitly converted to size_t), producing an out-of-bounds memory read — typically SEGV. A single authenticated short POST crashes the daemon.
  The same int → uint8_t narrowing pattern exists in issuetoken (src/wallet/rpcwallet.cpp:4396). That one does not crash but silently wraps vout=256 to vout=0, issuing tokens against the wrong UTXO.

H2 — issuetoken / reissuetoken / burntoken skip the wallet-passphrase gate 

Unlike sendtoaddress / sendmany / signrawtransactionwithwallet etc., these three Tapyrus-specific RPCs do not call EnsureWalletIsUnlocked(). The flow proceeds into IssueReissuableToken / IssueToken / BurnToken and fails late with a generic "Signing transaction failed" message instead of the documented RPC_WALLET_UNLOCK_NEEDED. 
  Worse, IssueReissuableToken constructs and broadcasts tx1 first, then constructs tx2. If the wallet was unlocked at tx1 but auto-relocked before tx2 (auto-relock window), tx1 is committed to the chain while tx2 fails to sign. The orphan UTXO created by tx1 may be hard for the user to spend cleanly.


azuchi (dismissed):
## Review: PR #412 (K1/K2/K3/H1/H2 + ReadBlockFromDisk regression + companion fixes)

Diff against master: 22 files, +383/-194. Reviewed against the security-audit findings the PR description claims to address.

---

### Confirmed fixes

**K1 — crash-consistency split (`a85f9e14`)**

The new `CIssuedColorIds` class owns the in-memory issued set plus a pending Changeset; `CommitToBatch()` is invoked from `CCoinsViewDB::BatchWrite` so the colorId writes ride in the same `CDBBatch` as `DB_BEST_BLOCK`. Storage moves from `pblocktree` to `pcoinsdbview`, which means `-reindex` and `-reindex-chainstate` automatically wipe colorIds via `fWipe=true` on the chainstate DB — no more dedicated `ClearIssuedColorIds()` path. Lifecycle in `init.cpp` is correct: construct → `SetColorIdState` → `Upgrade` → `ReplayBlocks` (DisconnectBlock during replay can stage erases that ride the replay flush) → `LoadIssuedColorIds` + `SetConfirmed`. Shutdown order in `Shutdown()` releases `pcoinsdbview` before `g_colorid_state`, but `pcoinsdbview`'s destructor never accesses the raw pointer so no UAF.

**K3 — commit ordering (`a85f9e14`)**

`g_colorid_state->Insert(allNewIssuances)` now runs after `WriteUndoDataForBlock`, `RaiseValidity`, and `setDirtyBlockIndex.insert`. The previous "no further failure can occur" comment correctly describes the new placement.

**H1 — getcolor signed-narrowing OOB (`6b080227`)**

`int n = request.params[2].get_int(); if (n < 0 || (size_t)n >= tx->vout.size()) throw;` correctly handles both negative and out-of-range inputs before the unsigned cast. The sister fix in `issuetoken` only checks `n < 0`, which is sufficient because `COutPoint(txid, INT_MAX)` is harmless — the UTXO lookup will simply miss.

**H2 — wallet unlock gate (`6b080227`)**

`issuetoken` / `reissuetoken` / `burntoken` now call `EnsureWalletIsUnlocked`. `IssueReissuableToken` is restructured so all keypool work (TopUpKeyPool, GetKeyFromPool, colorDest derivation) happens up front under `LOCK2`, followed by an `IsLocked()` re-check before tx1 is broadcast. With `cs_wallet` held throughout, no auto-relock can interleave between tx1 and tx2. `SetAddressBook` is moved to after both commits succeed.

**ReadBlockFromDisk regression from PR #411 (`aef65ac6`)**

The `CDiskBlockPos` overload now takes `int nHeight = -1`, and the `CBlockIndex` overload passes `pindex->nHeight`. Trusted-source-only invariant for the height passed to `CheckBlockHeader` is preserved (block-data-derived `block.GetHeight()` is never used). Fixes the `getblock` and CI failures (`p2p_federation_block`, `p2p_sendheaders`) introduced by the previous `-1` sentinel.

**Companion bug fixes in `6b080227`**

- `ColorIdentifier::operator<` and `ColorIdentifierCompare::operator()` were calling `memcmp(this, &colorId, COLOR_IDENTIFIER_SIZE)` over the whole struct, which includes uninitialized padding — undefined behavior with potentially unstable ordering. The replacement compares `type` then `payload` explicitly. This is a real bug fix beyond the listed scope.
- `key_io.cpp` CP2SH decode was extracting the ColorIdentifier using `c_pubkey_prefix.size()` as the offset instead of `c_script_prefix.size()`. Real bug; CP2SH addresses with prefixes of differing length would have decoded the wrong bytes.
- `federationparams.cpp` magic-byte derivation moves from `int` to `uint32_t` to avoid signed-overflow UB for high `nNetworkId` values. Bytewise output is unchanged on twos-complement targets.
- Genesis pubkey parsing now throws on unknown leading bytes (defensive; previously silently accepted).
- `xfieldhistory.Get(type, blockHash)` overload removed — was unused.
- `LoadIssuedColorIds` skips `TokenTypes::NONE` entries (defensive against malformed DB rows).

**Test coverage**

`feature_dbcrash.py` now drives `issuetoken` (REISSUABLE / NON_REISSUABLE / NFT), `sendtoaddress` for colored UTXOs, and `burntoken` across the crash loop. With the smaller `-dbbatchsize=100000` and the early-exit when both restart counts are non-zero, this exercises K1 directly. Comments in `feature_coloredcoin.py` for `test_reindex_chainstate_colorid_set` / `test_reindex_colorid_set` are updated to describe the new (correct) behavior.

---

### 🔴 Blocking: K2 fix breaks VerifyDB Level 4

`src/chainstate.cpp:312` (the K2 fix):
```cpp
if (!fDryRun) {
    g_colorid_state->Erase(outColorId);
}
```
skips the `m_confirmed` erase entirely when dryrun is true, while the rest of `DisconnectBlock` continues to mutate the local `coins` view. After Level 3's reverse walk, `m_confirmed` therefore still contains every NON_REISSUABLE/NFT colorId issued in the last `nCheckDepth` blocks.

`src/verifydb.cpp:107` (Level 4 forward walk):
```cpp
if (!g_chainstate.ConnectBlock(block, state, pindex, coins))
    return error("VerifyDB(): *** found unconnectable block at %d, hash=%s (%s)", ...);
```
goes through `ConnectBlock`, which calls `CheckColorIdentifierValidity` (`src/chainstate.cpp:677`). Inside, in `validation.cpp`:
```cpp
if (g_colorid_state && g_colorid_state->IsIssued(outColorId))
    return state.DoS(100, false, REJECT_INVALID, "bad-txns-colorid-already-issued");
```
returns true because `m_confirmed` was never rolled back. Every Level 4 reconnect of a block with a NON_REISSUABLE/NFT issuance will now fail with `bad-txns-colorid-already-issued`, causing `verifychain 4` to error out.

**Impact:** default `-checklevel=3` users are unaffected. Anyone running with `-checklevel=4`, or invoking `verifychain 4` over RPC, on a chain that has issued any NON_REISSUABLE/NFT token, will see VerifyDB fail spuriously. `rpc_blockchain.py:61` does call `verifychain(4, 0)` but its chain has no token issuances, so CI does not catch this.

**Suggested fixes (any one):**

1. **Memory-only erase under dryrun:** in dryrun mode, perform `m_confirmed.erase(id)` but do not push to `m_pending`. After VerifyDB returns, the in-memory set is back at tip state via the Level 4 reconnects' `Insert()` (or, if `-checklevel=3`, the in-memory drift is benign because re-issuance of NON_REISSUABLE/NFT requires reusing a spent prevout, which is impossible). Smallest patch.
2. **Sandbox `m_confirmed` alongside `coins`:** maintain a per-VerifyDB scratch set that overlays the global, and let Level 3/4 mutate the scratch only. Cleanest invariant; more code.
3. **Plumb a `fSkipIssuedCheck` (or generic `fJustCheck`-like) flag through `ConnectBlock` → `CheckColorIdentifierValidity`** and set it from VerifyDB Level 4. Targeted but adds a knob to a hot path.

---

### Non-blocking observations

- **`RollforwardBlock` does not stage colorIds.** During post-crash replay (`DB_HEAD_BLOCKS={N, N-1}`), UTXOs are rolled forward to N but `g_colorid_state->Insert()` is not called for any NON_REISSUABLE/NFT issuance in block N. After `SetConfirmed` reads the disk state (which only reflects ≤ N-1), `m_confirmed` is permanently missing block N's colorIds until a future restart. Functionally harmless because re-issuance of NON_REISSUABLE/NFT colorIds requires reusing a spent prevout, which is impossible by construction — but the documentation "single source of truth ... ever issued on-chain" is technically violated post-crash. Worth a comment, or a `RollforwardBlock` Insert pass for the same issuance set.
- **`file_io.cpp:317`** (`LoadExternalBlockFile` recursive child path) still uses default `nHeight=-1` because pindex isn't yet known. On a rotation chain this can cause valid pre-rotation blocks to be rejected during reindex. Inherited from the design (pre-existing constraint), out of scope for this PR but tracking is worthwhile.
- **Idempotent re-writes during VerifyDB Level 4** (assuming the K2 regression is fixed): each reconnect calls `Insert()` for colorIds already present in disk and `m_confirmed`. Wasteful but not incorrect — the next BatchWrite re-writes the same key/value.
- **Commit-message inaccuracy on `a85f9e14`:** the headline says "feature_dbcrash.py failure" was caused by the partial-write loop, which is accurate; just noting it for completeness.
- **Branch carries PR #411's commits (`086cbd18` … `3cfad3cd`)** in addition to its own work because PR #411 was merged via squash. Harmless if PR #412 is squashed too.

---

Requesting changes for the K2 / Level 4 regression. Once that is addressed, the rest of the PR looks ready to land.

azuchi (commented):
## Re-review: PR #412 update (`6f82f6d38`)

The new commit addresses the prior CHANGES_REQUESTED blocker (K2 / VerifyDB Level 4 regression) cleanly.

### Verified

- `DisconnectBlock` reverts to the unconditional `g_colorid_state->Erase(outColorId)` path; the `fDryRun` parameter is removed from both declaration and call sites.
- New `CIssuedColorIds::Clone()` returns a deep copy of `m_confirmed` with empty `m_pending` (cs_main-guarded).
- RAII `ColorIdSandbox` in `verifydb.cpp` swaps `g_colorid_state` for a clone for the duration of Level 3 + Level 4, and the destructor restores the original. Exception-safe; the destructor runs on every exit path including the early `error()` returns inside the verify loop.
- VerifyDB never calls `coins.Flush()`, so `CCoinsViewDB::BatchWrite` → `m_colorid_state->CommitToBatch()` is not reached during the sandbox window. The fact that `pcoinsdbview->m_colorid_state` (raw pointer) still points to the original `saved` instance rather than the clone is therefore not exercised; the staged pending changes on the clone get discarded along with the clone itself. Confirmed correct.
- cs_main is held throughout `VerifyDB`, so the brief moment between `std::move(g_colorid_state)` and `g_colorid_state = saved->Clone()` (during which `g_colorid_state` is nullptr) cannot be observed by any other code path.
- All other prior observations remain in the same state: pre-existing inherited constraints (`file_io.cpp:317` orphan-recursive nHeight, `RollforwardBlock` not staging colorIds) are unchanged but were already noted as non-actionable / functionally harmless in the previous review.

### Single non-blocking request: regression test

The sandbox fix is correct but isn't caught by any existing test. `rpc_blockchain.py:61` calls `verifychain(4, 0)` but its chain has no token issuances, so the original bug would not have surfaced there either. To lock this in, please add a one-line check to a colored-coin functional test — e.g. at the end of an existing scenario in `feature_coloredcoin.py` (or `feature_dbcrash.py`) after at least one NON_REISSUABLE/NFT issuance has been confirmed:

```python
assert_equal(node.verifychain(4, 0), True)
```

With the bug pre-`6f82f6d38`, that assertion would have failed with `bad-txns-colorid-already-issued`; with the sandbox in place, it should pass. Cheapest possible regression net for what was the blocking issue on the prior round.

Not blocking — happy for this to land as soon as CI is green and the test is added (or as a small follow-up). Dismissing the prior `CHANGES_REQUESTED` so this isn't held up procedurally.

azuchi (approved):
LGTM — approving.

The new commit `42a5d479f` adds exactly the regression test I requested in the previous COMMENT review:

```python
# Regression: verifyDB level 4 reconnects each block to validate it.
# Before the sandbox fix, reconnecting a block with a NON_REISSUABLE/NFT
# issuance would fail with bad-txns-colorid-already-issued because the
# colorId was still in g_issued_colorids.
assert_equal(node.verifychain(4, 0), True)
```

Placement is correct: it lives at the tail of `test_reindex_colorid_set()` in `test/functional/feature_coloredcoin.py`, after `-reindex` and after a fresh NFT issuance. At that point the chainstate contains multiple NON_REISSUABLE and NFT issuances, so `verifychain(4, 0)` exercises the Level-4 reconnect path against the exact precondition that triggered the original failure. The method is invoked unconditionally from `run_test()`, so the assertion runs in CI.

If the sandbox in `verifydb.cpp` were ever removed or weakened (e.g., if a future change moved `g_colorid_state` mutation back outside the sandbox), this assertion would fail with `bad-txns-colorid-already-issued`. That is exactly the canary we wanted.

All prior CHANGES_REQUESTED items are addressed:
- K1 (colorId crash-consistency split) — fixed by atomic CommitToBatch in `a85f9e14c`
- K2 (VerifyDB Level 4 mutates global state) — fixed by the `ColorIdSandbox` RAII in `6f82f6d38`, now covered by the new regression test in `42a5d479f`
- K3, H1, H2 — addressed earlier in the PR

Ready to merge once CI is green.

PR #413, opened 2026-05-15
Fix security vulnerabilities ij

Naviabheeman:
Fix
J1: Out-of-bounds read in IsColoredOutPointWith triggered via fundrawtransaction / walletcreatefundedpsbt 

  IsColoredOutPointWith dereferences wtx->tx->vout[outpoint.n].scriptPubKey without checking either that GetWalletTx returned non-null or that outpoint.n < vout.size(). The bounds check inside SelectCoins (if (pcoin->tx->vout.size() <= outpoint.n) return false;) sits after the IsColoredOutPointWith call.

J2: bumpfee silently destroys colored tokens when bumping a token transfer 

  The change-detection loop in feebumper was inherited from upstream Bitcoin Core and assumes a single asset class. After colored-coin integration, IsChange correctly classifies a colored change output as "change", but the subsequent line 
  poutput->nValue -= nDelta;   // nDelta is a fee delta, in TPC satoshis
  is then executed against an output whose nValue field holds a token amount, not satoshis. The result is silent destruction of nDelta tokens.

I2 — CheckBlockHeader uses GetLatest() instead of Get(height) when pxfieldHistory is supplied

  When the temp-history branch is taken (pxfieldHistory != nullptr, the ProcessNewBlockHeaders path), the function fetches the aggregate pubkey via pxfieldHistory->GetLatest(...) rather than Get(AGGPUBKEY, nHeight). The non-temp branch on line 1339 uses Get(nHeight) — so the same header has different validity depending on which path validates it. 
  Concrete consequence: after a confirmed AGGPUBKEY rotation at active-chain height M+1 (history contains {K0, K1@M+1}), a legitimate fork header at height N ≤ M (which must be signed under K0) is checked against GetLatest() = K1, fails verification, and is rejected as bad-proof. Legitimate alternate chains (e.g., emergency rotation, federation split recovery) cannot be learned by this node over the wire, even though they would validate correctly during reindex (which uses the non-temp path). 

additional warnings fixed:
  -IsXFieldNew predicate (xfieldhistory.cpp:91-97) only compares against the last entry
  - mapRelay growth bound is chain-progression dependent (net_processing.cpp:3765, 15-minute prune).                                  
  - signrawtransactionwithwallet SCHNORR selection bug (rpcwallet.cpp:3122)
  - FundTransaction colored-only-change leaks the change address back to the keypool (wallet.cpp:2603-2613)  
  - importwallet does not reconstruct colored address-book entries (rpcdump.cpp:538-665)

I1 is not applicable to tapyrus as we cannot mallet the transaction id(hashmalfix) by creating a new signature. So the follow up warning related to vExtraTxnForCompact is also not possible.

other warnings related to colored coin can be included in the PR for soft fork activation:

- walletcreatefundedpsbt returns only TPC change position (rpcwallet.cpp:4117)
 - CP2SH issuance keying (rpcwallet.cpp:4238-4243, 4296-4299)

### DIFF
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -863,7 +863,7 @@ bool CChainState::ConnectTip(CValidationState& state, CBlockIndex* pindexNew, co
         CXFieldHistory xfieldHistory;
         if(blockConnecting.xfield.IsValid()
             && pindexNew->nHeight > 0
-            && IsXFieldNew(blockConnecting.xfield, &xfieldHistory))
+            && IsXFieldNew(blockConnecting.xfield, &xfieldHistory, static_cast<uint32_t>(pindexNew->nHeight - 1)))
         {
             XFieldChange newChange(blockConnecting.xfield.xfieldValue, pindexNew->nHeight + 1, blockConnecting.GetHash());
             xfieldHistory.Add(blockConnecting.xfield.xfieldType, newChange);
@@ -1407,7 +1407,7 @@ bool CChainState::AcceptBlockHeader(const CBlockHeader& block, CValidationState&
     if(pxfieldHistory
         && block.xfield.IsValid()
         && pindex->nHeight > 0
-        && IsXFieldNew(block.xfield, pxfieldHistory))
+        && IsXFieldNew(block.xfield, pxfieldHistory, static_cast<uint32_t>(pindex->nHeight - 1)))
     {
         XFieldChange newChange(block.xfield.xfieldValue, pindex->nHeight + 1, block.GetHash());
         pxfieldHistory->Add(block.xfield.xfieldType, newChange);
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -303,13 +303,7 @@ DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockI
                         continue;
                     COutPoint prevout = tx.vin[j].prevout;
                     if (ColorIdentifier(prevout, outColorId.type) == outColorId) {
-                        // Skip global state writes when called from VerifyDB's dry-run
-                        // sandbox (level 3): chainActive is not actually disconnected, so
-                        // mutating g_colorid_state here would leave it permanently
-                        // inconsistent with the real chain tip.
-                        if (!fDryRun) {
-                            g_colorid_state->Erase(outColorId);
-                        }
+                        g_colorid_state->Erase(outColorId);
                         break;
                     }
                 }
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -247,7 +247,7 @@ int ApplyTxInUndo(Coin&& undo, CCoinsViewCache& view, const COutPoint& out)
 
 /** Undo the effects of this block (with given index) on the UTXO set represented by coins.
  *  When FAILED is returned, view is left in an indeterminate state. */
-DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockIndex* pindex, CCoinsViewCache& view)
+DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockIndex* pindex, CCoinsViewCache& view, bool fDryRun)
 {
     bool fClean = true;
 
@@ -303,7 +303,13 @@ DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockI
                         continue;
                     COutPoint prevout = tx.vin[j].prevout;
                     if (ColorIdentifier(prevout, outColorId.type) == outColorId) {
-                        g_colorid_state->Erase(outColorId);
+                        // Skip global state writes when called from VerifyDB's dry-run
+                        // sandbox (level 3): chainActive is not actually disconnected, so
+                        // mutating g_colorid_state here would leave it permanently
+                        // inconsistent with the real chain tip.
+                        if (!fDryRun) {
+                            g_colorid_state->Erase(outColorId);
+                        }
                         break;
                     }
                 }
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -718,6 +718,11 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
     }
 
     // Stage new issuances only after all other disk writes have succeeded.
+    // Placing this before WriteUndoDataForBlock would leave colorIds permanently
+    // in the chainstate DB if the undo write failed: ConnectBlock would return
+    // false without state.IsInvalid(), the block would not be marked
+    // BLOCK_FAILED_VALID, and the next ActivateBestChain retry would be rejected
+    // with bad-txns-colorid-already-issued, permanently poisoning that block.
     // CommitToBatch will write them to LevelDB atomically with DB_BEST_BLOCK
     // inside the next view.Flush() → BatchWrite call, preventing a crash
     // between the colorId write and the UTXO flush from permanently poisoning
--- a/src/chainstate.h
+++ b/src/chainstate.h
@@ -136,7 +136,7 @@ public:
     bool AcceptBlock(const std::shared_ptr<const CBlock>& pblock, CValidationState& state, CBlockIndex** ppindex, bool fRequested, const CDiskBlockPos* dbp, bool* fNewBlock, CXFieldHistoryMap* pxfieldHistory = nullptr) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
 
     // Block (dis)connection on a given view:
-    DisconnectResult DisconnectBlock(const CBlock& block, const CBlockIndex* pindex, CCoinsViewCache& view) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
+    DisconnectResult DisconnectBlock(const CBlock& block, const CBlockIndex* pindex, CCoinsViewCache& view, bool fDryRun = false) EXCLUSIVE_LOCKS_REQUIRED(cs_main);
     bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pindex,
                     CCoinsViewCache& view, bool fJustCheck = false);
 
--- a/src/file_io.cpp
+++ b/src/file_io.cpp
@@ -585,12 +585,12 @@ bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, CXFieldHistoryMa
         return error("%s: Deserialize or I/O error - %s at %s", __func__, e.what(), pos.ToString());
     }
 
+    // Use nHeight from the trusted block index (pindex->nHeight) when available.
+    // When nHeight == -1 (caller has no index context), fall back to UINT32_MAX so
+    // Get() returns the latest aggregate key — the safe conservative choice.
+    // Never use block.GetHeight() (coinbase prevout.n): that field is attacker-controlled.
     CValidationState state;
-    // Use -1 (→ UINT32_MAX) as the height sentinel so CheckBlockHeader selects
-    // the latest aggregate public key.  block.GetHeight() reads coinbase prevout.n
-    // which is attacker-controllable; passing it here would allow block data on
-    // disk to influence which signing key is used for verification (PR #411 footgun).
-    if(!CheckBlockHeader(block.GetBlockHeader(), state, pxfieldHistory, -1, true))
+    if(!CheckBlockHeader(block.GetBlockHeader(), state, pxfieldHistory, nHeight, true))
         return error("%s: ReadBlockFromDisk: %s", __func__, FormatStateMessage(state));
 
     return true;
--- a/src/file_io.cpp
+++ b/src/file_io.cpp
@@ -585,12 +585,12 @@ bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, CXFieldHistoryMa
         return error("%s: Deserialize or I/O error - %s at %s", __func__, e.what(), pos.ToString());
     }
 
-    // Use nHeight from the trusted block index (pindex->nHeight) when available.
-    // When nHeight == -1 (caller has no index context), fall back to UINT32_MAX so
-    // Get() returns the latest aggregate key — the safe conservative choice.
-    // Never use block.GetHeight() (coinbase prevout.n): that field is attacker-controlled.
     CValidationState state;
-    if(!CheckBlockHeader(block.GetBlockHeader(), state, pxfieldHistory, nHeight, true))
+    // Use -1 (→ UINT32_MAX) as the height sentinel so CheckBlockHeader selects
+    // the latest aggregate public key.  block.GetHeight() reads coinbase prevout.n
+    // which is attacker-controllable; passing it here would allow block data on
+    // disk to influence which signing key is used for verification (PR #411 footgun).
+    if(!CheckBlockHeader(block.GetBlockHeader(), state, pxfieldHistory, -1, true))
         return error("%s: ReadBlockFromDisk: %s", __func__, FormatStateMessage(state));
 
     return true;
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -47,6 +47,8 @@ static constexpr int64_t ORPHAN_TX_EXPIRE_TIME = 20 * 60;
 static constexpr int64_t ORPHAN_TX_EXPIRE_INTERVAL = 5 * 60;
 /** How long a transaction has to be in the mempool before it can unconditionally be relayed (even when not in mapRelay). */
 static constexpr std::chrono::seconds UNCONDITIONAL_RELAY_DELAY = std::chrono::minutes{2};
+/** Maximum number of entries in mapRelay. Oldest entries are evicted when this is reached. */
+static constexpr size_t MAX_RELAY_MAP_SIZE = 50000;
 /** Headers download timeout expressed in microseconds
  *  Timeout = base + per_header * (expected number of headers) */
 static constexpr int64_t HEADERS_DOWNLOAD_TIMEOUT_BASE = 15 * 60 * 1000000; // 15 minutes
@@ -3766,6 +3768,12 @@ bool PeerLogicValidation::SendMessages(CNode* pto)
                         // Expire old relay messages
                         while (!vRelayExpiration.empty() && vRelayExpiration.front().first < nNow)
                         {
+                            mapRelay.erase(vRelayExpiration.front().second);
+                            vRelayExpiration.pop_front();
+                        }
+
+                        // Enforce size cap: evict oldest entries if at limit
+                        while (mapRelay.size() >= MAX_RELAY_MAP_SIZE && !vRelayExpiration.empty()) {
                             mapRelay.erase(vRelayExpiration.front().second);
                             vRelayExpiration.pop_front();
                         }
--- a/src/test/chainstate_tests.cpp
+++ b/src/test/chainstate_tests.cpp
@@ -25,6 +25,8 @@
 #include <primitives/block.h>
 #include <primitives/transaction.h>
 #include <file_io.h>
+#include <xfieldhistory.h>
+#include <key.h>
 
 #include <boost/test/unit_test.hpp>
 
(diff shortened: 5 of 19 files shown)


azuchi (changes requested):
## Review: PR #413 (J1 / J2 / I2 + 5 companion warnings)

Reviewed the commit `c1cc3af1` against master, treating PR #412 as the underlying base since this PR is stacked on it. Diff against PR #412: 9 files, +67/-23. (The K2 / VerifyDB Level 4 regression I flagged on PR #412 is inherited here but should be resolved in #412 itself — not listed below as a PR #413 blocker.)

---

### Confirmed fixes

**J1 — `IsColoredOutPointWith` OOB (`src/wallet/wallet.cpp:3742`)**

Defense in depth: (1) the function itself now early-returns on `!wtx || outpoint.n >= wtx->tx->vout.size()`, and (2) `SelectCoins` performs the `vout.size() <= outpoint.n` bounds check *before* the `IsColoredOutPointWith` call. The new ordering means an out-of-range preset input now causes the whole coin selection to fail (`return false`) instead of crashing, which is the correct outcome for what is unambiguously a malformed input.

**J2 — bumpfee token destruction (`src/wallet/feebumper.cpp:97-104`)**

The change-detection loop now skips any output whose `scriptPubKey` carries a colorId before evaluating `IsChange`, so the subsequent `poutput->nValue -= nDelta` never lands on a token-denominated output. The error wording was tightened (`"Cannot bump fee: transaction has no TPC change output to absorb the fee delta"`) which actually explains what the user must do.

**`IsXFieldNew` is now parent-height aware (`src/xfieldhistory.h:226-241`, `src/chainstate.cpp:861, 1405`)**

Replaces `history->GetListCopy(X).back()` with `history->Get(X, parentHeight)`, so during reorg the comparison happens against the entry that was active at the parent's height — not whatever happens to be in the tail of the in-memory list. Both call sites guard `pindex->nHeight > 0` before invoking, so `parentHeight = pindex->nHeight - 1` cannot underflow.

**`MAX_RELAY_MAP_SIZE = 50000` (`src/net_processing.cpp:50, 3775-3779`)**

Unbounded growth of `mapRelay` is now capped via oldest-first eviction. The eviction iterates over `vRelayExpiration.front()` which is the earliest insertion (entries are pushed to the back), so the order is correct and the size remains ≤ MAX_RELAY_MAP_SIZE after the subsequent insert.

**`signrawtransactionwithwallet` SCHNORR selection (`src/wallet/rpcwallet.cpp:3184`)**

The `request.params[4].isNull()` check was guarding a `get_str()` call on `params[3]` — clear typo. Now `params[3].isNull()` correctly. (`-DEFAULTED` would be a follow-up nit, but the on-the-wire behaviour is now correct.)

**`FundTransaction` colored-only-change key leak (`src/wallet/wallet.cpp:2603-2614`)**

The old code only called `reservekey.KeepKey()` when `mapChangePosInOut[ColorIdentifier()] != -1` (TPC change present). For a tx with only colored change, the reservekey returned to the pool and the same address could be handed back by `getnewaddress`. Now any non-`-1` entry in `mapChangePosInOut` triggers `KeepKey()`. Correct.

**`importwallet` / `dumpwallet` for `CColorKeyID` / `CColorScriptID` (`src/wallet/rpcdump.cpp:655-664, 847-853`)**

`dumpwallet` now appends `address label=...` lines for every colored address-book entry; `importwallet`'s else-branch (non-key line) detects `CColorKeyID` / `CColorScriptID` destinations and calls `SetAddressBook`. Private keys are still handled via the base `CKeyID` section, so this preserves the labels (and the existence of the address-book entry) without trying to import a colored key directly.

---

### Requested changes

**1. `CheckBlockHeader` nHeight ≤ 0 fallback regressed behavior (`src/validation.cpp:1336`)**

The I2 fix itself is correct — both the temp and non-temp branches now uniformly call `Get(AGGPUBKEY, uHeight)`, eliminating the original asymmetry. However, the conversion changes the meaning of `nHeight = -1`:

```cpp
const uint32_t uHeight = (nHeight > 0) ? static_cast<uint32_t>(nHeight) : 0;
```

Previously, `int nHeight = -1` flowed into the `Get(type, uint32_t)` call as `UINT32_MAX`, and `Get()` interpreted that as "latest entry in the history." After this PR, `nHeight = -1` is mapped to `uHeight = 0`, i.e. the genesis aggpubkey.

This silently changes behavior for two existing call sites that pass `-1`:

- `file_io.cpp:593` (the `ReadBlockFromDisk(CDiskBlockPos, …)` overload with default `nHeight = -1`), reached from `file_io.cpp:317` in `LoadExternalBlockFile` for the orphan-recursive-child path. During reindex of a chain that has rotated, a post-rotation block read by this path will fail `CheckBlockHeader` sanity against the genesis key with `bad-proof`. The old code at least accepted blocks at the latest rotation level via LATEST.
- `AcceptBlockHeader` (`chainstate.cpp:1368-1370`) passes `nBlockHeight = -1` when `pindexPrev` is not in `mapBlockIndex` yet. The comment in the new code expects that the header "will be rejected for missing parent," but the call order is: `CheckBlockHeader` first (line 1370) → `prev-blk-not-found` check second (line 1373-1374). Under the new fallback, a legitimate peer-supplied header signed by the current rotation key now early-returns `bad-proof` instead of reaching the `prev-blk-not-found` DoS=10 path, which is harsher on peers and changes the failure mode during initial sync.

Please change the fallback to preserve the prior LATEST semantics:

```cpp
const uint32_t uHeight = (nHeight > 0) ? static_cast<uint32_t>(nHeight) : UINT32_MAX;
```

This keeps the I2 fix (no temp/non-temp asymmetry) while leaving the `-1` callers' behavior matching what they had before. If GENESIS is actually intended for some reason, please update the comment to call out the behavior change for the `LoadExternalBlockFile` reindex and `AcceptBlockHeader` orphan-header paths.

**2. No regression tests for J1 / J2 / I2**

All three are direct bug fixes for issues that were not caught by the existing suite. Please add functional tests:

- **J1**: `fundrawtransaction` / `walletcreatefundedpsbt` with a preset `inputs` entry whose `vout` is out of range of the referenced tx → should error cleanly, not crash. A unit-level test exercising `IsColoredOutPointWith` directly with an unknown txid or oversized `n` would also suffice.
- **J2**: send a colored token to self, then `bumpfee` the resulting transfer tx. After the bump, the recipient's token balance must equal the original amount (no silent decrement by `nDelta`). The current PR description correctly identifies this as the failure mode; a test that diff's the colored balance pre/post-bump would lock the fix in.
- **I2**: run `submitheader` / `processnewblockheader` against a chain that has had an aggregate-pubkey rotation, then submit a header that's valid at its height but would be rejected against `GetLatest()`. The header must be accepted (or rejected for an unrelated reason like `prev-blk-not-found`), confirming both branches resolve the same way.

---

### Non-blocking observations

- **`FundTransaction` colored vs TPC change address reuse:** the same `scriptChange` (one reservekey) is now correctly kept for any change, but if a tx produces both TPC change and colored change they share the same destination. That's a privacy leak linking the two flows — pre-existing, out of scope for this PR, but worth opening a separate issue for.
- **`MAX_RELAY_MAP_SIZE` eviction logging:** the new eviction loop drops entries silently. A `LogPrint(BCLog::NET, ...)` on first eviction (or a once-per-N counter) would help diagnose if a node ever hits the cap, which is otherwise invisible.
- **`importwallet` else-branch parsing reuses the for-loop pattern from the key branch** (`label=` extraction, `#` comment skip). Duplicating the same parsing logic invites drift — consider extracting a small helper if the dump format gains more keyed fields in the future. Out of scope here.
- **CI not yet run** on `FixSecurityVulnerabilities-IJ`. Once it does, the J1/J2/I2 fixes should at least keep the existing tests green.

---

The seven small fixes are clean. Requesting changes on (1) the `CheckBlockHeader` fallback and (2) the test gap; once those land, this should be ready.

azuchi (changes requested):
## Re-review after `25c95c5e8` (review feedback) + `ac43a536f` (post-rebase fixes)

Reviewed both new commits against the previous CHANGES_REQUESTED. The prior items are addressed cleanly, but `ac43a536f`'s no-TPC-change feebumper fallback introduces three issues that need attention before merge.

---

### Confirmed: prior CHANGES_REQUESTED resolved

**1. `CheckBlockHeader` nHeight fallback (`src/validation.cpp:1337`)** — fixed across the two commits:

```cpp
// nHeight <  0 → UINT32_MAX (Get returns latest rotation entry) — orphan path
// nHeight == 0 → 0          (Get returns genesis key)            — genesis block
// nHeight >  0 → nHeight    (Get returns active entry at height)
const uint32_t uHeight = (nHeight >= 0) ? static_cast<uint32_t>(nHeight) : UINT32_MAX;
```

`25c95c5e8` initially mapped `nHeight > 0` to UINT32_MAX which then misrouted genesis (`nHeight == 0`) to the latest key; `ac43a536f` refined the predicate to `>= 0` so all three buckets resolve correctly. Both `LoadExternalBlockFile`'s orphan-recursive-child path and `AcceptBlockHeader`'s missing-parent path now behave as before the I2 fix.

**2. Regression tests for J1 / J2 / I2** — three tests added:

- **I2**: `chainstate_tests.cpp::check_block_header_orphan_uses_latest_aggpubkey` — uses `CTempXFieldHistory` to install a rotation at height 5, builds an orphan header signed by the rotated key, asserts `CheckBlockHeader(nHeight=-1)` returns `true` and `AcceptBlockHeader` rejects with `prev-blk-not-found` (not `bad-proof`). Exactly the regression we wanted to lock in.
- **J1**: `create_transaction_tests.cpp::IsColoredOutPointWith_bounds_check` — unknown txid, `n == vout.size()`, and `n == vout.size() + 100` all must return `false` without crashing. Good coverage.
- **J2**: `wallet_bumpfee.py::test_bumpfee_preserves_colored_token_balance` — see H3 below.

---

## Requested changes

### H1 — `feebumper.cpp` no-TPC-change fallback underpays the fee

In the new branch added by `ac43a536f` (`src/wallet/feebumper.cpp:112-140`):

```cpp
    if (nOutput == -1) {
        // ... finds newInputOp/newInputValue ...
    }

    // Calculate the expected size of the new transaction.
    int64_t txSize = GetTransactionSize(*(wtx.tx));
    const int64_t maxNewTxSize = CalculateMaximumSignedTxSize(*wtx.tx, wallet);
    // ...
    new_fee = nNewFeeRate.GetFee(maxNewTxSize);
```

Then later in the `else` branch for `nOutput == -1`:

```cpp
        mtx.vin.push_back(CTxIn(newInputOp, ...));    // + ~148 B P2PKH input
        // ...
        if (changeValue > GetDustThreshold(...)) {
            mtx.vout.push_back(changeOut);            // + ~34  B P2PKH output
        }
```

`maxNewTxSize` is the OLD transaction's signed size. `new_fee` is derived from it, and `nDelta = new_fee - old_fee` is the increment the new input has to absorb. But the actual bumped tx is `old_size + ~182 bytes` (one P2PKH input + one P2PKH output if the change survives dust). The effective feerate is `new_fee / (old_size + 182)`, which is strictly below the target `nNewFeeRate`.

Downstream impact: the standard BIP125 replacement rule requires the replacement to pay `≥ old_fee + incrementalRelayFee × new_size`. For tight bumpfees (where `nNewFeeRate ≈ old_feerate + incrementalRelayFee`) the under-sized fee will fail the mempool's incremental relay check and the bumped tx silently won't relay. Even when the replacement is accepted, the feerate is wrong.

Fix: when `nOutput == -1`, compute `maxNewTxSize` against a `CMutableTransaction` that already has the prospective new input and change output appended (or add a constant ~182 byte adjustment for the P2PKH input/output pair before calling `nNewFeeRate.GetFee(...)`). The cleanest approach is to build the candidate `mtx` upfront, then call `CalculateMaximumSignedTxSize(CTransaction(mtx), wallet)` against that.

### H2 — `AvailableCoins(..., nullptr)` ignores user coin-control filters

Same block (`src/wallet/feebumper.cpp:123`):

```cpp
        wallet->AvailableCoins(vAvailableCoins, true, nullptr);
```

The `nullptr` discards the caller-provided `coin_control`. Locked outpoints (`wallet->LockCoin`), minconf, and any other constraints the user (or `rpcwallet.cpp::bumpfee`) set on `coin_control` are bypassed. A locked TPC UTXO can end up selected for the bumped tx against the user's wishes, and any tracking of "do not touch these coins" is silently broken.

Fix: pass `&coin_control` instead of `nullptr`:

```cpp
        wallet->AvailableCoins(vAvailableCoins, true, &coin_control);
```

### H3 — The J2 regression test doesn't exercise the no-TPC-change fallback path

`test_bumpfee_preserves_colored_token_balance` calls `rbf_node.transfertoken(dest_addr, 60)`. `transfertoken` is a thin wrapper that delegates to `sendtoaddress` (`src/wallet/rpcwallet.cpp:4629`), and `sendtoaddress` was modified in `25c95c5e8` to always pre-`Select()` a TPC UTXO via:

```cpp
if (colorId.type != TokenTypes::NONE) {
    // ... AvailableCoins, pick first TPC, coin_control.Select(...), fAllowOtherInputs = true
}
```

The pre-selected TPC input guarantees a TPC change output, so in this test the bumped tx **always** has a TPC change output and `feebumper::CreateTransaction` takes the standard `nOutput != -1` path. The new `if (nOutput == -1)` fallback added by `ac43a536f` — the one with H1/H2 — is never executed in CI.

Fix: extend (or duplicate) the test to construct a colored transfer that genuinely lacks a TPC change output, then bumpfee it. Two viable shapes:

1. Build the transfer with `createrawtransaction` + `signrawtransactionwithwallet` so it bypasses the `sendtoaddress` pre-select, with TPC inputs chosen to cover fee exactly (no change). Then `bumpfee` and assert the result.
2. Mock the pre-selection by temporarily emptying the wallet's TPC UTXOs except one of exact-fee size, then `transfertoken` — the BnB path will produce no change. (Fragile; option 1 is cleaner.)

After the fix, the test should assert both that the colored balance is preserved AND that `gettransaction(bump_txid)['details']` shows the expected added input.

---

## Non-blocking observations

**M1 — `reservekey.KeepKey()` called eagerly (`src/wallet/rpcwallet.cpp:3306-3313`)**

```cpp
if (reservekey.GetReservedKey(pubkey, true)) {
    coin_control.destChange = pubkey.GetID();
    reservekey.KeepKey();   // ← consumed regardless of feebumper outcome
}
```

If `feebumper::CreateTransaction` returns `WALLET_ERROR` afterwards (insufficient TPC, dust, signing failure), the keypool key has already been committed. Each failed bumpfee call leaks one address. Recommend deferring `KeepKey()` until after `res == Result::OK`, or using the standard RAII pattern of returning the key on the failure path.

**M2 — first-found vs largest TPC UTXO (`src/wallet/feebumper.cpp:124-135`)**

The loop `break`s on the first eligible TPC UTXO, even if it's smaller than `nDelta`. A wallet with a dust TPC UTXO at the front and a fat one behind it will hit the "new TPC input insufficient to cover fee delta" error. Sorting `vAvailableCoins` by `nValue` descending — or iterating and tracking the largest — would make this robust for the common case.

**O1 — `sendtoaddress` pre-select block is a layering kludge (`src/wallet/rpcwallet.cpp:498-515`)**

The TPC UTXO pre-selection logic added in `25c95c5e8` lives at the RPC layer. As a result:

- `transfertoken` (which delegates to `sendtoaddress`) is covered.
- Raw-tx workflows (`createrawtransaction` → `sendrawtransaction`) and PSBT flows are not — colored transfers built that way still produce no TPC change.
- Future RPCs that construct colored transfers will silently bypass it.

The right home for this logic is `CWallet::CreateTransaction`, which already has the colored/TPC bookkeeping (`coin_control.m_colorTxType == TRANSFER` + `coin_control.m_colorId.type != NONE`). Moving the pre-select there would also make the H1/H2/H3 surface smaller (the fallback in feebumper becomes a hard error rather than a code path with its own fee-calculation logic). Out of scope for this PR — file a follow-up.

**O2 — commit message inaccuracy on `ac43a536f`**

The message claims "`rpcwallet.cpp (bumpfee)`: ... remove the leftover colorId-based pre-selection block that referenced an undefined variable. sendtoaddress is left untouched." The actual diff removes the pre-selection block from `sendtoaddress`, and no undefined variable is visible in the removed code. The intent ("this block is now redundant because the bumpfee fallback handles it") would be clearer.

Non-actionable; flag during squash.

**O3 — `sendmany` colored-address rejection (`src/wallet/rpcwallet.cpp:948-950`)**

New defensive check that throws `RPC_INVALID_ADDRESS_OR_KEY` for colored destinations. Good fix; would benefit from a one-line test in `wallet_basic.py` or similar.

**O4 — `listunspent` CColorScriptID redeemScript (`src/wallet/rpcwallet.cpp:2890-2895`)**

Now returns `redeemScript` for CP2SH UTXOs as well as P2SH. Consistent with the existing P2SH branch. Good.

**O5 — `signrawtransactionwithwallet` SCHNORR selection (`src/wallet/rpcwallet.cpp:3184`)**

The typo fix (`params[4].isNull()` → `params[3].isNull()`) from the earlier round is unchanged and correct.

---

### Summary

Prior CHANGES_REQUESTED items 1 and 2 are resolved. Three new issues in the `ac43a536f` feebumper fallback need to land before merge: fee underpayment (H1), `coin_control` bypass (H2), and the regression test gap that masks both of them (H3). M1/M2/O1–O5 are non-blocking — please address H-series, then I'll re-review for approval.

azuchi (approved):
LGTM — approving.

All three prior blockers (H1/H2/H3) and the two medium nits (M1/M2) plus the sendmany test (O3) are addressed cleanly in `219417a03`. The branch-history cleanup (`373867e8f` collapsing the prior `25c95c5e8` + `ac43a536f`) also drops the `sendtoaddress` pre-select TPC-UTXO kludge entirely (was O1), which is a better outcome than the "move it into `CreateTransaction`" workaround I had suggested — `CreateTransaction` already handles TPC fee selection via `mapValue[ColorIdentifier()] = 0` + `targetValue += nFeeRet` (`src/wallet/wallet.cpp:2638, 2841`), so the kludge was unnecessary.

### H1 — fee calculation now accounts for added input/output

`src/wallet/feebumper.cpp:145-152`:

```cpp
bool rbf = coin_control.m_signal_bip125_rbf.get_value_or(wallet->m_signal_rbf);
mtx = CMutableTransaction{*wtx.tx};
if (nOutput == -1) {
    mtx.vin.push_back(CTxIn(newInputOp, CScript(), rbf ? MAX_BIP125_RBF_SEQUENCE : 0xfffffffe));
    mtx.vout.push_back(CTxOut(newInputValue, GetScriptForDestination(coin_control.destChange)));
}

const int64_t maxNewTxSize = CalculateMaximumSignedTxSize(CTransaction(mtx), wallet);
```

The candidate `mtx` carries the prospective new input and a placeholder change output before `CalculateMaximumSignedTxSize` is called, so the fee estimate reflects the actual bumped tx size. The post-computation dust-handling (`mtx.vout.pop_back()` when the change is dust, with the dust value added to `new_fee`) means the dust branch slightly over-pays — which is fine, vs. the previous under-payment that could fail BIP125's incremental-relay rule.

### H2 — `AvailableCoins` now honors `coin_control`

`src/wallet/feebumper.cpp:123`: now `wallet->AvailableCoins(vAvailableCoins, true, &coin_control);`. Locked UTXOs (`wallet->LockCoin`) and any user-supplied exclusions are respected.

### H3 — `test_bumpfee_no_tpc_change` exercises the fallback path

The new test in `test/functional/wallet_bumpfee.py:287-352` deliberately constructs an unbumped colored transfer with no TPC change output:

1. Creates a small TPC UTXO of exactly `0.00005 TPC`.
2. Builds a raw colored transfer using `createrawtransaction` with the token input + the small TPC input, outputs only the colored recipient (no TPC change) — the small TPC input is consumed entirely as fee.
3. Signs and sends via `sendrawtransaction` so it bypasses `sendtoaddress` and lands in the mempool exactly as constructed.
4. Captures `orig_vin_count`, calls `bumpfee(orig_txid)`, and asserts both `bumped['errors'] == []` and `len(bumped_decoded['vin']) > orig_vin_count` — the `>` confirms the `nOutput == -1` fallback in `feebumper.cpp` was actually entered and added a fresh TPC input.

The later token-balance assertion (`balance == 100`) covers the integrity guarantee. This is precisely the regression test gap the previous review flagged.

### M1 — `CReserveKey` lifetime is now RAII-safe

`src/wallet/rpcwallet.cpp:3298, 3349`: `CReserveKey bumpfeeReserveKey(pwallet)` lives at function scope, and `KeepKey()` is deferred until after `feebumper::CreateTransaction` returns `Result::OK`. On any throw or early return, `~CReserveKey()` runs `ReturnKey()`, returning the keypool index to the pool. The `KeepKey()` is also safe to call when `GetReservedKey()` was never invoked (the inner `nIndex != -1` guard in `src/wallet/wallet.cpp:3765` handles that path).

### M2 — largest TPC UTXO is now selected

`src/wallet/feebumper.cpp:133-138`: the loop now tracks the largest eligible TPC UTXO instead of breaking on the first. A wallet with dust TPC at the front and a fat UTXO behind it now succeeds at bumpfee.

### O3 — `sendmany` colored-address rejection has a test

`test/functional/wallet_coloredcoin.py:660-668` covers the rejection path, including a CLI-mode JSON-string fallback for `--usecli` runs.

### O1 — `sendtoaddress` pre-select kludge removed entirely

The rebase into `373867e8f` simply drops the block. `CreateTransaction` reliably picks TPC inputs for the fee via `mapValue[NONE] = 0` + `targetValue += nFeeRet` in its existing loop — no special-case at the RPC layer is needed.

### Test stability (`c50263533`)

Minor follow-up: `issuetoken` now picks the largest TPC UTXO as the colorId seed (`max(..., key=lambda u: u['amount'])`), so the test's later TPC-input availability for fees is deterministic. Pure test stability.

---

Ready to merge once CI is green.

PR #414, opened 2026-05-15
Add CP2SH colored coin softfork manager and activation gating

Naviabheeman:
Introduces CSoftForkManager with HeightActivation functor to manage per-network softfork activation without scattering #ifdefs through validation logic.

Key design decisions:
- CSoftFork stores a HeightActivation functor (operator()) as the canonical activation predicate; no raw lambdas at call sites.
- Manager stores only networks with deferred activation (Chaintope testnet 1939510133 at block 261594). All other networks activate CP2SH colored from genesis — IsActive returns true when no entry is found.
- VerifyScript no longer hardcodes (flags & SCRIPT_VERIFY_CP2SH_COLORED); it calls FederationParams().SoftForkManager().IsEnforced() so all softfork-dependent code goes through the manager API.
- CI builds inject -DCP2SH_ACTIVATION_TEST_HEIGHT=120 via CMake so feature_cp2sh_softfork.py can exercise the pre/post-activation transition at a low block height. GUIX/release builds never define this macro; production testnet activates at 261594.

New files:
  src/softforkmanager.h          — HeightActivation, CSoftFork, CSoftForkManager
  test/functional/feature_cp2sh_softfork.py — activation transition test

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -549,6 +549,9 @@ jobs:
           else
             # Native builds: enable tests and benchmarks
             CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DENABLE_BENCH=ON")
+            # Inject the CP2SH softfork activation height for functional tests.
+            # Not included in cross-compiled (release) builds above.
+            CMAKE_OPTS+=("-DCP2SH_ACTIVATION_TEST_HEIGHT=120")
           fi
 
           # Use depends-generated CMake toolchain file
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -548,7 +548,7 @@ jobs:
             CMAKE_OPTS+=("-DENABLE_TESTS=OFF" "-DENABLE_BENCH=OFF" "-DBUILD_UTILS=ON")
           else
             # Native builds: enable tests and benchmarks
-            CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DENABLE_BENCH=ON")
+            CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DENABLE_BENCH=ON" "-DCP2SH_ACTIVATION_TEST_HEIGHT=120")
           fi
 
           # Use depends-generated CMake toolchain file
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -549,9 +549,6 @@ jobs:
           else
             # Native builds: enable tests and benchmarks
             CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DENABLE_BENCH=ON")
-            # Inject the CP2SH softfork activation height for functional tests.
-            # Not included in cross-compiled (release) builds above.
-            CMAKE_OPTS+=("-DCP2SH_ACTIVATION_TEST_HEIGHT=120")
           fi
 
           # Use depends-generated CMake toolchain file
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -200,7 +200,7 @@ jobs:
             CMAKE_OPTS+=("-DENABLE_TESTS=OFF")  # Skip tests for cross-compilation
           else
             # Native builds (Linux/macOS)
-            CMAKE_OPTS+=("-DENABLE_TESTS=ON")
+            CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DCP2SH_ACTIVATION_TEST_HEIGHT=120")
 
             # Disable BDB version warning for macOS if using newer version
             if [ "${{ matrix.config.platform }}" == "macos" ]; then
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -551,15 +551,12 @@ target_compile_definitions(core_interface INTERFACE ${DEPENDS_COMPILE_DEFINITION
 target_compile_definitions(core_interface_relwithdebinfo INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_RELWITHDEBINFO})
 target_compile_definitions(core_interface_debug INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_DEBUG})
 
-# Set the CP2SH colored softfork activation height for test builds so that
-# functional tests can exercise the pre/post-activation transition without
-# mining hundreds of thousands of blocks.  Defaults to 120 whenever tests are
-# enabled (native builds on all platforms).  Can be overridden via
-# -DCP2SH_ACTIVATION_TEST_HEIGHT=<n>.  Not set when ENABLE_TESTS is OFF (e.g.
-# cross-compiled release builds).
-if(ENABLE_TESTS AND NOT DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)
-  set(CP2SH_ACTIVATION_TEST_HEIGHT 120)
-endif()
+# CI-only: override the CP2SH colored softfork activation height so functional
+# tests can exercise the pre/post-activation transition without mining hundreds
+# of thousands of blocks.  Must be passed explicitly (e.g. by CI via
+# -DCP2SH_ACTIVATION_TEST_HEIGHT=120).  Never set automatically — doing so
+# would activate the softfork at block 120 on all networks including testnet,
+# breaking IBD of any historical chain that has CP2SH colored outputs.
 if(DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)
   target_compile_definitions(core_interface INTERFACE
     CP2SH_ACTIVATION_TEST_HEIGHT=${CP2SH_ACTIVATION_TEST_HEIGHT}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -551,6 +551,15 @@ target_compile_definitions(core_interface INTERFACE ${DEPENDS_COMPILE_DEFINITION
 target_compile_definitions(core_interface_relwithdebinfo INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_RELWITHDEBINFO})
 target_compile_definitions(core_interface_debug INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_DEBUG})
 
+# CI-only: override the CP2SH colored softfork activation height so functional
+# tests can exercise the pre/post-activation transition without mining hundreds
+# of thousands of blocks.  Not set in GUIX/release builds.
+if(DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)
+  target_compile_definitions(core_interface INTERFACE
+    CP2SH_ACTIVATION_TEST_HEIGHT=${CP2SH_ACTIVATION_TEST_HEIGHT}
+  )
+endif()
+
 # If the {CXX,LD}FLAGS environment variables are defined during building depends
 # and configuring this build system, their content might be duplicated.
 if(DEFINED ENV{CXXFLAGS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -551,9 +551,15 @@ target_compile_definitions(core_interface INTERFACE ${DEPENDS_COMPILE_DEFINITION
 target_compile_definitions(core_interface_relwithdebinfo INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_RELWITHDEBINFO})
 target_compile_definitions(core_interface_debug INTERFACE ${DEPENDS_COMPILE_DEFINITIONS_DEBUG})
 
-# CI-only: override the CP2SH colored softfork activation height so functional
-# tests can exercise the pre/post-activation transition without mining hundreds
-# of thousands of blocks.  Not set in GUIX/release builds.
+# Set the CP2SH colored softfork activation height for test builds so that
+# functional tests can exercise the pre/post-activation transition without
+# mining hundreds of thousands of blocks.  Defaults to 120 whenever tests are
+# enabled (native builds on all platforms).  Can be overridden via
+# -DCP2SH_ACTIVATION_TEST_HEIGHT=<n>.  Not set when ENABLE_TESTS is OFF (e.g.
+# cross-compiled release builds).
+if(ENABLE_TESTS AND NOT DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)
+  set(CP2SH_ACTIVATION_TEST_HEIGHT 120)
+endif()
 if(DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)
   target_compile_definitions(core_interface INTERFACE
     CP2SH_ACTIVATION_TEST_HEIGHT=${CP2SH_ACTIVATION_TEST_HEIGHT}
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -303,7 +303,13 @@ DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockI
                         continue;
                     COutPoint prevout = tx.vin[j].prevout;
                     if (ColorIdentifier(prevout, outColorId.type) == outColorId) {
-                        g_colorid_state->Erase(outColorId);
+                        // Skip global state writes when called from VerifyDB's dry-run
+                        // sandbox (level 3): chainActive is not actually disconnected, so
+                        // mutating g_colorid_state here would leave it permanently
+                        // inconsistent with the real chain tip.
+                        if (!fDryRun) {
+                            g_colorid_state->Erase(outColorId);
+                        }
                         break;
                     }
                 }
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -303,13 +303,7 @@ DisconnectResult CChainState::DisconnectBlock(const CBlock& block, const CBlockI
                         continue;
                     COutPoint prevout = tx.vin[j].prevout;
                     if (ColorIdentifier(prevout, outColorId.type) == outColorId) {
-                        // Skip global state writes when called from VerifyDB's dry-run
-                        // sandbox (level 3): chainActive is not actually disconnected, so
-                        // mutating g_colorid_state here would leave it permanently
-                        // inconsistent with the real chain tip.
-                        if (!fDryRun) {
-                            g_colorid_state->Erase(outColorId);
-                        }
+                        g_colorid_state->Erase(outColorId);
                         break;
                     }
                 }
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -661,20 +661,20 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
         {
             std::vector<CScriptCheck> vChecks;
             bool fCacheResults = fJustCheck; /* Don't cache results if we're actually connecting blocks (still consult the cache, though) */
-            if (!CheckInputs(tx, state, view, fScriptChecks, SCRIPT_VERIFY_NONE, fCacheResults, fCacheResults, txdata[i], nScriptCheckThreads ? &vChecks : nullptr))
+            if (!CheckInputs(tx, state, view, fScriptChecks, GetBlockScriptFlags(pindex), fCacheResults, fCacheResults, txdata[i], nScriptCheckThreads ? &vChecks : nullptr))
                 return error("ConnectBlock(): CheckInputs on %s failed with %s",
                     tx.GetHashMalFix().ToString(), FormatStateMessage(state));
             control.Add(std::move(vChecks));
         }
 
         //if there are colored coins in the output verify their colorids
-        if(!CheckColorIdentifierValidity(tx, state, view))
+        if(!CheckColorIdentifierValidity(tx, state, view, pindex->nHeight))
             return false;
 
         //verify token balances (coinbase has no real inputs so balance check does not apply)
         if (!tx.IsCoinBase()) {
             std::set<ColorIdentifier> newIssuances;
-            if (!VerifyTokenBalances(tx, state, view, txfee, !fJustCheck ? &newIssuances : nullptr))
+            if (!VerifyTokenBalances(tx, state, view, txfee, !fJustCheck ? &newIssuances : nullptr, pindex->nHeight))
                 return false;
             allNewIssuances.insert(newIssuances.begin(), newIssuances.end());
         }
--- a/src/federationparams.cpp
+++ b/src/federationparams.cpp
@@ -124,8 +124,7 @@ std::unique_ptr<CFederationParams> CreateFederationParams(const TAPYRUS_OP_MODE 
     // blocks.  The override applies to whichever network the test node is running.
     //
     // Production builds register only the Chaintope testnet (1939510133) entry,
-    // deferring activation to block 693367 (last legacy CP2SH colored spend on
-    // testnet was at block 693366; all earlier CP2SH outputs predate this rule).
+    // deferring activation to TESTNET_CP2SH_ACTIVATION_HEIGHT
     // All other networks have no entry and therefore activate CP2SH at genesis
     // (CSoftForkManager::IsActive returns true when no entry is found).
 #ifdef CP2SH_ACTIVATION_TEST_HEIGHT
@@ -137,7 +136,7 @@ std::unique_ptr<CFederationParams> CreateFederationParams(const TAPYRUS_OP_MODE 
     if (networkId == 1939510133u) {
         params->RegisterSoftFork(CSoftFork(
             networkId, SCRIPT_VERIFY_CP2SH_COLORED,
-            HeightActivation(693367)
+            HeightActivation(TESTNET_CP2SH_ACTIVATION_HEIGHT)
         ));
     }
 #endif
--- a/src/federationparams.cpp
+++ b/src/federationparams.cpp
@@ -154,6 +154,11 @@ const CSoftForkManager& GetSoftForkManager()
     return FederationParams().SoftForkManager();
 }
 
+bool CSoftForkManager::IsActive(unsigned int flag, int32_t blockHeight) const
+{
+    return IsActive(FederationParams().NetworkId(), flag, blockHeight);
+}
+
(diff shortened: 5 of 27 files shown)


azuchi (changes requested):
## Review: PR #414 (CP2SH softfork manager and activation gating)

Reviewed against `origin/master` (which already contains the squash-merged PR #411). Net diff: 10 files, +386/-2. The intent — a per-network softfork registry with a `HeightActivation` predicate — is reasonable. The implementation, however, has two blockers and one scope question.

---

### Blocking: activation flag is never plumbed into the verification path

The design assumes the `SCRIPT_VERIFY_CP2SH_COLORED` bit flows `GetBlockScriptFlags(pindex)` → `CheckInputs(..., flags, ...)` → `CScriptCheck` → `EvalScript` / `VerifyScript`, and that `VerifyScript` then asks the manager `IsEnforced(SCRIPT_VERIFY_CP2SH_COLORED, flags)` to decide whether to evaluate the redeem script. The manager hook is wired up correctly, but **no caller of `GetBlockScriptFlags(pindex)` exists in the tree.**

- `src/chainstate.cpp:671` (`ConnectBlock`) still passes `SCRIPT_VERIFY_NONE` to `CheckInputs`:
  ```cpp
  if (!CheckInputs(tx, state, view, fScriptChecks, SCRIPT_VERIFY_NONE, fCacheResults, fCacheResults, txdata[i], ...))
  ```
- `src/validation.cpp:959` (`AcceptToMemoryPool`) passes `STANDARD_SCRIPT_VERIFY_FLAGS`, which does not include `SCRIPT_VERIFY_CP2SH_COLORED`.

So the bit never reaches `flags`, and `IsEnforced` in `src/softforkmanager.h:109-114` degenerates:

```cpp
bool IsEnforced(unsigned int flag, unsigned int scriptFlags) const {
    for (const auto& sf : m_softforks)
        if (sf.scriptVerifyFlag & flag)
            return sf.IsEnforced(scriptFlags);  // (scriptFlags & flag) != 0 → false
    return (scriptFlags & flag) != 0;            // also false
}
```

Both branches return false, so the added clause in `VerifyScript`

```cpp
if (scriptPubKey.IsPayToScriptHash() ||
    (scriptPubKey.IsColoredPayToScriptHash() &&
     FederationParams().SoftForkManager().IsEnforced(SCRIPT_VERIFY_CP2SH_COLORED, flags)))
```

never takes the colored branch. CP2SH redeem-script evaluation is **disabled on every network at every height**, which silently reintroduces the original P0 — anyone holding redeem-script bytes can spend a CP2SH output via hash match alone, the same vulnerability PR #411 closed.

**Evidence:** the `colored_coin_cp2sh_redeem_scripts` unit-test suite from PR #411 is failing across every native CI build on this PR with messages like `OK where UNSATISFIED_LOCKTIME expected` / `OK where OP_COLOR_UNEXPECTED expected` / `OK where EVAL_FALSE expected`. Functional `feature_cp2sh_softfork.py` doesn't get to run because the unit tests fail first, but its post-activation assertion (`expect_reject="mandatory-script-verify-flag-failed"`) would also fail for the same reason.

**Fix:** compute `unsigned int flags = GetBlockScriptFlags(pindex)` in `ConnectBlock` and pass it to `CheckInputs`. In `AcceptToMemoryPool`, OR the bit in alongside `STANDARD_SCRIPT_VERIFY_FLAGS` (or take it from the active tip via `GetBlockScriptFlags(chainActive.Tip())`). After that change please confirm `colored_coin_cp2sh_redeem_scripts` and `feature_cp2sh_softfork.py` both pass.

---

### Blocking: new circular dependency `federationparams → script/interpreter → federationparams`

`lint-circular-dependencies` fails for this exact cycle. `src/script/interpreter.cpp` now `#include <federationparams.h>` for the `FederationParams().SoftForkManager()` call, and `src/federationparams.cpp` already pulls in `<script/interpreter.h>` for the `SCRIPT_VERIFY_CP2SH_COLORED` enum value at the register site.

A couple of clean ways out:

1. Move `SCRIPT_VERIFY_*` (or at least the new `SCRIPT_VERIFY_CP2SH_COLORED` enum and a small companion) into a leaf header that has no other dependencies. Both `federationparams` and `script/interpreter` can include the leaf without crossing each other.
2. **Better:** once the first blocker is fixed and the bit is properly plumbed in `flags`, `VerifyScript` no longer needs to call the manager at all — `(flags & SCRIPT_VERIFY_CP2SH_COLORED) != 0` is the same check. Drop the `#include <federationparams.h>` from `script/interpreter.cpp`, and the cycle dissolves naturally. The manager API stays inside consensus code (`GetBlockScriptFlags`) where it belongs.

---

### Scope: legacy custom-OP_COLOR rejection in `CheckColorIdentifierValidity` is still unconditional

This came up in the prior conversation about PR #406 — the chain-sync concern that an upgraded node cannot replay history on a network that has ever produced a custom-OP_COLOR transaction. This PR only gates the CP2SH redeem-script evaluation in `VerifyScript`. The three `bad-txns-nonstandard-opcolor` rejections at `validation.cpp:469`, `:493`, and `:567` (inside `CheckColorIdentifierValidity` / `VerifyTokenBalances`) remain unconditional and will still fail blocks at any height regardless of the softfork status.

This means that even with the gating wired up correctly, an upgraded node on Chaintope testnet (`networkId == 1939510133`) that has even one legacy custom-OP_COLOR transaction in its pre-261594 history will still fail to sync from scratch — the activation height only affects the redeem-script evaluation, not the broader CP2PKH/CP2SH-only consensus rule.

Please clarify the design intent in the PR description:

- If a follow-up PR will extend the manager to also gate the `bad-txns-nonstandard-opcolor` checks, note that here and link the issue.
- If the chain-sync problem is being solved out-of-band (e.g. "we verified the testnet history contains no custom-OP_COLOR transactions"), the PR description should record that audit.

Without one of those, this PR doesn't actually resolve the issue it appears to be motivated by, and the activation height for 1939510133 is misleading: it implies the network can keep its legacy state through 261594, but the legacy state would have already been blocked at sync time.

---

### Non-blocking observations

- **`CSoftForkManager::IsEnforced` is a tautology.** Both branches degenerate to `(scriptFlags & flag) != 0`. The whole function can collapse to that single expression. Combined with the recommended fix to remove `manager.IsEnforced` from `VerifyScript`, it likely just disappears.
- **`IsEnforced` doesn't take `networkId`.** If two networks ever register the same `scriptVerifyFlag` the first hit in the vector wins. Not exploitable today (only one entry), but the API shape doesn't match `IsActive` / `GetScriptFlags`, which both take a networkId. Worth aligning before more callers depend on it.
- **Production-height path untested.** `feature_cp2sh_softfork.py` only exercises the CI-injected 120. The `#else` branch (`networkId == 1939510133u`, `HeightActivation(261594)`) has no automated coverage. A small unit test asserting `SoftForkManager().IsActive(1939510133u, SCRIPT_VERIFY_CP2SH_COLORED, 261593) == false` and `... 261594) == true` would lock this in.
- **`CMakeLists.txt` injection lives next to depends config.** The `if(DEFINED CP2SH_ACTIVATION_TEST_HEIGHT)` block is fine; placing it nearer the script-flag definition (or under a `cmake/softforks.cmake` once there is a second one) might age better, but that's a future-proofing nit.

---

The scaffolding is in the right shape — once the `flags` plumbing is hooked up and the cycle is broken (these tend to be the same fix), the design is sound. Requesting changes on (1) the activation gating having no effect, (2) the circular dependency, and (3) clarifying scope vs. the legacy-UTXO chain-sync problem.

azuchi (changes requested):
## Re-review after `fb4fe9333` (review feedback)

The three prior blockers are addressed cleanly. The API encapsulation is a real improvement. One new boundary bug introduced by the mempool wiring needs to land before merge.

---

### Confirmed: prior CHANGES_REQUESTED resolved

**1. `SCRIPT_VERIFY_CP2SH_COLORED` plumbing into the verification path**

`ConnectBlock` now passes the real flags computed from the block's height:

```cpp
// src/chainstate.cpp:664
if (!CheckInputs(tx, state, view, fScriptChecks, GetBlockScriptFlags(pindex), ...))
```

The interpreter check at `src/script/interpreter.cpp:1655` is now `GetSoftForkManager().IsEnabled(SCRIPT_VERIFY_CP2SH_COLORED, flags)` which collapses to the trivial `(flags & SCRIPT_VERIFY_CP2SH_COLORED) != 0` — exactly what we wanted. CP2SH redeem-script evaluation now runs at and after the activation height on every registered network, and the previously-failing `colored_coin_cp2sh_redeem_scripts` suite should pass once CI runs.

**2. Circular dependency `federationparams → script/interpreter → federationparams` broken**

The solution is clean:

- `softforkmanager.h` only declares `const CSoftForkManager& GetSoftForkManager();` (no `federationparams.h` include)
- The implementation lives in `federationparams.cpp:152`, which is free to include `script/interpreter.h`
- `script/interpreter.cpp` now only `#include <softforkmanager.h>`, never `<federationparams.h>`

I ran `test/lint/lint-circular-dependencies.sh` locally against `fb4fe9333` and it passes.

**3. Scope: legacy custom-OP_COLOR rejection now properly gated**

`CheckColorIdentifierValidity` and `VerifyTokenBalances` both gained an `int32_t blockHeight = INT32_MAX` parameter. The three `bad-txns-nonstandard-opcolor` checks in `src/validation.cpp` (output-side at `:457`, input-side at `:479`, balance-side at `:553`) all now wrap the rejection in `GetSoftForkManager().IsActive(SCRIPT_VERIFY_CP2SH_COLORED, blockHeight)`. Pre-activation: legacy custom-OP_COLOR scripts are tolerated, which lets an upgraded node sync from genesis on testnet through the pre-261594 history. Post-activation: the strict CP2PKH/CP2SH-only rule applies. The `INT32_MAX` default keeps every legacy caller (the EXCLUSIVE_LOCKS_REQUIRED annotation removal also opens up the mempool path) on the strict path.

This is exactly the semantic the prior review was asking for.

**API encapsulation improvements (drive-by, but worth calling out)**

- `CSoftFork::IsEnforced` and `CSoftForkManager::IsEnforced` deleted — the tautology is gone
- `CSoftForkManager::IsActive(flag, blockHeight)` overload added (resolves the network via `FederationParams()` internally) — matches the prior nit about asymmetry with `IsActive(networkId, ...)`
- `IsEnabled(flag, scriptFlags)` is the single source of truth for the interpreter and is a trivial bitmask check, no hidden manager state
- The hand-rolled CP2PKH/CP2SH structure tests in the old `CheckColorIdentifierValidity` are replaced with calls to `IsColoredPayToPubkeyHash` / `IsColoredPayToScriptHash`, removing the duplicate definition

---

## Requested changes

### N1 — Mempool script-flag uses `tip` height, not `tip + 1`, causing an activation-boundary inconsistency

In `src/validation.cpp:943-957`:

```cpp
        // CheckColorIdentifierValidity earlier:
        if (!CheckColorIdentifierValidity(tx, state, view, chainActive.Tip()->nHeight + 1))   // tip+1 ✓
            return false;
        ...
        // The new block here:
        const unsigned int mempoolScriptFlags = STANDARD_SCRIPT_VERIFY_FLAGS |
                                                GetBlockScriptFlags(chainActive.Tip());        // tip ✗
        if (!CheckInputs(tx, state, view, true, mempoolScriptFlags, true, false, txdata)) ...

        const unsigned int tipScriptFlags = GetBlockScriptFlags(chainActive.Tip());           // tip ✗
        if (!CheckInputsFromMempoolAndCache(opt.context, tx, opt.state, view, pool, tipScriptFlags, true, txdata)) ...

        if (!VerifyTokenBalances(tx, opt.state, view, ::minRelayTxFee.GetFee(nSize), nullptr, chainActive.Tip()->nHeight + 1))   // tip+1 ✓
            return false;
```

`GetBlockScriptFlags(pindex)` queries `IsActive(SCRIPT_VERIFY_CP2SH_COLORED, pindex->nHeight)`. When `chainActive.Tip()->nHeight == activationHeight - 1`, the flag is **not** set, but `CheckColorIdentifierValidity` and `VerifyTokenBalances` correctly query at `tip+1 == activationHeight` and **do** apply the post-activation rules. The mempool can then admit a tx where:

1. The colored outputs pass the strict CP2PKH/CP2SH-only check (post-activation rules)
2. The CP2SH redeem script is **not** evaluated, because `mempoolScriptFlags` lacks `SCRIPT_VERIFY_CP2SH_COLORED`

When the next block at `activationHeight` is mined and includes this tx, `ConnectBlock` runs `CheckInputs` with the correct flags (`GetBlockScriptFlags(pindex)` where `pindex->nHeight == activationHeight`), evaluates the redeem script, and rejects the block. The result is a single-block window where the mempool produces invalid blocks at activation.

This window is exactly one block wide, but for the Chaintope testnet activation at height 261594 (and the CI-injected height 120) it is real and reproducible — and `feature_cp2sh_softfork.py` does not exercise it. The test crosses the boundary at the block level, never tests mempool admission at `tip == activationHeight - 1` followed by the activation block.

**Fix.** Compute the flags for the next block height, not the tip:

```cpp
// preferred: add an overload that takes a height
unsigned int GetBlockScriptFlagsForHeight(int32_t blockHeight);
// or inline:
unsigned int nextFlags = SCRIPT_VERIFY_NONE;
if (GetSoftForkManager().IsActive(SCRIPT_VERIFY_CP2SH_COLORED, chainActive.Tip()->nHeight + 1))
    nextFlags |= SCRIPT_VERIFY_CP2SH_COLORED;
const unsigned int mempoolScriptFlags = STANDARD_SCRIPT_VERIFY_FLAGS | nextFlags;
```

Apply the same change to `tipScriptFlags`.

**Regression test.** Add a boundary test to `feature_cp2sh_softfork.py` that:

1. Builds a CP2SH tx with a bad redeemScript (OP_0)
2. Mines blocks until `getblockcount() == ACTIVATION_HEIGHT - 1`
3. Submits the tx via `sendrawtransaction` — it must be **rejected** (not just at the next block, but at mempool admission). With the fix above, mempool flags will already include `SCRIPT_VERIFY_CP2SH_COLORED` and the tx is rejected with `mandatory-script-verify-flag-failed`. Without the fix, the tx is admitted and only fails when the activation block tries to include it.

---

## Non-blocking observations

**N2 — Loss of distinct `bad-txns-invalid-colorid` reject code (`src/validation.cpp:451-461`)**

The original code distinguished two cases when `outColorId.type == TokenTypes::NONE`:

- Structurally CP2PKH/CP2SH but invalid colorId type byte → `bad-txns-invalid-colorid`
- Anything else with OP_COLOR → `bad-txns-nonstandard-opcolor`

The new code rejects both as `bad-txns-nonstandard-opcolor` post-activation. The new comment ("`IsColoredPayToPubkeyHash` / `IsColoredPayToScriptHash` check both structure and type byte; reaching here means neither matched") is correct, but the lumping makes debugging which case fired harder. A trivial fix is to re-test the structural-but-invalid-type-byte case and emit the distinct code; pure debuggability nit.

**O1 — Commit-message inaccuracy on `fb4fe9333`**

The message says "`Fix colored_coin_cp2sh_redeem_scripts`: use `SCRIPT_VERIFY_CHECKLOCKTIMEVERIFY` for CLTV tests and `SCRIPT_VERIFY_CHECKSEQUENCEVERIFY` for CSV tests". The actual diff only changes the flag from `SCRIPT_VERIFY_NONE` to `SCRIPT_VERIFY_CP2SH_COLORED` — the CLTV/CSV-specific flags are never set. This works in Tapyrus because `OP_CHECKLOCKTIMEVERIFY` and `OP_CHECKSEQUENCEVERIFY` are always enforced (never gated by their `SCRIPT_VERIFY_*` flag in the interpreter switch), so the tests do exercise the CLTV/CSV failure paths and produce `SCRIPT_ERR_UNSATISFIED_LOCKTIME` as expected. The implementation is correct; the message should be updated when squashing.

**O2 — `feature_cp2sh_softfork.py` `expect_reject` change to `block-validation-failed`**

The original `expect_reject="mandatory-script-verify-flag-failed"` would have been the per-script error code; switching to `block-validation-failed` matches the new block-level rejection (the bad-redeem tx is rejected by `ConnectBlock` as part of the block, not by mempool). Acceptable, but with the N1 fix in place this test could also assert mempool admission failure at `tip == activationHeight - 1`, which is the stronger guarantee.

**O3 — Co-authored-by `Claude Sonnet 4.6` in `fb4fe9333`**

Non-actionable observation; the commit metadata reflects AI-assisted authorship. The implementation quality is consistent with a thoughtful human review, and the cleanups (deleting the tautology, encapsulating the network lookup) match the prior review's intent precisely.

---

### Summary

The three prior blockers are resolved. The API surface is cleaner than before. One new issue — the mempool/tip vs next-block-height inconsistency at the activation boundary — needs to land before merge, ideally with a boundary test in `feature_cp2sh_softfork.py` that exercises mempool admission at `tip == activationHeight - 1`. N2/O1/O2/O3 are non-blocking.

azuchi (changes requested):
## Re-review after the latest update (top of branch `00bc440b9`)

The **N1** boundary issue from the previous review is cleanly resolved, the wider design improvements (CP2PKH issuance, activation-aware signing, CMakeLists CI-only test height) are well-scoped, and the two recent deadlock fixes are correct. **One blocker remains**: a parallel signing path was missed when `ProduceSignature` was made activation-aware, so `signrawtransaction*` and `tapyrus-tx -sign` will silently produce invalid signatures for CP2SH inputs post-activation.

---

### Confirmed resolved (prior `CHANGES_REQUESTED`)

**N1 — mempool script flags now query tip+1** (`f2bd2cef2`)

- `src/validation.cpp:631-650` introduces `GetBlockScriptFlags(int32_t blockHeight)` and `AcceptToMemoryPoolWorker` at `:945-957` uses `nextBlockHeight = chainActive.Tip()->nHeight + 1` consistently for both `mempoolScriptFlags` and `tipScriptFlags`. This matches the height already used by `CheckColorIdentifierValidity` and `VerifyTokenBalances`.
- `test/functional/feature_cp2sh_softfork.py:194-223` adds a boundary test that confirms a bad-redeemScript CP2SH UTXO at `ACTIVATION_HEIGHT - 1` and then asserts `sendrawtransaction` is rejected with `mandatory-script-verify-flag-failed` *while the tip is still at that height* — directly exercising the one-block window the prior review identified.

---

### Other improvements in this update (drive-by, all good)

- **Activation-aware signing inside `ProduceSignature`** (`97cfa5a88`, `src/script/sign.cpp:176-218`). Pre-activation: `scriptSig = push(redeemScript)` (Stage 1, matches legacy hash-match semantics). Post-activation: `scriptSig = push(sig) push(pubkey) push(redeemScript)` (Stage 2, verified with `SCRIPT_VERIFY_CP2SH_COLORED`). Default of `STANDARD_SCRIPT_VERIFY_FLAGS` preserves the historical behaviour for callers that don't opt in.
- **`IssueToken` / `IssueReissuableToken` switched to CP2PKH** (`src/wallet/rpcwallet.cpp:4280-4305, 4411-4413`). Removes the only wallet sites that produced CP2SH outputs, aligning issuance with `getnewaddress`/`getrawchangeaddress`/colored-change and eliminating the pre-activation signing ambiguity for wallet-created tokens. `wallet_balance.py` balance assertions correctly adjusted for the 40-sat fee increase (CP2PKH 60B vs CP2SH 58B at 20 sat/B).
- **CMakeLists test-height is CI-only again** (`4adf440bb` reverting the regression in `d7f7011813`). The auto-set would have activated CP2SH at block 120 on **all** networks for any developer build with `ENABLE_TESTS=ON`, breaking IBD of any historical chain (notably testnet 1939510133) that has CP2SH spends above height 120. Good catch and clean revert.
- **`AssertLockHeld(cs_main)` in `CWallet::SignTransaction` / `CreateTransaction`** (`b68a9de5f`). Replaces the inline `{ LOCK(cs_main); ... }` that would have deadlocked when the caller already held `cs_main` (which is the common pattern via `LOCK2(cs_main, pwallet->cs_wallet)`). I verified the contract holds: `feebumper::SignTransaction` takes `LOCK2(cs_main, wallet->cs_wallet)` at `src/wallet/feebumper.cpp:264-266`, every CreateTransaction caller in `src/wallet/rpcwallet.cpp` (sendmany, SendMoney via sendtoaddress, IssueToken, IssueReissuableToken, transfertoken) is under `LOCK2(cs_main, pwallet->cs_wallet)` in its outer function, and the test fixture at `src/wallet/test/test_tapyrus_wallet.cpp:139` was correctly upgraded from `LOCK(wallet->cs_wallet)` to `LOCK2(cs_main, wallet->cs_wallet)`.
- **Lock-order inversion fix in `ProcessGetData`** (`00bc440b9`, `src/net_processing.cpp:1389-1400`). Splits the `cs_inventory` check from the `cs_main` insert, restoring the `cs_main → cs_inventory` ordering used by `SendMessages`. The introduced TOCTOU between releasing `cs_inventory` and acquiring `cs_main` is benign: at most one duplicate inv announcement in a rare race window, because `m_recently_announced_invs.insert` is idempotent.
- **Testnet activation height updated to 693367** (`b68a9de5f`). Commit message records the rationale (block 693366 was the last legacy CP2SH colored spend on testnet); the deferral now covers the actual history rather than the placeholder 261594.

---

## Requested changes

### B1 — Blocking: `signrawtransaction*` and `tapyrus-tx -sign` are not activation-aware

`ProduceSignature` was given a `scriptVerifyFlags` parameter, but the parallel signing path through `src/rpc/rawtransaction.cpp` and `src/tapyrus-tx.cpp` was missed. Both call `ProduceSignature` without the new argument, so they pick up the default `STANDARD_SCRIPT_VERIFY_FLAGS` — which does **not** include `SCRIPT_VERIFY_CP2SH_COLORED`.

For CP2SH inputs post-activation this produces a **Stage 1 signature** (`scriptSig = push(redeemScript)`), which the network rejects with `mandatory-script-verify-flag-failed`.

Worse, the same path does its own self-verification with `STANDARD_SCRIPT_VERIFY_FLAGS`, so it cannot detect that the signature is invalid for the next block. The RPC returns `"complete": true`, the user broadcasts, the broadcast fails.

**Affected call sites:**

- `src/rpc/rawtransaction.cpp:738` — `combinerawtransaction` merging signatures
- `src/rpc/rawtransaction.cpp:854` — the central `SignTransaction()` helper used by `signrawtransactionwithwallet`, `signrawtransactionwithkey`, and the deprecated `signrawtransaction`
- `src/rpc/rawtransaction.cpp:861` — the matching `VerifyScript` self-check that reports `"complete": true`
- `src/tapyrus-tx.cpp:643` — offline `tapyrus-tx -sign`

**Impact assessment.** The CP2PKH switch in `IssueToken`/`IssueReissuableToken` means Tapyrus wallets no longer create new CP2SH outputs natively, so the affected universe is bounded to:
1. Pre-activation CP2SH UTXOs (whatever exists on chain up to block 693367 on testnet)
2. Watch-only or externally-imported CP2SH scripts
3. Multi-party or cold-signing flows that rely on `signrawtransactionwithwallet` / `tapyrus-tx -sign`

For anyone in that set, the post-activation spend path is silently broken. The mitigation "drain CP2SH UTXOs before activation" is plausible operationally but should not be the only defence — the signing path itself should be wired correctly.

**Fix.** Thread the same activation-aware flag computation that `CWallet::SignTransaction`/`CreateTransaction` use (`src/wallet/wallet.cpp:2558-2560, 3081-3083`) into the `SignTransaction()` helper in `rawtransaction.cpp` and into `tapyrus-tx.cpp`. Pattern:

```cpp
unsigned int signVerifyFlags = STANDARD_SCRIPT_VERIFY_FLAGS;
if (GetSoftForkManager().IsActive(SCRIPT_VERIFY_CP2SH_COLORED, chainActive.Height() + 1))
    signVerifyFlags |= SCRIPT_VERIFY_CP2SH_COLORED;
// ... pass to BOTH ProduceSignature() and the subsequent VerifyScript() self-check
```

For `tapyrus-tx -sign` there is no chain context, so the flag should follow the same convention as the network identifier handling — either default to active (post-fork-only build) or take an explicit `-cp2shcolored` flag. The simplest correct default is "active", which produces Stage 2 signatures that are also valid pre-activation (the OP_HASH160 == H_REDEEM check still passes because the redeem-script bytes are present at the bottom of the stack via `PushAll`). Worth confirming this on the unit-test corpus before adopting.

**Regression test.** Add a functional case that:
1. Imports a CP2SH redeem script into the wallet's keystore
2. Mines past the activation height
3. Builds a raw spend of the CP2SH UTXO
4. Calls `signrawtransactionwithwallet` and asserts the resulting tx is accepted by `sendrawtransaction` (not just `"complete": true`)

Without this test, the gap is invisible because `feature_cp2sh_softfork.py` only exercises the validation layer with hand-built scriptSigs.

---

## Non-blocking

### M1 — `federationparams_tests.cpp` is out of sync with production

`src/test/federationparams_tests.cpp:166-185` registers `HeightActivation(261594)` and asserts pre/post activation at 261593/261594. The comment says *"Mirror the production registration for Chaintope testnet (networkId 1939510133)"*, but production was updated to **693367** in `b68a9de5f`. The test still passes (it uses a local `CSoftForkManager` instance) but no longer reflects production — defeating the point of the test.

The least-fragile fix is to expose the height as a named constant (e.g. `CHAINTOPE_TESTNET_CP2SH_ACTIVATION_HEIGHT = 693367`) and reference it from both `src/federationparams.cpp:140` and the test. Failing that, just hard-code 693367 in the test.

### O1 — `AssertLockHeld(cs_main)` placement

`src/wallet/wallet.cpp:2558` and `:3081` put `AssertLockHeld(cs_main)` inside the per-input `for` loop. The assertion only needs to hold once for the function (the lock cannot be dropped mid-iteration), so a single assert at function entry (alongside the existing `cs_wallet` requirement) is cleaner. Pure style; no correctness implication.

### O2 — `ProcessGetData` lock split

Discussed above under "good drive-by fixes". The TOCTOU window is benign because the recorded `m_recently_announced_invs.insert` is idempotent and only affects relay re-announcement, not validity. Worth a short comment in the code noting that the duplicate-insert race is intentional.

---

### Summary

- **B1 blocks merge**: thread the activation-aware flag computation into `src/rpc/rawtransaction.cpp` (lines 738 and 854/861) and `src/tapyrus-tx.cpp:643`, and add a `signrawtransactionwithwallet` regression test for post-activation CP2SH signing.
- **M1** should be fixed in the same push to avoid drift between test and production constants.
- **O1, O2** are non-blocking.
- Everything else is in good shape — the design is coherent and the resolved issues are well-tested.

azuchi (approved):

PR #417, opened 2026-06-01
CI with non std functional test

Naviabheeman:
Move the functional tests in debug mode to extended tests as debug mode CI does not run the tests at all. So some errors like debug lock order were missed in the previous testing of v0.7.2 changes. Non-standard scripts used in the debug mode are changed to other standard scripts.
A new CI runs every week to verify the long running tests like feature_pruning.

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -670,6 +670,14 @@ jobs:
           fi
 
           if [ "${{ matrix.config.build_type }}" == "Debug" ]; then
+            echo "---------------- Start debug mode functional tests ----------------"
+            cd "$BASE_ROOT_DIR"
+            if python3 test/functional/test_runner.py --tmpdir="$TAPYRUS_TEST_DIR" -j 1 --combinedlogslen=4000 --failfast --quiet --debugscripts; then
+              echo "Debug mode functional tests passed"
+            else
+              echo "Debug mode functional tests failed"
+              exit 1
+            fi
             echo "--------------- DAILY DEBUG BUILD FINISHED ---------------------"
             exit 0
           fi
--- a/.github/workflows/heavy-functional-tests.yml
+++ b/.github/workflows/heavy-functional-tests.yml
@@ -1,0 +1,220 @@
+# Copyright (c) 2025 Chaintope Inc.
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+name: Heavy Functional Tests
+
+# Runs tests that are too slow for the daily CI matrix (feature_pruning: 1.5–4 h, 4 GB).
+# Scheduled weekly; can also be triggered manually or after a successful daily build.
+on:
+  schedule:
+    # Sunday 5 AM JST (Saturday 8 PM UTC)
+    - cron: '0 20 * * 0'
+  workflow_dispatch:
+    inputs:
+      test_branch:
+        description: 'Branch to test (default: master)'
+        type: string
+        default: ''
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-heavy
+  cancel-in-progress: true
+
+env:
+  CI_FAILFAST_TEST_LEAVE_DANGLING: 1
+  CTEST_OUTPUT_ON_FAILURE: 'ON'
+  CMAKE_BUILD_PARALLEL_LEVEL: 4
+  LC_ALL: C.UTF-8
+  BOOST_TEST_RANDOM: 1
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: 'true'
+
+jobs:
+  heavy-tests:
+    name: Pruning + USDT tests (Linux x86_64 RelWithDebInfo)
+    runs-on: ubuntu-latest
+    # feature_pruning uses up to 4 GB disk and takes 1.5–4 hours; allow 8 hours total.
+    timeout-minutes: 480
+    env:
+      BASE_ROOT_DIR: ${{ github.workspace }}
+      CCACHE_DIR: ${{ github.workspace }}/ccache
+      BASE_BUILD_DIR: ${{ github.workspace }}/build
+      DEPENDS_DIR: ${{ github.workspace }}/depends
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.test_branch || github.ref || 'master' }}
+          submodules: recursive
+
+      # Share the ccache populated by the daily-test workflow for this exact config.
+      - name: Restore Ccache cache
+        id: ccache-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: daily-linux-x86_64-RelWithDebInfo-true-true-true-ccache-${{ github.run_id }}
+          restore-keys: daily-linux-x86_64-RelWithDebInfo-true-true-true-ccache-
+
+      - name: Restore Depends cache
+        id: depends-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ env.DEPENDS_DIR }}
+          key: depends-v6-qt6-linux-x86_64-${{ hashFiles('depends/packages/*', 'depends/hosts/*.mk', 'depends/builders/*.mk', 'depends/funcs.mk', 'depends/toolchain.cmake.in', 'depends/patches/**') }}
+          restore-keys: depends-v6-qt6-linux-x86_64-
+
+      - name: Setup Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc g++ ccache build-essential cmake ninja-build meson pkgconf \
+            python3-zmq bsdmainutils systemtap-sdt-dev bpfcc-tools bpftrace python3-venv \
+            libboost-test-dev
+          python3 -m venv $HOME/venv
+          source $HOME/venv/bin/activate
+          pip3 install pyzmq
+
+      - name: Build Depends
+        run: |
+          export HOST="x86_64-pc-linux-gnu"
+          DEPENDS_PREFIX="$DEPENDS_DIR/$HOST"
+
+          if [ -d "$DEPENDS_PREFIX" ] && [ -f "$DEPENDS_PREFIX/lib/libevent.a" ] && \
+             [ -f "$DEPENDS_PREFIX/lib/cmake/Qt6/Qt6Config.cmake" ] && \
+             [ -f "$DEPENDS_PREFIX/lib/libdb_cxx-4.8.a" ]; then
+            echo "Depends cache hit — skipping rebuild"
+          else
+            echo "Building depends for $HOST"
+            make -C depends HOST=$HOST -j$(nproc) install_cmake
+          fi
+
+          echo "DEPENDS_PREFIX=$DEPENDS_PREFIX" >> $GITHUB_ENV
+
+      - name: Enable core dumps and set test dir
+        run: |
+          CORE_DUMP_DIR="${{ github.workspace }}/core_dumps"
+          mkdir -p "$CORE_DUMP_DIR"
+          chmod 777 "$CORE_DUMP_DIR"
+          sudo sysctl -w kernel.core_pattern="$CORE_DUMP_DIR/core.%e.%p"
+          ulimit -c unlimited
+          TEST_DIR="${{ github.workspace }}/tapyrus_test"
+          mkdir -p "$TEST_DIR"
+          chmod -R 777 "$TEST_DIR"
+          echo "TAPYRUS_TEST_DIR=$TEST_DIR" >> $GITHUB_ENV
+
+      - name: Configure and build
+        run: |
+          HOST="x86_64-pc-linux-gnu"
+          DEPENDS_PREFIX="$DEPENDS_DIR/$HOST"
+          export CC="gcc"
+          export CXX="g++"
+          export PKG_CONFIG_PATH="$DEPENDS_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH"
+          export LD_LIBRARY_PATH="$DEPENDS_PREFIX/lib:$LD_LIBRARY_PATH"
+          echo ">" $HOME/.tapyrus
+
+          cmake -S "$BASE_ROOT_DIR" -B "$BASE_BUILD_DIR" -GNinja \
+            -DENABLE_ZMQ=ON \
+            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DCMAKE_PREFIX_PATH="$DEPENDS_PREFIX" \
+            -DCMAKE_TOOLCHAIN_FILE="$DEPENDS_PREFIX/toolchain.cmake" \
+            -DQt6_DIR="$DEPENDS_PREFIX/lib/cmake/Qt6" \
+            -DBUILD_GUI=ON \
+            -DENABLE_TRACING=ON \
+            -DENABLE_WALLET=ON \
+            -DWITH_BDB=ON \
+            -DENABLE_TESTS=ON \
+            -DENABLE_BENCH=OFF \
+            -DBOOST_ROOT="$DEPENDS_PREFIX" \
+            -DBOOST_INCLUDEDIR="$DEPENDS_PREFIX/include" \
+            -DBOOST_LIBRARYDIR="$DEPENDS_PREFIX/lib" \
+            -DZEROMQ_ROOT="$DEPENDS_PREFIX" \
+            -DZEROMQ_INCLUDE_DIR="$DEPENDS_PREFIX/include" \
+            -DZEROMQ_LIBRARY="$DEPENDS_PREFIX/lib/libzmq.a" \
+            -DBerkeleyDB_ROOT="$DEPENDS_PREFIX" \
+            -DBerkeleyDB_INCLUDE_DIR="$DEPENDS_PREFIX/include" \
+            -DBerkeleyDB_LIBRARY="$DEPENDS_PREFIX/lib/libdb_cxx-4.8.a"
+
+          cmake --build "$BASE_BUILD_DIR" --parallel -j$(nproc) --target all
+
+      - name: Run feature_pruning test
+        run: |
+          source $HOME/venv/bin/activate
+          cd "$BASE_ROOT_DIR"
+          export PATH="$BASE_BUILD_DIR/bin:$PATH"
+          echo "Disk space before pruning test:"
+          df -h .
+          echo "---------------- Start feature_pruning test ----------------"
+          if python3 test/functional/test_runner.py \
+            --tmpdir="$TAPYRUS_TEST_DIR" \
+            -j 1 \
+            --combinedlogslen=4000 \
+            --failfast \
+            feature_pruning.py; then
+            echo "feature_pruning test passed"
+          else
+            echo "feature_pruning test FAILED"
+            exit 1
+          fi
+          echo "Disk space after pruning test:"
+          df -h .
+
+      # TODO: Uncomment this section once USDT tests are ready to run in CI.
+      # The scripts exist in test/functional/ and are listed in USDT_SCRIPTS in test_runner.py.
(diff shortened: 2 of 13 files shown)


azuchi (changes requested):
## Review of PR #417 — CI with non-std functional test

**Verified scope**: PR #417 head `575bd1a56`, base `master`. Branched from `1f0759daa` (PR #409 merge). 3 commits actually touch only 10 CI/test files (+472 / -163), but the branch base is **before PR #411–414 were merged**.

---

## ✅ Strategic intent — strongly endorsed

The PR description correctly identifies a real testing gap: debug-mode functional tests were never executed in CI, which is precisely why the `cs_inventory → cs_main` lock-order bug slipped through pre-merge testing of PR #414 and only surfaced in functional testing post-merge. This PR materially raises CI quality.

Content-level changes look sensible:

- `feature_block.py`: replace bare OP_TRUE outputs with P2SH(OP_TRUE) so tests run under default policy — correct standardization
- `feature_cltv.py`: drop `-acceptnonstdtxn=1`, adapt testmempoolaccept assertion — consistent with the policy-vs-consensus split
- `feature_csv_activation.py`: remove BIP9-activation scaffolding and `nFeatures=2` paths — correct (Tapyrus does not use BIP9 nor accept nFeatures=2)
- `feature_pruning.py` + `blocktools.py::create_tx_with_large_script`: replace ~10 KB push-only non-standard scriptPubKey with `OP_RETURN` + null data — the existing test was silently producing tiny blocks and never exercising the prune path; the OP_RETURN form is standard, doesn't create UTXOs, and the `-datacarriersize=10000` plumbing is correct
- `feature_nonstd_txn.py` (new): two-node fixture (`-acceptnonstdtxn=0` vs `=1`) verifying that mempool policy diverges while consensus accepts non-standard txns in blocks — exactly the right shape
- `heavy-functional-tests.yml` (new): weekly schedule with 480-minute timeout, separate from daily matrix — good fit for `feature_pruning.py` (1.5–4 h, 4 GB disk)
- `daily-test.yml`: invoke `test_runner.py ... --debugscripts` in the Debug build_type before early exit so debug-only tests actually run

---

## 🚨 Blocker — stale base, MUST rebase

The merge base of this PR is `1f0759daa` ("Improve CI stability (#409)", merged before PR #411–414). All four security PRs are **absent from this branch**:

```
Missing on PR #417 head:
  src/softforkmanager.h
  src/issuedcolorids.{cpp,h}
  test/functional/feature_cp2sh_softfork.py
  (and all the consensus/script/wallet/RPC changes from #411–414)
```

This is why `git diff master..pr417 --stat` reports 65 files / +847 / -2414 — the deletions are stale-base artifacts, not the PR's intended changes (`git diff $(git merge-base master pr417)..pr417 --stat` shows the real scope: 10 files, +472 / -163).

**Without a rebase, two real risks at merge time:**

### B1. After rebase, `feature_cp2sh_softfork.py` could silently disappear from CI

Master's `test_runner.py:185` (added by PR #414) lists:

```python
'feature_cp2sh_softfork.py',
```

in `BASE_SCRIPTS`. PR #417's `test_runner.py` does NOT list it (it was written before PR #414 existed). The two branches will conflict on `test_runner.py`. **The rebase MUST preserve master's `feature_cp2sh_softfork.py` entry**; resolving the conflict by taking PR #417's version wholesale will silently drop the entire CP2SH softfork test from CI.

Suggested resolution: keep the new BASE_SCRIPTS ordering from PR #417 AND re-insert `feature_cp2sh_softfork.py` in the appropriate group.

### B2. Debug build_type CI run must have `CP2SH_ACTIVATION_TEST_HEIGHT=120` defined

PR #417's `daily-test.yml` change adds, in the Debug build branch:

```bash
if [ "${{ matrix.config.build_type }}" == "Debug" ]; then
    python3 test/functional/test_runner.py ... --debugscripts
    ...
fi
```

`--debugscripts` does `test_list += DEBUG_MODE_SCRIPTS` (line 346–347), so the actual list = `BASE_SCRIPTS + DEBUG_MODE_SCRIPTS`. After rebase, `BASE_SCRIPTS` will include `feature_cp2sh_softfork.py`, which **requires the binary to be built with `-DCP2SH_ACTIVATION_TEST_HEIGHT=120`** (otherwise the softfork activates at the production height of 693367 and the test will hang / timeout / falsely pass).

Verify that the Debug build_type matrix entry in `daily-test.yml` also injects `-DCP2SH_ACTIVATION_TEST_HEIGHT=120` into `CMAKE_OPTS`. If it currently only injects it for the native build path (master's `daily-test.yml:547`), this needs to be extended.

---

## ⚠️ Minor observations

### O1. `p2p_compactblocks.py` moved from DEBUG_MODE_SCRIPTS to EXTENDED_SCRIPTS

The move is justified (compact blocks are standard and should run unconditionally), but EXTENDED tests do not run on every daily build. Given the recently-confirmed M2 finding on master (`MarkBlockAsReceived` leaks `mapBlocksInFlight` entries when `MAX_CMPCTBLOCKS_INFLIGHT_PER_BLOCK` is reached), please verify EXTENDED is run at least nightly so this regression class is exercised on a regular cadence.

### O2. `heavy-functional-tests.yml` weekly cadence — confirm Sunday matches release rhythm

`schedule: cron` runs every Sunday. If release tagging happens Monday–Wednesday, the most recent heavy-test run will always be 0–2 days stale. Acceptable, but worth confirming.

### O3. `feature_csv_activation.py` CSV-active-from-genesis claim should be cross-checked

The PR removes the BIP9-activation paths because "CSV is always active from genesis in Tapyrus." This is true (Tapyrus never used `nVersion`-bit signaling), but please verify the residual CSV tests still cover both pre/post activation regimes via timing rather than version-bit signaling. The diff removes a substantial body of `bip113`, `bip68`, `bip112` test paths — confirm what remains still exercises CSV correctness.

---

## Recommended actions

**Required before merge:**
1. Rebase onto current `master`
2. Resolve `test_runner.py` conflict by **preserving master's `feature_cp2sh_softfork.py` entry** alongside PR #417's reorganization (B1)
3. Verify the Debug build_type CI entry in `daily-test.yml` passes `-DCP2SH_ACTIVATION_TEST_HEIGHT=120` to CMake (B2)
4. After rebase, post a fresh diff so reviewers can verify the security PRs (#411–414) are unaffected

**Recommended:**
5. Address O1 (compact-block test cadence)
6. Address O3 (residual CSV test coverage)

---

*Verification: merge base computed via `git merge-base master pr417` = `1f0759daa` (PR #409). Real PR scope via `git diff $(git merge-base master pr417)..pr417 --stat` = 10 files. Conflict candidates via set intersection of `git diff $(merge-base)..pr417 --name-only` vs `git diff $(merge-base)..master --name-only` = `daily-test.yml`, `test_runner.py`.*

azuchi (changes requested):
## Re-review after rebase (head `90eb83ba8`)

### ✅ Prior blockers resolved

- **B1**: `test_runner.py:191` preserves `feature_cp2sh_softfork.py` in `BASE_SCRIPTS` ✓
- **B2**: `daily-test.yml:551` injects `-DCP2SH_ACTIVATION_TEST_HEIGHT=120` for all native builds; `:698` runs `--debugscripts` in the Debug branch ✓
- Rebase clean (merge base now `54a27e153`, diff scope 10 files / +472 / -163)

### 🚨 New blocker — Smoke Test #600 fails on 4/6 jobs

Commit `2f66ee7ae` rewrote `test_framework/blocktools.py::create_tx_with_large_script` to emit `OP_RETURN [MAX_SCRIPT_SIZE - 10]` (≈10 KB). `MAX_OP_RETURN_RELAY` defaults to **83 bytes** (`src/script/standard.h:42`), so the tx is rejected by mempool standardness unless `-datacarriersize=10000` is set.

Only **one of three callers** was updated:

| Caller | Updated? |
|---|---|
| `test/functional/feature_pruning.py` | ✅ adds `-datacarriersize=10000` |
| `test/functional/feature_largeblocksize.py` | ❌ — causes the Smoke Test failure (`AssertionError: <txid> not found in mempool` at line 102) |
| `test/functional/feature_reindex.py` | ❌ — masked by `--failfast`, but same vector |

**Fix**: pass `-datacarriersize=10000` to nodes in `feature_largeblocksize.py::set_test_params` and `feature_reindex.py` (same pattern as `feature_pruning.py`).

Failing jobs: [Linux x86_64](https://github.com/chaintope/tapyrus-core/actions/runs/26833639167/job/79121585312), [macOS ARM64](https://github.com/chaintope/tapyrus-core/actions/runs/26833639167/job/79121585348), [macOS x86_64](https://github.com/chaintope/tapyrus-core/actions/runs/26833639167/job/79121585395), [Linux ARM64](https://github.com/chaintope/tapyrus-core/actions/runs/26833639167/job/79121585418).

azuchi (changes requested):
## Re-review after rebase (head `8a46a7ec8`)

### ✅ Prior blocker resolved
- `-datacarriersize=10000` is now set in both `feature_largeblocksize.py` and `feature_reindex.py` (commit `896fccc5a`).

### 🚨 New blocker — two functional tests fail on Linux daily CI

#### NB1. `feature_nonstd_txn.py` fails immediately with `KeyError: 0`

Run: https://github.com/chaintope/tapyrus-core/actions/runs/27773430807/job/82179336049

```
Traceback (most recent call last):
  File ".../test/functional/feature_nonstd_txn.py", line 60, in run_test
    assert_equal(result[0]['allowed'], False)
                 ~~~~~~^^^
KeyError: 0
```

`testmempoolaccept` was changed in PR #418 (M1 fix) to return a `UniValue::VOBJ` — a dict keyed by `hashMalFix`, not an array. Server side: `src/rpc/mempool.cpp` builds `UniValue result(UniValue::VOBJ)` and fills it via `encodePackageResult(...)`. Reference usage on master: `test/functional/rpc_packages.py:184` —

```python
result_expected={ package[0].hashMalFix: {'allowed': True},
                  package[1].hashMalFix: {'allowed': False, 'reject-reason': 'missing-inputs'}}
```

Affected call sites in the new test: lines 60, 64, 65. Suggested form for the single-tx case:

```python
from test_framework.messages import CTransaction
from io import BytesIO

tx = CTransaction()
tx.deserialize(BytesIO(hex_str_to_bytes(nonstd_hex0)))
tx.rehash()
result = node0.testmempoolaccept([nonstd_hex0])
assert_equal(result[tx.hashMalFix]['allowed'], False)
```

(Or use the flat `result['allowed']` form if Tapyrus collapses single-tx packages — please confirm via a quick repl against current master.)

#### NB2. `p2p_compactblocks.py::test_cmpctblock_failed_not_first_in_flight` fails

Run: https://github.com/chaintope/tapyrus-core/actions/runs/27773430807/job/82179335821

```
AssertionError: node should request missing txns from first peer
File ".../p2p_compactblocks.py", line 808, in test_cmpctblock_failed_not_first_in_flight
    assert "getblocktxn" in stalling_peer.last_message
```

This is the regression test for the M2 cmpctblock UAF fix from PR #418. Commit `f2c2c3806` in this PR replaces the bare `OP_TRUE` outputs with P2SH(OP_TRUE) and drops `-acceptnonstdtxn=1` from `extra_args`. After the change the node never issues `getblocktxn` to the first peer for the in-flight block — so the regression test exits before exercising the UAF code path it was designed to guard.

Likely causes to investigate (in order):
1. **State leak from a prior sub-test**: the txs built for the test might already be in node 1's mempool when the CMPCTBLOCK arrives, so `partialBlock` reconstructs the block without needing `getblocktxn`. Worth dumping `node.getrawmempool()` right before `stalling_peer.send_and_ping(...)` to confirm.
2. **Short-ID collision** caused by the size change of P2SH outputs/inputs vs the old bare-script tx layout.
3. **`-txindex` semantics on node 1** combined with the absence of `-acceptnonstdtxn` allowing the test framework to indirectly seed the mempool.

Please reproduce locally with `python3 test/functional/p2p_compactblocks.py --tracerpc --loglevel=debug` and confirm the node-side reason it skips `getblocktxn`.

---

### 🟡 Scope creep — please split out the CMake/toolchain refactor

Commit `8a46a7ec8` ("Simplify daily CI cmake command...") is a substantial build-system change that is unrelated to the stated PR goal ("move debug-mode tests to extended"):

- `cmake/module/FindLibevent.cmake` — promote imported targets to `IMPORTED_GLOBAL` for CMake 3.28+
- `cmake/module/FindQt.cmake` — switch `REQUIRED_VARS` from `Qt_DIR` to `Qt_FOUND`
- `depends/toolchain.cmake.in` — add `PKG_CONFIG_PATH` and `QT_HOST_PATH` plumbing
- `.github/workflows/daily-test.yml` — remove BOOST/ZEROMQ/Qt environment-variable injection from the workflow and rely on the toolchain file

These changes are reasonable on their own but they:
1. Make this PR much harder to review and bisect (a CI test reorganization PR shouldn't be touching `FindLibevent.cmake`).
2. Mix two independent regression surfaces — if a Qt build issue appears post-merge, git-blame points at "Fix intermittent CI failures" rather than a clearly scoped build refactor.
3. Increase the risk of conflicting with other in-flight build/CMake work.

**Request**: please pull commit `8a46a7ec8` into a separate PR titled something like "Move CMake depends configuration from workflow to toolchain file", reviewable on its own. The remaining commits here stay focused on the test reorganization.

---

### CI summary

- ✅ GitHub Actions main builds (Linux x86_64 / ARM64 / Windows / macOS x86_64 / Lint): all pass
- ❌ macOS-ARM jobs: same `sed: 1: "qtbase/src/plugins/plat ...": extra characters at the end of q command` qt-depends failure as PR #426 — infrastructure, not this PR's fault
- ❌ Linux daily-test jobs: NB1 / NB2 as above; other long-running failures are fail-fast cancellations after those

Once NB1 and NB2 are fixed and `8a46a7ec8` is moved to its own PR, this should be ready for approval.

azuchi (changes requested):
## Re-review after head `adaf13394`

### ✅ Resolved
- CMake/toolchain refactor split out into PR #428 (merged) — `cd325ec62` ✓

### 🚨 NB1 — `feature_nonstd_txn.py` not fixed (CI failing)

3 Debug CI jobs FAIL with `KeyError: 0` at line 60:

```python
result = node0.testmempoolaccept([nonstd_hex0])
assert_equal(result[0]['allowed'], False)   # <- testmempoolaccept returns a dict, not array
```

`testmempoolaccept` returns `UniValue::VOBJ` keyed by `hashMalFix` (since PR #418 M1). Reference: `feature_cltv.py:114-115`, `rpc_packages.py:184`. Fix pattern:

```python
tx0 = CTransaction(); tx0.deserialize(BytesIO(hex_str_to_bytes(nonstd_hex0))); tx0.rehash()
assert_equal(result[tx0.hashMalFix]['allowed'], False)
```

Same fix needed for lines 62 and 64.

Failing jobs: [L-X86-Deb](https://github.com/chaintope/tapyrus-core/actions/runs/28103970484/job/83215499329), [L-ARM-Deb](https://github.com/chaintope/tapyrus-core/actions/runs/28103970484/job/83215499351), [M-ARM-Deb](https://github.com/chaintope/tapyrus-core/actions/runs/28103970484/job/83215499361).

### 🚨 NB3 — `test_cmpctblock_failed_not_first_in_flight` passes vacuously

`adaf13394` registers `stalling_peer` as a cmpct peer (good — fixes the first assertion). But `attacker_peer` was not given `msg_sendcmpct`, so its CMPCTBLOCK is rejected at `net_processing.cpp:2601`:

```cpp
if (!nodestate->fSupportsDesiredCmpctVersion) return true;
```

i.e. it never reaches the `READ_STATUS_FAILED` branch (`net_processing.cpp:2630-2640`) that the PR #418 M2 UAF fix was designed to guard. The assertion

```python
assert "getblocktxn" not in attacker_peer.last_message
```

now passes — but trivially. **If the `return true` after `MarkBlockAsReceived()` is ever removed and the UAF is reintroduced, this test will still pass.**

Fix: send `msg_sendcmpct(version=1, announce=True)` from `attacker_peer` as well, before its CMPCTBLOCK.

### 🟡 Minor — `daily-test.yml` redundancy

`cd325ec62` re-injects `BOOST_ROOT` / `ZEROMQ_ROOT` / `BerkeleyDB_ROOT` / `Qt6_DIR` into the workflow, but PR #428 already sets these via `depends/toolchain.cmake.in:160-167`. Not a blocker — but `BerkeleyDB_ROOT` is fully redundant and the prior comment "toolchain sets BerkeleyDB_ROOT but not the specific header/library paths" is now correct again and should be restored.

### Verdict

Blockers: NB1 (CI red), NB3 (silent false-negative). Once both addressed, this should be approval-ready.

azuchi:
### 🟡 Request — please split commit `cd0c0ad5` into its own PR

`cd0c0ad5` ("Fix macos depends build failure: add fix_qyieldcpu_arm_acle.patch") touches `depends/packages/qt.mk`, `depends/packages/native_qt.mk`, and adds `depends/patches/qt/fix_qyieldcpu_arm_acle.patch`. These changes are unrelated to the stated PR goal (CI integration of non-standard functional tests).

Mixing them here has two downsides:
1. The macOS qt depends fix becomes hard to bisect / revert if it causes a regression later — git-blame will point at a CI-test PR rather than a focused depends fix.
2. Future depends bumps (Qt version upgrades, patch reorganization) are more likely to conflict with this PR's history.

This is the same pattern as the earlier CMake/toolchain refactor that was successfully split into PR #428 — please do the same here:

- Move `cd0c0ad5` to a separate PR titled something like *"depends: fix macOS qt build with `__yield` builtin (add `arm_acle.h` include)"*.
- Once that lands, rebase this branch and drop the commit.

The remaining commits in this PR will then stay focused on the test-policy / CI reorganization, which is what reviewers signed up to review.

azuchi (changes requested):
Two blockers, both fixed by a single rebase onto current `origin/master` (`dc7f617a6`).

### 🚨 R1 — PR reverts PR #428's already-merged CMake/toolchain changes

Commit `77d450fe0` ("Review feedback - remove daily ci simplification from this branch") removes changes that are **already on master via PR #428** (`755c07344`):

- `cmake/module/FindLibevent.cmake`: deletes the `IMPORTED_GLOBAL TRUE` promotion block and the `IMPORTED GLOBAL` flag on `add_library(libevent::${component} …)`
- `cmake/module/FindQt.cmake`: reverts `REQUIRED_VARS Qt_FOUND` back to `Qt_DIR`
- `depends/toolchain.cmake.in`: removes `PKG_CONFIG_PATH` prepend and `QT_HOST_PATH` / `QT_HOST_PATH_CMAKE_DIR` plumbing
- `depends/packages/qt.mk`: 7-line delta
- `.github/workflows/daily-test.yml`: re-adds the `BOOST_ROOT` / `ZEROMQ_ROOT` / `BerkeleyDB_ROOT` / `Qt6_DIR` env injection that PR #428 removed

Merge-base is `d08068506` (post-PR #428), so `git diff origin/master..origin/pr/417` currently shows these as net removals. Merging as-is silently undoes PR #428.

### 🚨 R2 — PR #432's Qt `arm_acle.h` fix is missing

`M-ARM-Deb-GUI+Wallet` fails at 1m31s ([log](https://github.com/chaintope/tapyrus-core/actions/runs/28370357155/job/84046371186)):

```
qtbase/src/corelib/thread/qyieldcpu.h:37:5:
  error: implicitly declaring library function '__yield'
    with type 'void ()' [-Werror,-Wimplicit-function-declaration]
```

Fixed by PR #432 (merged `dc7f617a6`).

### Fix

```bash
git rebase origin/master   # picks up PR #432, drops the now-empty 77d450fe0
git push --force-with-lease
```

Post-rebase, expected scope is the pure test/CI reorg: `feature_block.py`, `feature_cltv.py`, `feature_csv_activation.py`, `feature_nonstd_txn.py`, `feature_pruning.py`, `feature_reindex.py`, `feature_largeblocksize.py`, `p2p_compactblocks.py`, `test_framework/blocktools.py`, `test_runner.py`, `wallet_basic.py`, `daily-test.yml` (debug-scripts wiring only), and `heavy-functional-tests.yml`.

The prior NB1 (`feature_nonstd_txn.py` `hashMalFix` dict) and NB3 (`attacker_peer.send_and_ping(sendcmpct)`) fixes are intact.

azuchi (approved):

PR #418, opened 2026-06-02
V72 security audit fixes

Naviabheeman:
**One Fix for H1, H3, H4, and L4: SCRIPT_VERIFY_CP2SH_COLORED was added to STANDARD_SCRIPT_VERIFY_FLAG**
_(H1. GetSigOpCount / GetP2SHSigOpCount treat CP2SH as P2SH unconditionally 
H3. SignPSBTInput does not thread scriptVerifyFlags — CP2SH PSBT signing broken post-activation
H4. IsSolvable and remaining ProduceSignature callers drop CP2SH activation flag
L4. DataFromTransaction uses hardcoded STANDARD_SCRIPT_VERIFY_FLAGS — combinerawtransaction may lose Stage 2 CP2SH signatures)_

CP2SH activates at block 0 on every network except testnet (networkId 1939510133). Testnet defers it to height 693367, but testnet is already at height 716467+ and testate synch is already successful with this soft fork code. so no pre-activation testnet tx would ever need to sign without the flag. Block validation (GetBlockScriptFlags) is independent — it starts from SCRIPT_VERIFY_NONE and uses IsActive(blockHeight), so
pre-activation testnet blocks are still validated correctly without the flag.

Note on H1: As the testnet synch is already verified with the soft fork code we are sure that there are no blocks with high sigOps that fail validation. SO the gating is not done.

**L3. XFieldMaxBlockSize has no upper bound — federation can partition the network**
This is not correct. A larger block is transmitted by splitting into multiple 4mb messages. We have a functional test that successfully tests unto 7 MB feature_largeblocksize. The conclusion that the larger blocks cause the network to halt is not correct.

#### Fixed issues
**H2. testmempoolaccept actually submits and broadcasts transactions**
Description: testmempoolaccept and submitpackage share the same code path in SubmitPackageToMempool. The function calls AcceptToMemoryPool(tx, opt) without ever setting opt.flags = MempoolAcceptanceFlags::TEST_ONLY. The TEST_ONLY flag exists (validation.cpp:967 short-circuits before insertion) but testmempoolaccept does not opt in.

**M1. SubmitPackageToMempool is non-atomic — partial admission leaves orphaned middle txs and gossips invalid txs**
Description: SubmitPackageToMempool processes each tx of the package sequentially through AcceptToMemoryPool inside cs_main. If tx N is rejected (e.g. mempool size limit exceeded after N-1 already admitted), txs 1..N-1 stay in the mempool. There is no rollback, and RelayTransaction has already been called for the admitted prefix.

**M2. MarkBlockAsReceived leaks mapBlocksInFlight entries when MAX_CMPCTBLOCKS_INFLIGHT_PER_BLOCK is reached**
Description: When mapBlocksInFlight.count(hash) >= MAX_CMPCTBLOCKS_INFLIGHT_PER_BLOCK (3), MarkBlockAsReceived logs and return false without erasing this peer's entry from mapBlocksInFlight or vBlocksInFlight, and without decrementing the in-flight counters. The function then exits without touching the matching entry.

**M3. Address-processing token bucket is a single shared global, not per-peer**
Description: m_addr_token_bucket, m_addr_token_timestamp, m_addr_rate_limited, m_addr_processed are declared inside an anonymous namespace at file scope, so they are global across all peers. The surrounding code comments describe them as if they were per-peer rate-limiting state. Bitcoin Core upstream moved this state into CNode for exactly this reason.

**L1. RollforwardBlock does not stage colorId issuances — g_colorid_state can desync after crash+replay**
RollforwardBlock only spends inputs and adds coins; it never inspects outputs for NON_REISSUABLE / NFT colorIds and never calls g_colorid_state->Insert. The rollback side (DisconnectBlock, chainstate sandbox) does manage the state correctly. Practical exploitability is low because NON_REISSUABLE / NFT colorIds are Hash(prevout) of an already-spent TPC input, so duplicate-issuance via this vector is blocked by UTXO uniqueness.

**L2. Mempool does not re-evaluate already-admitted transactions against new script flags when CP2SH activates**
AcceptToMemoryPoolWorker correctly queries flags for chainActive.Tip()->nHeight + 1, but a tx admitted at height H-1 with no CP2SH flag remains in the mempool indefinitely. removeForBlock only removes mined txs;
removeForReorg only re-checks CheckFinalTx / CheckSequenceLocks. Around the activation block, miners using the local mempool may include txs that pass admission but fail ConnectBlock, producing invalid block templates.

**L5. bumpfee consumes reserved keypool key even when signing/commit fails**
In the no-TPC-change bumpfee path the wallet reserves a fresh change key (bumpfeeReserveKey.GetReservedKey, line 3315) and immediately calls bumpfeeReserveKey.KeepKey() at line 3349 — before feebumper::CreateTransaction / CommitTransaction complete. On signing failure, the key is permanently removed from the keypool but never used. Repeated failed bumpfee calls can drain the keypool.


azuchi (commented):
## Review Summary

Reviewed all 10 commits against the original audit findings. The PR resolves 8 of the prior findings cleanly and CI is green (20/20). Two items deserve follow-up before merge — neither hard blockers, but worth addressing.

---

### ✅ Verified fixes

| Finding | Fix | Notes |
|---|---|---|
| **H2** testmempoolaccept submits/broadcasts | `opt.flags = TEST_ONLY` in `rpc/mempool.cpp:115`; short-circuit at `validation.cpp:969` | Correct |
| **H3/H4/L4** Signing APIs drop CP2SH flag | Added `SCRIPT_VERIFY_CP2SH_COLORED` to `STANDARD_SCRIPT_VERIFY_FLAGS` (`interpreter.h:124`) | Elegant — `ProduceSignature` / `SignPSBTInput` / `IsSolvable` / `DataFromTransaction` all inherit through defaults |
| **M1** Partial-package gossip amplification | `RelayTransaction` moved outside the per-tx loop, gated on full success + non-`TEST_ONLY` (`policy/packages.cpp:121`) | Gossip-amplification resolved. The orphan-middle-tx aspect of M1 (mempool atomicity) remains, but is a separate hardening item |
| **M2** `mapBlocksInFlight` leak on `MAX_CMPCTBLOCKS_INFLIGHT_PER_BLOCK` | Removed the early `return false`; cleanup loop now runs unconditionally (`net_processing.cpp:397-400`) | Verified the cleanup loop correctly erases `from_peer`'s entry |
| **M3** Address rate-limit globals shared across peers | All four fields moved into `CNodeState`, two short `cs_main` critical sections wrap the bucket snapshot / write-back | Per-peer state confirmed; address processing still runs outside `cs_main` |
| **L1** `RollforwardBlock` desyncs `g_colorid_state` | Issuance detection added before `SpendCoin`, batched insert post-loop, mirrors `ConnectBlock` pattern (`chainstate.cpp:1581-1618`) | Correct |
| **L2** Mempool not re-evaluated on activation boundary | New `CSoftForkManager::ActivationFlagsChange()` + `CTxMemPool::removeForScriptFlagChange()`, invoked from `ConnectTip` after `removeForBlock` | Boundary semantics correct: `(pindexNew->nHeight, +1)` correctly fires at the transition block |
| **L5** `bumpfee` consumes keypool key on signing failure | `KeepKey()` moved past `CommitTransaction` (`wallet/rpcwallet.cpp:3357`) | Correct |

The additional commit `8bb84399c` (mandatoryFlags on `CheckInputs`) is a necessary follow-on to the STANDARD-flag change and is correctly scoped: `tipScriptFlags` is passed as `mandatoryFlags`, so flags active at `nextBlockHeight` are still treated as consensus-mandatory even after being added to STANDARD.

---

### 🟡 Items to address before merge

#### 1. Document the H1 deferral in code (recommended)

The rationale ("testnet sync verified → no historical CP2SH-shaped blocks bust the sigop limit") is defensible given:

- Production activates CP2SH at block 0 (no risk window on prod)
- Testnet activation height 693367 is already past current height (716467+)
- v0.7.2 nodes all count sigops identically (no intra-version split)
- Residual chain-split risk is only between v0.7.1 and v0.7.2 on testnet, where the federation is trusted

**Suggestion**: add a comment at `src/script/script.cpp:184` and `src/consensus/tx_verify.cpp:133` explaining *why* CP2SH sigop counting is intentionally not gated. Without it, a future maintainer is likely to either "fix" this or get burned by the asymmetry.

#### 2. Lock contract on `removeForScriptFlagChange`

`src/txmempool.cpp:604` says *"Caller must hold cs_main (for pcoinsTip access)."* but only acquires `cs`. The current caller in `ConnectTip` does hold `cs_main`, so this works today, but the API is a footgun.

**Suggestion**:
```cpp
void CTxMemPool::removeForScriptFlagChange(unsigned int newMempoolScriptFlags)
{
    AssertLockHeld(cs_main);  // pcoinsTip access
    LOCK(cs);
    ...
}
```

---

### nits (non-blocking)

- **`CSoftForkManager::ActivationFlagsChange`** returns `after` whenever `before != after`. If a flag were ever *removed* in a future fork definition, this would also trigger re-evaluation against the now-reduced flag set, which is not the intent. Consider `(after & ~before)` to express "newly-activated flags only" — cosmetic for now since all current forks only activate.
- **`src/policy/packages.cpp`** ends without a trailing newline (`\ No newline at end of file` in the diff).
- **L3 rebuttal** — the response ("4 MB chunking handles transmission") addresses transport, not the original concern (federation can set `XFieldMaxBlockSize` to a value that causes OOM during validation). The dismissal is still acceptable on "federation is trusted" grounds, but the technical argument as stated doesn't match.

---

No objection to merging once the H1 comment is added and the `AssertLockHeld(cs_main)` is in place. Nice work on this round.

azuchi (approved):

PR #424, opened 2026-06-15
Accept arm64 or aarch64 as the arch type in depends build

Naviabheeman:
When CI was running on macos-14 (darwin24), and config.guess there returned `arm64-apple-darwin24..`
GitHub Actions recently upgraded macos-latest to macOS 15 (darwin25). On macOS 15, the newer config.guess switched to the GNU canonical name `aarch64-apple-darwin25...` So the build is not identified correctly.


azuchi:
This already merged by #425 .

PR #425, opened 2026-06-15
V72 423 vulnerability fix

Naviabheeman:
Fix issues seen in Pr#423:

- H1 — `wait*` RPC 10-min cap only fires when `timeout <= 0`; `INT_MAX` bypass exhausts the RPC thread pool
- H2 — `-persistmempool=true` still silently disables mempool persistence
- M1 — HTTPAUTH backoff LRU eviction picks the wrong victim
- M4 — `ConnectBlock` DoS-score wrapper bans honest relayers on `bad-txns-premature-spend-of-coinbase`
- M5 — `VerifyTokenBalances` `tpcin <= 0` upgraded from DoS=0 → DoS=100 at mempool level
- And all low priority feedback

M2 - IPv6 is not supported in tapyrus p2p layer too. this should be a separate feature in HTTP and P2P layers.
M3 - restricting body size might affect genuine large transactions. we should identify attach using persistent high memory or cpu usage

### DIFF
--- a/depends/builders/darwin.mk
+++ b/depends/builders/darwin.mk
@@ -1,7 +1,12 @@
-# Strip Darwin version suffix from build triplet so native-build detection works.
-# config.guess returns e.g. arm64-apple-darwin24.5.0 while HOST is arm64-apple-darwin,
-# so strip the trailing version to make the $(host) == $(build) comparison succeed.
-build:=$(shell echo "$(build)" | sed 's/\(.*-apple-darwin\)[0-9.]*$$/\1/')
+# Normalize the build triplet so native-build detection works.
+# config.guess on Apple Silicon returns aarch64-apple-darwin25.x.x; the CI sets
+# HOST=arm64-apple-darwin (from uname -m).  Two normalizations are needed:
+#   1. Strip the Darwin version suffix (24.5.0 → nothing)
+#   2. Rename aarch64 → arm64 to match the HOST naming convention
+# Without both, $(host) != $(build) and a native arm64 build is wrongly treated
+# as cross-compilation, causing the GNU-only `sed -i` in qt.mk to run under
+# BSD sed and fail with "extra characters at the end of q command".
+build:=$(shell echo "$(build)" | sed 's/\(.*-apple-darwin\)[0-9.]*$$/\1/' | sed 's/^aarch64-apple-darwin$$/arm64-apple-darwin/')
 
 build_darwin_CC:=$(shell xcrun -f clang) -isysroot$(shell xcrun --show-sdk-path)
 build_darwin_CXX:=$(shell xcrun -f clang++) -isysroot$(shell xcrun --show-sdk-path)
--- a/depends/packages/qt.mk
+++ b/depends/packages/qt.mk
@@ -270,10 +270,15 @@ ifeq ($(host),$(build))
 endif
 # Fix CGDisplayCreateImageForRect obsoleted in macOS 15.0
 # Comment out the call and return empty pixmap (screen grab will fail gracefully)
-# Only needed for cross-compilation (native macOS builds use BSD sed which requires -i '')
+# Cross-compilation only (native darwin builds already satisfy host==build and are excluded).
+# BSD sed (build machine is macOS) needs -i '' while GNU sed (Linux build machine) uses -i.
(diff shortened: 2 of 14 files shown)


azuchi (changes requested):
## PR #425 review

Strong fixes overall — H1/H2/M1/M4/M5 land correctly with good regression tests. Three blockers must be resolved before CI can pass.

### 🚨 Blockers

**1. C++ build failure (`src/rpc/blockchain.cpp:217, 263, 309`)**

```cpp
const int64_t ms = std::max(0LL, std::min((int64_t)timeout, MAX_WAIT_FOR_BLOCK_MS));
```

`std::max` template deduction fails on LP64 Linux (gcc 13): `0LL` is `long long`, `int64_t` resolves to `long`. CI fails across all Linux/ARM matrix entries. Fix:

```cpp
const int64_t ms = std::clamp<int64_t>(timeout, 0, MAX_WAIT_FOR_BLOCK_MS);
```

**2. Lint failure (`src/util.cpp:171`)**

The new case-insensitive `iequals` lambda uses `tolower((unsigned char)s[i])` — locale-dependent. `lint-locale-dependence` fails.

```cpp
auto ascii_tolower = [](unsigned char c) { return (c >= 'A' && c <= 'Z') ? c + 32 : c; };
```

Or use `ToLower` from `src/utilstrencodings.h`.

**3. Functional test / implementation mismatch (`test/functional/rpc_blockchain.py:282-289`)**

The test expects negative timeouts to be rejected:

```python
assert_raises_rpc_error(-8, \"timeout must be non-negative\",
    node.waitforblockheight, current_height + 1, bad_timeout)
```

But the implementation clamps negative to 0 via `std::max(0LL, ...)` and returns the current tip without throwing. Either:

- Add `if (timeout < 0) throw JSONRPCError(RPC_INVALID_PARAMETER, \"timeout must be non-negative\");` (matches the audit recommendation and is cleaner)
- Or update the test to expect an immediate return

---

### ✅ Verified fixes

- **H1** — positive-timeout cap restored; INT_MAX no longer bypasses. Indefinite `timeout=0` preserved as documented (intentional design trade-off — worth a release-notes line about the residual 4-worker pin risk for authenticated callers).
- **H2** — `InterpretBool` accepts case-insensitive `true/yes/on/false/no/off` plus full-string-consumed numeric check (so `\"1abc\"` no longer silently becomes `true`) plus an operator-visible `LogPrintf` warning for unrecognized values. The 27-case unit test in `getarg_tests.cpp` is thorough.
- **M1** — `min_element → max_element` for victim selection; bonus refactor adds an atomic `g_failed_auths_size` for a lock-free success-path fast path and trims the critical section to map updates only. This single change also closes L5 (lock-across-IO) and L7 (success-path always locks). Excellent.
- **M4** — wrapper reverted to plain `error(...)` so `bad-txns-premature-spend-of-coinbase` keeps DoS=0; regression test `connectblock_premature_coinbase_dos_score_is_zero` covers it.
- **M5** — reverted to `state.Invalid(false, REJECT_INSUFFICIENTFEE, ...)`; regression test `verifytoken_no_tpc_input_dos_score_is_zero` covers it.
- **L1** (dumptxoutset Windows) — predicate inverted to `rel.empty() || *rel.begin() == \"..\"`. Exactly the recommended fix.
- **L2** (P2P-1A symmetric) — `return true;` added after `MarkBlockAsReceived` at line 2689.
- **L6** (VerifyDB Level-4) — added `LogPrintf(\"[ABORTED] ...\")` + incomplete-verification message. Conservative (still returns \`true\`), but operator visibility is the main goal here.
- **L8** (ADDR AddressCount gate) — restored.

### Excluded findings (rationale acknowledged)

- **M2** (IPv6 /64 bucketing) — fair to defer; IPv6 is not first-class in the P2P layer either, and an HTTP-layer fix in isolation has limited value.
- **M3** (REST pre-read cap) — fair to defer with the caveat that legitimate large transactions should be size-bounded by consensus (block size), so a smaller per-request cap on REST endpoints would still be safe. Worth tracking for v0.7.3.

### Recommendation

Approve once the three blockers are resolved. Once green, ready to merge to master and backport to v0.7.

Excellent quality of fixes — the M1 atomic-counter design is particularly nice.

azuchi (approved):

PR #428, opened 2026-06-19
Fix native qt build and ci

Naviabheeman:
Simplify cmake command in daily ci by moving depends library paths to toolchain file. 
Fix daily CI failure in macOS depends qt build on arm64 by using a patch file instead of the sed command which was failing

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -523,24 +523,9 @@ jobs:
 
           # Common CMake configuration options
           CMAKE_OPTS=(
-            "-DENABLE_ZMQ=ON"
             "-DCMAKE_BUILD_TYPE=${{ matrix.config.build_type }}"
             -DCMAKE_PREFIX_PATH="$DEPENDS_PREFIX"
           )
-
-          # Set Qt6_DIR explicitly if GUI is enabled to help CMake find Qt6
-          if [ "${{ matrix.config.enable_gui }}" == "true" ]; then
-            QT6_CMAKE_DIR="$DEPENDS_PREFIX/lib/cmake/Qt6"
-            if [ -d "$QT6_CMAKE_DIR" ]; then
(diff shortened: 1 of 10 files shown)


azuchi (changes requested):
## 🚨 Blocker — `QT_HOST_PATH` points to wrong directory

All `GUI=true` jobs fail with `Could NOT find Qt6CoreTools (missing: Qt6CoreTools_DIR) → Failed to find required Qt component "Core"`. All `GUI=false` jobs pass.

**Cause**: `depends/packages/qt.mk` now builds `native_qt` unconditionally:

```diff
-ifneq ($(host),$(build))
 $(package)_dependencies := native_$(package)
-endif
```

So native_qt tools install into `build_prefix` = `$(host_prefix)/native/` (per `depends/Makefile:83`). But `depends/toolchain.cmake.in` native branch sets:

```cmake
set(QT_HOST_PATH "${CMAKE_CURRENT_LIST_DIR}" ...)  # = host_prefix, missing /native
```

Qt6CoreConfig.cmake then looks for `Qt6CoreTools` at `$QT_HOST_PATH/lib/cmake/Qt6CoreTools/` and misses it (real location: `$host_prefix/native/lib/cmake/Qt6CoreTools/`).

**Fix** — either:

```cmake
# A. Update toolchain to follow the new layout
else()
  set(QT_HOST_PATH "${CMAKE_CURRENT_LIST_DIR}/native" CACHE PATH ...)
endif()
```

or restore the `ifneq ($(host),$(build))` guards in qt.mk for both `native_$(package)` dep and `-qt-host-path $(build_prefix)`.

Example failing job: https://github.com/chaintope/tapyrus-core/actions/runs/27831029998/job/82367626588

Rest of the PR (miniupnpc URL, fallback download fix, MAKEOVERRIDES hardening, `build_TAR`/`build_TOUCH` abstraction, CMake refactor) looks good.

azuchi (approved):

PR #433, opened 2026-06-29
Port fix CVE-2024-52911

Naviabheeman:
Tapyrus master already mitigatesCVE-2024-52911via commit 7978d1536 (declares txdata before CCheckQueueControl in ConnectBlock) — this PR brings the code in line with upstream's PR #31112 that send the specific ScriptError from parallel validation

**Details:**

> So far, detailed information about script validation failures is only reported when running with -par=1, due to a lack of ability to transfer information from the script validation threads to the validation thread. Fix this by extending the CCheckQueue functionality to pass more results through than just success/failure, and use this to report the exact Script error, as well as the transaction input in which it occurred.

### DIFF
--- a/src/bench/checkqueue.cpp
+++ b/src/bench/checkqueue.cpp
@@ -7,6 +7,7 @@
 #include <validation.h>
 #include <checkqueue.h>
 #include <prevector.h>
+#include <optional>
 #include <vector>
 #include <random.h>
 
@@ -29,9 +30,9 @@ static void CCheckQueueSpeedPrevectorJob(benchmark::State& state)
         explicit PrevectorJob(FastRandomContext& insecure_rand){
             p.resize(insecure_rand.randrange(PREVECTOR_SIZE*2));
         }
-        bool operator()()
+        std::optional<bool> operator()()
         {
-            return true;
+            return std::nullopt;
         }
         void swap(PrevectorJob& x){p.swap(x.p);};
     };
@@ -50,7 +51,7 @@ static void CCheckQueueSpeedPrevectorJob(benchmark::State& state)
         }
         // control waits for completion by RAII, but
         // it is done explicitly here for clarity
-        control.Wait();
+        control.Complete();
     }
 }
 BENCHMARK(CCheckQueueSpeedPrevectorJob, 1400);
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -720,8 +720,8 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                                block.vtx[0]->GetValueOut(ColorIdentifier()), blockReward),
                                REJECT_INVALID, "bad-cb-amount");
 
-    if (!control.Wait())
-        return state.DoS(100, error("%s: CheckQueue failed", __func__), REJECT_INVALID, "block-validation-failed");
+    if (auto scriptErr = control.Complete())
+        return state.DoS(100, error("%s: CheckQueue failed: %s", __func__, ScriptErrorString(*scriptErr)), REJECT_INVALID, "block-validation-failed");
     int64_t nTime4 = GetTimeMicros(); nTimeVerify += nTime4 - nTime2;
     LogPrint(BCLog::BENCH, "    - Verify %u txins: %.2fms (%.3fms/txin) [%.2fs (%.2fms/blk)]\n", nInputs - 1, MILLI * (nTime4 - nTime2), nInputs <= 1 ? 0 : MILLI * (nTime4 - nTime2) / (nInputs-1), nTimeVerify * MICRO, nTimeVerify * MILLI / nBlocksTotal);
 
--- a/src/chainstate.cpp
+++ b/src/chainstate.cpp
@@ -668,8 +668,9 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
             std::vector<CScriptCheck> vChecks;
             bool fCacheResults = fJustCheck; /* Don't cache results if we're actually connecting blocks (still consult the cache, though) */
             if (!CheckInputs(tx, state, view, fScriptChecks, GetBlockScriptFlags(pindex), fCacheResults, fCacheResults, txdata[i], nScriptCheckThreads ? &vChecks : nullptr, GetBlockScriptFlags(pindex)))
-                // FormatStateMessage is evaluated before DoS() overwrites state, preserving
-                // the per-input detail in the log while enforcing DoS 100 at the block level.
+                // FormatStateMessage is evaluated before DoS() re-sets state, preserving
+                // the per-input detail in the log.  state.GetRejectReason() already holds
+                // the specific error string set by CheckInputs, so it is propagated as-is.
                 //
                 // DoS scoring asymmetry (informational, not a bug):
                 // - Multi-thread (nScriptCheckThreads > 0): script checks are deferred to
@@ -684,7 +685,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                 //   which is why it cannot simply be removed.
                 return state.DoS(100, error("ConnectBlock(): CheckInputs on %s failed with %s",
                     tx.GetHashMalFix().ToString(), FormatStateMessage(state)),
-                    REJECT_INVALID, "mandatory-script-verify-flag-failed");
+                    REJECT_INVALID, state.GetRejectReason());
             control.Add(std::move(vChecks));
         }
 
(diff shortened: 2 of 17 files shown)


azuchi (changes requested):
## Review

**Verified scope**: head `d6db4b724`, 2 commits / 15 files / +124 / -126 vs current `origin/master` (`d08068506`).

The mechanical port of upstream PR #31112 in `e57494d3c` is well-executed — template parameter `R` deduced from `T::operator()()`, `std::optional<R>` aggregation, `Wait() → Complete()` rename, and Tapyrus' predicate-based cond-var wait + minimum-batch-size fix are all preserved correctly. The second commit `d6db4b724` ("unify script validation error reporting") is where my concerns are.

---

### 🚨 C1 — Blocker: `src/test/miner_tests.cpp:473` not updated, breaks CI

```cpp
// Should throw block-validation-failed
BOOST_CHECK_EXCEPTION(AssemblerForTest(Params()).CreateNewBlock(scriptPubKey),
    std::runtime_error, HasReason("block-validation-failed"));
```

The PR changes the block-path reject reason from `block-validation-failed` to `script-verification-failed`, but this assertion was missed. Result: `miner_tests/CreateNewBlock_validity` fails on every Linux x86_64 / ARM64 / macOS CI run ([Linux x86_64 log](https://github.com/chaintope/tapyrus-core/actions/runs/28374537619/job/84060696251)):

```
test/miner_tests.cpp(473): error: in "miner_tests/CreateNewBlock_validity":
  exception "std::runtime_error" raised as expected:
  validation on the raised exception through predicate "HasReason("block-validation-failed")"
```

Update the assertion to `HasReason("script-verification-failed")` and re-grep the tree for any other stragglers.

---

### 🚨 C2 — Logic bug: `flags == SCRIPT_VERIFY_NONE` is the wrong discriminator for "block path"

In `src/validation.cpp` `CheckInputs`:

```cpp
// For block validation (flags == SCRIPT_VERIFY_NONE), use a generic reject
// reason so the P2P message is consistent across all script error types.
// For mempool (flags != 0), expose the specific error for RPC callers.
const char* reject_reason = (flags == SCRIPT_VERIFY_NONE)
    ? "script-verification-failed"
    : ScriptErrorString(*err);
```

`GetBlockScriptFlags(pindex)` (validation.cpp:632–642) returns:

```cpp
unsigned int flags = SCRIPT_VERIFY_NONE;
if (GetSoftForkManager().IsActive(SCRIPT_VERIFY_CP2SH_COLORED, blockHeight))
    flags |= SCRIPT_VERIFY_CP2SH_COLORED;
return flags;
```

So once the CP2SH softfork activates (height 693367 in production, 120 in CI runs with `CP2SH_ACTIVATION_TEST_HEIGHT=120`), `flags == SCRIPT_VERIFY_CP2SH_COLORED` (non-zero), and **single-thread block validation falls into the mempool branch** — leaking the specific script error over P2P:

| Path | Pre-CP2SH-activation | Post-CP2SH-activation |
|---|---|---|
| Parallel (`nScriptCheckThreads > 0`) — `chainstate.cpp:723` | `script-verification-failed` | `script-verification-failed` |
| Single-thread block (`nScriptCheckThreads == 0`) — this branch | `script-verification-failed` | **`ScriptErrorString(*err)`** ← bug |
| Mempool — this branch | `ScriptErrorString(*err)` | `ScriptErrorString(*err)` |

Consequences:
- P2P reject reasons become inconsistent depending on `-par` setting × block height
- Functional tests `feature_cltv.py`, `feature_dersig.py`, `feature_serialization.py`, `feature_csv_activation.py` hard-assert `b'script-verification-failed'` — they will fail in single-thread + post-activation runs.

**Fix**: don't infer caller intent from `flags`. Either
1. Add an explicit `bool fBlockPath` parameter to `CheckInputs`, or
2. Drop the block-path generic-reason logic from `CheckInputs` entirely and let `chainstate.cpp` overwrite the reject reason after `CheckInputs` returns false in the single-thread path (mirrors what the parallel `control.Complete()` branch already does).

Option 2 keeps `CheckInputs` mempool-shaped and is simpler.

---

### 🟡 D1 — Please clarify rationale in PR description

Two things would help reviewers and future readers of `git log`:

1. **CVE status**: Tapyrus master already mitigates CVE-2024-52911 via commit `7978d1536` (*"declare txdata before CCheckQueueControl in ConnectBlock"*) using a different mechanism (declaration-order trick that exploits C++ destruction order) from upstream PR #31112 (which fixes it by removing the early-return path). Please add a note like:

   > Tapyrus already mitigates CVE-2024-52911 via `7978d1536`. This PR brings the codebase in line with upstream's structural fix and additionally enables the parallel script validator to report specific `ScriptError` values.

2. **Direction inverted from upstream**: upstream PR #31112's stated goal is to **make parallel validation report the specific `ScriptError`** (matching single-thread behavior). The second commit here goes the **opposite way** — it makes single-thread block validation emit a generic `script-verification-failed` to match parallel's generic message, instead of letting parallel emit the specific error to match single-thread. The end result (uniform reject reason) is reasonable, but it discards information upstream worked to preserve. Please explain why in the PR description / commit message (privacy / fingerprinting concern? matching the existing parallel-path behavior to avoid touching more code? something else?).

   Worth considering as an alternative: keep the upstream direction — let block validation emit the specific `ScriptErrorString` in both modes. That stays closer to upstream intent and avoids the C2 discriminator entirely.

---

### 🟡 M1 — `bool do_work;` declared uninitialized before the loop

`checkqueue.h` `Loop()`:

```cpp
std::optional<R> local_result;
bool do_work;          // uninitialized
do {
    {
        WaitableLock lock(mutex);
        ...
        do_work = !m_result.has_value();   // assigned inside lock
    }
    if (do_work) { ... }
} while (true);
```

A future refactor that adds an early `return` between the declaration and the assignment would invoke UB. Cleanest fix is to declare `do_work` inside the loop body:

```cpp
do {
    bool do_work;
    {
        WaitableLock lock(mutex);
        ...
        do_work = !m_result.has_value();
    }
    if (do_work) { ... }
} while (true);
```

---

### 🟡 M2 — `feature_csv_activation.py:264` test-height vs CP2SH activation

```python
self.sync_blocks([self.create_test_block([bip112tx_special_v1])], success=False,
    reject_code=16, reject_reason=b'script-verification-failed')
```

This works only if the BIP112 test block height stays below `CP2SH_ACTIVATION_TEST_HEIGHT`. Once C2 is fixed properly this becomes moot, but worth a quick check.

---

### Summary

- **C1** (block-on-merge): update `src/test/miner_tests.cpp:473`
- **C2** (block-on-merge): replace the `flags == SCRIPT_VERIFY_NONE` discriminator with an explicit "is-block-path" signal, or shift the responsibility to `chainstate.cpp`
- **D1**: extend PR description with CVE status + rationale for inverting upstream's direction
- **M1**: avoid uninitialized `bool do_work;`
- **M2**: recheck after C2 fix

The upstream-port commit `e57494d3c` is solid. The unification commit needs to be reworked.

azuchi (changes requested):
## Re-review after head `0238a1bfe`

Great progress — commit `5ae2ae81c` correctly reverses direction to follow upstream PR #31112 (specific `ScriptError` in every path), which eliminates the C2 discriminator concern from the previous review. Commit `6d6240a3d` updates `miner_tests.cpp:473` (C1) and CP2SH softfork test. Commit `0238a1bfe` updates the NULLDUMMY string.

### ✅ Prior findings resolved

- **C1** — `src/test/miner_tests.cpp:473` now checks the specific string `mandatory-script-verify-flag-failed (Script evaluated without error but finished with a false/empty top stack element)` ✓
- **C2** — the `flags == SCRIPT_VERIFY_NONE` discriminator is gone. `CheckInputs` uniformly emits `mandatory-script-verify-flag-failed (%s)`, and `chainstate.cpp:723` (parallel path) does the same via `control.Complete()`, so P2P reject reasons are consistent across `-par` settings and softfork activation. Single-thread block path in `ConnectBlock` now propagates `state.GetRejectReason()` as-is. ✓
- **M1** — `bool do_work;` moved inside the do-while body ✓
- **M2** — `feature_csv_activation.py:264` and `feature_dersig.py` / `feature_cltv.py` / `feature_serialization.py` all updated to the new `mandatory-script-verify-flag-failed (SpecificError)` form ✓
- **D2** — direction now matches upstream; no further explanation needed ✓

### 🚨 Blocker — one functional test was missed

`test/functional/mempool_accept.py:262` still expects the interim (unwrapped) form from commit `d6db4b724` and was not reverted in `5ae2ae81c` to match the new unified format:

```python
# Current (fails on Linux x86_64, macOS ARM64, macOS x86_64):
result_expected={tx.rehash():{ 'allowed': False,
    'reject-reason': '16: Signature must be zero for failed CHECK(MULTI)SIG operation'}},

# Should be:
result_expected={tx.rehash():{ 'allowed': False,
    'reject-reason': '16: mandatory-script-verify-flag-failed (Signature must be zero for failed CHECK(MULTI)SIG operation)'}},
```

CI proof ([Linux x86_64 log](https://github.com/chaintope/tapyrus-core/actions/runs/28452290729/job/84317641670)):

```
mempool_accept.py failed, Duration: 18 s
AssertionError: not(
  {'…': {'allowed': False, 'reject-reason': '16: Signature must be zero for failed CHECK(MULTI)SIG operation'}}
    ==
  {'…': {'allowed': False, 'reject-reason': '16: mandatory-script-verify-flag-failed (Signature must be zero for failed CHECK(MULTI)SIG operation)'}}
)
```

Since mempool_accept.py runs in the `--failfast` fast-test bucket, its failure short-circuits the rest of the daily-test matrix and is what's actually turning the CI matrix red — everything else that ran (Debug jobs, cross builds, unit tests) is green.

### 🟡 D1 (recap, optional)

Not a blocker, and the title is fine (upstream disclosure confirms PR #31112 is the CVE-2024-52911 fix). Consider adding a one-liner to the PR description noting that **Tapyrus master already mitigates CVE-2024-52911 via commit `7978d1536`** (declares `txdata` before `CCheckQueueControl` in `ConnectBlock`) — this PR brings the code in line with upstream's structural fix and additionally surfaces the specific `ScriptError` from parallel validation. Helps future readers understand why the port is happening on top of an already-mitigated CVE.

### Summary

Fix `mempool_accept.py:262` (mirror the wrapping used elsewhere) and this is ready to approve.

azuchi (approved):

PR #436, opened 2026-07-02
IPv6 support: fix link-local detection, observability, and test coverage

Naviabheeman:
Core networking fixes

src/netaddress.h / src/netaddress.cpp
- Added IsLinkLocal() implementing the correct RFC 4291 §2.5.6 check (fe80::/10, bitwise) — the old IsRFC4862() only matched /64, silently passing fe81:: through febf:: as routable.
- IsRFC4862() now delegates to IsLinkLocal() for backward compatibility.
- IsRoutable() updated to call IsLinkLocal() instead of IsRFC4862().
- Comment added to GetGroup() explaining the IPv6 /32 group-key design.

src/net.cpp
- AddLocal: added warning log when a non-routable address is passed via -externalip (e.g. link-local fe80::1) — was silently dropped before.
- ThreadMapPort (UPnP): early-return with log when NET_IPV4 is unreachable — UPnP IGD discovers only IPv4 external addresses, so attempting it on a v6-only node is pointless.
- InitBinds default wildcard path: changed IPv4 wildcard bind from conditional BF_NONE/BF_REPORT_ERROR to unconditional BF_REPORT_ERROR so a failed IPv4 bind always reports to the user. Added comment documenting the intentional BF_NONE on IPv6 (IPv6 absence is tolerable; IPv4 is not).

src/netbase.cpp
- CreateSocket: socket() syscall failure now logs "socket() failed for <addr>: <errno>" — was silently swallowed before, for both families.
- CreateSocket: fixed mislabelled log message "ConnectSocketDirectly: Setting socket to non-blocking failed" → "CreateSocket: ...".

src/zmq/zmqpublishnotifier.cpp
- Sets ZMQ_IPV6 = 1 on the PUB socket before zmq_bind — without this, libzmq defaults to IPv4-only on some builds and silently fails to bind IPv6 endpoints like tcp://[::1]:port.

---
Help text / documentation

src/init.cpp
- -addnode, -connect: updated arg name to <ip[:port]>, added IPv6 bracket-notation example.
- -externalip: added IPv6 bracket-notation example.
- -whitelist: added IPv6 address and CIDR examples.
- -rpcallowip: added IPv6 address and CIDR examples.
- -zmqpubhashblock/tx/rawblock/rawtx: added dual IPv4+IPv6 endpoint examples.

src/tapyrus-cli.cpp
- -rpcconnect: added note about IPv6 bracket notation ([::1]).

src/addrman.h
- Comment clarifying that nVersion in Serialize/Unserialize only controls bucket-position trust, not entry deserialization — all IPv4 and IPv6 entries are always read unconditionally.

src/httpserver.cpp
- Comment explaining the 127.0.0.0/8 vs ::1/128 loopback asymmetry (RFC 5735 vs RFC 4291 §2.5.3).
- Comment documenting that evhttp_bind_socket_with_handle handles bare IPv6 strings via internal getaddrinfo (libevent ≥ 2.1.12).

---
Functional tests

test/functional/test_framework/netutil.py
- all_interfaces() extended to include IPv6 interfaces via /proc/net/if_inet6 (was IPv4-only via SIOCGIFCONF). Variable renamed bytes → bytes_needed to avoid shadowing the builtin.

test/functional/rpc_bind.py
- Loopback-skip now filters both 127.0.0.1 and ::1 when searching for a non-loopback interface.

test/functional/feature_ipv6.py (new)
- 8 sub-tests guarded by test_ipv6_local(): P2P connection via [::1]; addnode with IPv4+IPv6; -connect=[::1]; -externalip routable IPv6 accepted; -externalip=[fe80::1] rejected with log warning; -whitelist IPv4/IPv6 CIDR; -rpcallowip=::1/128; IPv6 subnet ban persistence across restart + invalid-subnet RPC error.

test/functional/interface_zmq.py
- Added _zmq_test_ipv6(): subscribes on tcp://[::1]:28340, restarts node with that endpoint, mines a block, verifies hashblock notification is received. Guards with test_ipv6_local(). Added _ZMQ_IPV6_PORT = 28340 constant.

test/functional/rpc_net.py
- _test_getaddednodeinfo: after the IPv4 addnode round-trip, also runs the same with [::1]:p2p_port(2) if IPv6 is available. Both are cleaned up with addnode(..., 'remove').

test/functional/rpc_zmq.py
- address renamed to address_v4; address_v6 = "tcp://[::1]:28342" added. _test_getzmqnotifications runs getzmqnotifications against both endpoints.

test/functional/p2p_disconnect_ban.py
- IPv6 ban block added (guarded by test_ipv6_local()): /129 prefix rejected, ::/24 ban added, ::1 already-banned error verified, subnet unbanned.

test/functional/test_runner.py
- feature_ipv6.py added to the test list.

### DIFF
--- a/src/addrman.h
+++ b/src/addrman.h
@@ -402,6 +402,8 @@ public:
         }
 
         // Deserialize entries from the new table.
+        // All entries (IPv4 and IPv6) are always deserialized unconditionally; nVersion
+        // only controls whether on-disk bucket positions are trusted (v1) or recomputed (v0).
         for (int n = 0; n < nNew; n++) {
             CAddrInfo &info = mapInfo[n];
             s >> info;
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -169,8 +169,11 @@ static bool InitHTTPAllowList()
     CNetAddr localv6;
     LookupHost("127.0.0.1", localv4, false);
     LookupHost("::1", localv6, false);
-    rpc_allow_subnets.push_back(CSubNet(localv4, 8));      // always allow IPv4 local subnet
-    rpc_allow_subnets.push_back(CSubNet(localv6));         // always allow IPv6 localhost
+    // RFC 5735 reserves 127.0.0.0/8 for IPv4 loopback; allow the full range.
+    // RFC 4291 §2.5.3 defines only ::1/128 as the IPv6 loopback — there is no
+    // equivalent loopback /8 in IPv6, so the asymmetry here is intentional.
+    rpc_allow_subnets.push_back(CSubNet(localv4, 8));      // always allow IPv4 local subnet (127.0.0.0/8)
+    rpc_allow_subnets.push_back(CSubNet(localv6));         // always allow IPv6 localhost (::1/128)
     for (const std::string& strAllow : gArgs.GetArgs("-rpcallowip")) {
         CSubNet subnet;
         LookupSubNet(strAllow.c_str(), subnet);
@@ -322,6 +325,9 @@ static bool HTTPBindAddresses(struct evhttp* http)
     // Bind addresses
     for (std::vector<std::pair<std::string, uint16_t> >::iterator i = endpoints.begin(); i != endpoints.end(); ++i) {
         LogPrint(BCLog::HTTP, "Binding RPC on address %s port %i\n", i->first, i->second);
+        // evhttp_bind_socket_with_handle resolves the host string via getaddrinfo
+        // internally. Bare IPv6 addresses (e.g. "::1", "::") without brackets are
+        // handled correctly by libevent ≥ 2.1.x (our minimum is 2.1.12).
         evhttp_bound_socket *bind_handle = evhttp_bind_socket_with_handle(http, i->first.empty() ? nullptr : i->first.c_str(), i->second);
         if (bind_handle) {
             boundSockets.push_back(bind_handle);
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -394,16 +394,16 @@ void SetupServerArgs()
 #endif
     gArgs.AddArg("-txindex", strprintf("Maintain a full transaction index, used by the getrawtransaction rpc call (default: %u)", DEFAULT_TXINDEX), false, OptionsCategory::OPTIONS);
 
-    gArgs.AddArg("-addnode=<ip>", "Add a node to connect to and attempt to keep the connection open (see the `addnode` RPC command help for more info). This option can be specified multiple times to add multiple nodes.", false, OptionsCategory::CONNECTION);
+    gArgs.AddArg("-addnode=<ip[:port]>", "Add a node to connect to and attempt to keep the connection open (see the `addnode` RPC command help for more info). Use [host]:port notation for IPv6 (e.g. [2001:db8::1]:8383). This option can be specified multiple times to add multiple nodes.", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-banscore=<n>", strprintf("Threshold for disconnecting misbehaving peers (default: %u)", DEFAULT_BANSCORE_THRESHOLD), false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-bantime=<n>", strprintf("Number of seconds to keep misbehaving peers from reconnecting (default: %u)", DEFAULT_MISBEHAVING_BANTIME), false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-bind=<addr>", "Bind to given address and always listen on it. Use [host]:port notation for IPv6", false, OptionsCategory::CONNECTION);
-    gArgs.AddArg("-connect=<ip>", "Connect only to the specified node; -noconnect disables automatic connections (the rules for this peer are the same as for -addnode). This option can be specified multiple times to connect to multiple nodes.", false, OptionsCategory::CONNECTION);
+    gArgs.AddArg("-connect=<ip[:port]>", "Connect only to the specified node; -noconnect disables automatic connections (the rules for this peer are the same as for -addnode). Use [host]:port notation for IPv6 (e.g. [2001:db8::1]:8383). This option can be specified multiple times to connect to multiple nodes.", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-discover", "Discover own IP addresses (default: 1 when listening and no -externalip or -proxy)", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-dns", strprintf("Allow DNS lookups for -addnode, -seednode and -connect (default: %u)", DEFAULT_NAME_LOOKUP), false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-dnsseed", "Query for peer addresses via DNS lookup, if low on addresses (default: 1 unless -connect used)", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-addseeder=<ip>", "IP address of the dns seeder server to use", false, OptionsCategory::CONNECTION);
-    gArgs.AddArg("-externalip=<ip>", "Specify your own public address", false, OptionsCategory::CONNECTION);
+    gArgs.AddArg("-externalip=<ip>", "Specify your own public address. Use bracket notation for IPv6 (e.g. [2001:db8::1])", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-forcednsseed", strprintf("Always query for peer addresses via DNS lookup (default: %u)", DEFAULT_FORCEDNSSEED), false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-listen", "Accept connections from outside (default: 1 if no -proxy or -connect)", false, OptionsCategory::CONNECTION);
     gArgs.AddArg("-listenonion", strprintf("Automatically create Tor hidden service (default: %d)", DEFAULT_LISTEN_ONION), false, OptionsCategory::CONNECTION);
@@ -431,16 +431,16 @@ void SetupServerArgs()
     hidden_args.emplace_back("-upnp");
 #endif
     gArgs.AddArg("-whitebind=<addr>", "Bind to given address and whitelist peers connecting to it. Use [host]:port notation for IPv6", false, OptionsCategory::CONNECTION);
-    gArgs.AddArg("-whitelist=<IP address or network>", "Whitelist peers connecting from the given IP address (e.g. 1.2.3.4) or CIDR notated network (e.g. 1.2.3.0/24). Can be specified multiple times."
+    gArgs.AddArg("-whitelist=<IP address or network>", "Whitelist peers connecting from the given IP address (e.g. 1.2.3.4 or 2001:db8::1) or CIDR notated network (e.g. 1.2.3.0/24 or 2001:db8::/32). Can be specified multiple times."
         " Whitelisted peers cannot be DoS banned and their transactions are always relayed, even if they are already in the mempool, useful e.g. for a gateway", false, OptionsCategory::CONNECTION);
 
     g_wallet_init_interface.AddWalletOptions();
 
 #if ENABLE_ZMQ
-    gArgs.AddArg("-zmqpubhashblock=<address>", "Enable publish hash block in <address>", false, OptionsCategory::ZMQ);
-    gArgs.AddArg("-zmqpubhashtx=<address>", "Enable publish hash transaction in <address>", false, OptionsCategory::ZMQ);
-    gArgs.AddArg("-zmqpubrawblock=<address>", "Enable publish raw block in <address>", false, OptionsCategory::ZMQ);
-    gArgs.AddArg("-zmqpubrawtx=<address>", "Enable publish raw transaction in <address>", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubhashblock=<address>", "Enable publish hash block in <address> (e.g. tcp://127.0.0.1:28332 or tcp://[::1]:28332)", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubhashtx=<address>", "Enable publish hash transaction in <address> (e.g. tcp://127.0.0.1:28333 or tcp://[::1]:28333)", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubrawblock=<address>", "Enable publish raw block in <address> (e.g. tcp://127.0.0.1:28334 or tcp://[::1]:28334)", false, OptionsCategory::ZMQ);
+    gArgs.AddArg("-zmqpubrawtx=<address>", "Enable publish raw transaction in <address> (e.g. tcp://127.0.0.1:28335 or tcp://[::1]:28335)", false, OptionsCategory::ZMQ);
 #else
     hidden_args.emplace_back("-zmqpubhashblock=<address>");
     hidden_args.emplace_back("-zmqpubhashtx=<address>");
@@ -499,7 +499,7 @@ void SetupServerArgs()
     gArgs.AddArg("-blockmaxsize=<n>", strprintf("Set maximum block size in bytes (default: %d)", DEFAULT_BLOCK_MAX_SIZE), false, OptionsCategory::BLOCK_CREATION);
 
     gArgs.AddArg("-rest", strprintf("Accept public REST requests (default: %u)", DEFAULT_REST_ENABLE), false, OptionsCategory::RPC);
-    gArgs.AddArg("-rpcallowip=<ip>", "Allow JSON-RPC connections from specified source. Valid for <ip> are a single IP (e.g. 1.2.3.4), a network/netmask (e.g. 1.2.3.4/255.255.255.0) or a network/CIDR (e.g. 1.2.3.4/24). This option can be specified multiple times", false, OptionsCategory::RPC);
+    gArgs.AddArg("-rpcallowip=<ip>", "Allow JSON-RPC connections from specified source. Valid for <ip> are a single IP (e.g. 1.2.3.4 or 2001:db8::1), a network/netmask (e.g. 1.2.3.4/255.255.255.0) or a network/CIDR (e.g. 1.2.3.4/24 or 2001:db8::/32). This option can be specified multiple times", false, OptionsCategory::RPC);
     gArgs.AddArg("-rpcauth=<userpw>", "Username and hashed password for JSON-RPC connections. The field <userpw> comes in the format: <USERNAME>:<SALT>$<HASH>. A canonical python script is included in share/rpcauth. The client then connects normally using the rpcuser=<USERNAME>/rpcpassword=<PASSWORD> pair of arguments. This option can be specified multiple times", false, OptionsCategory::RPC);
     gArgs.AddArg("-rpcbind=<addr>[:port]", "Bind to given address to listen for JSON-RPC connections. This option is ignored unless -rpcallowip is also passed. Port is optional and overrides -rpcport. Use [host]:port notation for IPv6. This option can be specified multiple times (default: 127.0.0.1 and ::1 i.e., localhost, or if -rpcallowip has been specified, 0.0.0.0 and :: i.e., all addresses)", false, OptionsCategory::RPC);
     gArgs.AddArg("-rpccookiefile=<loc>", "Location of the auth cookie. Relative paths will be prefixed by a net-specific datadir location. (default: data dir)", false, OptionsCategory::RPC);
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -210,8 +210,11 @@ void AdvertiseLocal(CNode *pnode)
 // learn a new local address
 bool AddLocal(const CService& addr, int nScore)
 {
-    if (!addr.IsRoutable())
+    if (!addr.IsRoutable()) {
+        if (nScore >= LOCAL_MANUAL)
+            LogPrintf("AddLocal: ignoring non-routable address %s passed via -externalip\n", addr.ToString());
         return false;
+    }
 
     if (!fDiscover && nScore < LOCAL_MANUAL)
         return false;
@@ -1484,6 +1487,10 @@ static CThreadInterrupt g_upnp_interrupt;
 static std::thread g_upnp_thread;
 static void ThreadMapPort()
 {
+    if (!IsReachable(NET_IPV4)) {
+        LogPrintf("UPnP: skipping port mapping — UPnP IGD only discovers IPv4 external addresses and the node has no IPv4 reachability\n");
+        return;
+    }
     std::string port = strprintf("%u", GetListenPort());
     const char * multicastif = nullptr;
     const char * minissdpdpath = nullptr;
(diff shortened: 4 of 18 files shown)


azuchi (changes requested):
The IPv6 core work is solid — `IsLinkLocal()` correctly implements RFC 4291 §2.5.6 (fe80::/10, bit-level check), `IsRoutable()` picks it up so fe81::–febf:: no longer pass as routable, `ZMQ_IPV6 = 1` before `zmq_bind` enables IPv6 publisher endpoints on all libzmq builds, and the `CreateSocket()`/`AddLocal()` logging fixes remove real observability gaps. Documentation updates and test coverage across `feature_ipv6.py`, `p2p_disconnect_ban.py`, `rpc_net.py`, `rpc_zmq.py`, `rpc_bind.py`, and `netutil.py` are thorough.

One CI failure (real, reproducible), one policy regression that pulls against the PR's IPv6 direction, and a few minor tidy-ups below.

### 🚨 B1 — Blocker: `interface_zmq.py::_zmq_test_ipv6` is racy, failing CI

Fast-test `Linux x86_64` and `Linux ARM64` both fail here ([job log](https://github.com/chaintope/tapyrus-core/actions/runs/28593622890/job/84783299370)):

```
interface_zmq.py failed, Duration: 63 s
File "test/functional/interface_zmq.py", line 147, in _zmq_test_ipv6
    recv_hash = bytes_to_hex_str(hashblock6.receive())
File "test/functional/interface_zmq.py", line 33, in receive
    topic, body, seq = self.socket.recv_multipart()
zmq.error.Again: Resource temporarily unavailable
```

63 s = the RCVTIMEO firing at 60 s — the subscriber never gets the hashblock message. Not a code bug: the C++ side (`ZMQ_IPV6 = 1` + `zmq_bind`) works. The test flow is:

```python
socket6.connect(ipv6_addr)              # SUB connect (async, no publisher yet)
hashblock6 = ZMQSubscriber(socket6, b"hashblock")   # SUBSCRIBE filter set
self.restart_node(0, extra_args=["-zmqpubhashblock=" + ipv6_addr])
genhashes = self.nodes[0].generate(1, ...)          # ONE block, immediately
recv_hash = hashblock6.receive()                    # timeouts here
```

ZMQ PUB drops messages sent before the SUB's filter has propagated. `restart_node` returns as soon as RPC comes up, but that's not the same instant the ZMQ publisher has finished its `zmq_bind` and accepted the subscriber's TCP handshake + registered its SUBSCRIBE frame. Publishing one and only one block right after restart means every dropped notification is a test failure.

`_zmq_test` (the IPv4 variant on the same file) avoids this by generating `num_blocks = 5` — even if the first message is lost during handshake, later ones arrive.

**Fix**: any of

1. `self.nodes[0].generate(5, self.signblockprivkey_wif)` and consume only the last hash (or run the ZMQSubscriber loop 5 times), matching `_zmq_test`.
2. Wrap `hashblock6.receive()` in `wait_until` and re-issue `generate(1, ...)` until one hash arrives.
3. Explicit sync: poll `getzmqnotifications()` for the new endpoint after restart, sleep briefly, then generate.

(1) is the shape the rest of the file already uses and is the smallest diff.

### 🟡 N1 — `InitBinds` change makes IPv6-only hosts trip a user-visible error

```cpp
- fBound |= Bind(CService(inaddr_any, GetListenPort()), !fBound ? BF_REPORT_ERROR : BF_NONE);
+ fBound |= Bind(CService(inaddr_any, GetListenPort()), BF_REPORT_ERROR);
```

The old semantics were "at least one of {IPv6, IPv4} must succeed; if IPv6 already bound, IPv4 failure is silent". The new semantics are "IPv4 must succeed unconditionally". `BF_REPORT_ERROR` calls `clientInterface->ThreadSafeMessageBox(strError, "", MSG_ERROR)` — a user-visible modal error, on top of the log line.

On the common dual-stack host this is a no-op (both binds succeed). But on a genuine IPv6-only host — the environment this PR is nominally opening the door to — IPv6 bind succeeds, IPv4 bind fails (nothing listens on `0.0.0.0`, or the port collides on a v4 side you don't care about), and the user sees `Cannot bind …` in a message box, despite the node running fine on `::`. That's an inversion of the "at least one must succeed" invariant.

The comment in the diff explains the design intent, but the intent ("IPv4 is assumed always present") is at odds with the rest of the PR. Suggestion: keep the old "at least one succeeded" gating (or track `v6_bound`/`v4_bound` separately and only report if both failed), so IPv6-only deployments don't get an error dialog on every start-up.

### 🟡 N2 — No unit-test coverage for the IsLinkLocal /10 extension

The bug being fixed is exactly that `fe81::`–`febf::` used to slip past `IsRFC4862()` (which checked /64) as routable. The functional test `_test_externalip_linklocal_rejected` covers `fe80::1`, but the *actual* regression window this fix closes is the extra 6 bits — `fe81::` and up.

A minimal addition to `src/test/netbase_tests.cpp` catches any future regression at the boundary:

```cpp
BOOST_CHECK(ResolveIP("fe80::").IsLinkLocal());
BOOST_CHECK(ResolveIP("fe81::").IsLinkLocal());   // was silently routable before
BOOST_CHECK(ResolveIP("febf::").IsLinkLocal());   // top of /10
BOOST_CHECK(!ResolveIP("fec0::").IsLinkLocal());  // below /10 (formerly site-local)
BOOST_CHECK(!ResolveIP("fe81::").IsRoutable());   // routability guard
```

### 🟡 N3 — `_ROUTABLE_IPV6` comment claim about `2001:db8::/32` is inaccurate

`feature_ipv6.py:15` says `2001:db8::/32` "fails IsValid()". `CNetAddr::IsValid()` doesn't specifically reject the RFC 3849 documentation prefix — the address parses and is considered valid; `IsRoutable()` is what excludes some non-routable IPv6 ranges, but not RFC 3849 either (that block *is* topologically routable, it's just reserved for docs). The chosen alternative `2603:3005::1` is fine, so the practical outcome is correct; only the rationale in the comment is off.

Suggestion: reword to "avoid RFC 3849 documentation prefix so this doesn't look like a copy-pasted example" or similar.

### 🟢 Notes for later, non-blocking

- `AddLocal()` warning fires only for `nScore >= LOCAL_MANUAL`, which is the right cut (auto-discovered non-routables shouldn't spam the log). Good.
- `ThreadMapPort` early-return is accurate for the current miniupnpc IGDv1 usage. If IGDv2 IPv6 UPnP is ever wired up here, remember to revisit.
- `_zmq_test_ipv6` restarts node 0 with only the IPv6 `hashblock` publisher — the four IPv4 subscribers created in `setup_nodes` are left dangling on the destroyed publisher. Not a correctness issue (teardown handles it), just noted.

### Summary

Fix B1 (the ZMQ subscriber race) and this is ready — the core changes are correct and well-motivated. N1 would be worth addressing before merge given it works against the PR's own goal; N2 and N3 are cleanups.

azuchi (changes requested):
N1, N2, N3 are all addressed cleanly — thanks. Reopening on B1: the fix as applied doesn't close the CI failure, and I want to correct my own previous diagnosis of the root cause. My earlier note said "the C++ side (`ZMQ_IPV6 = 1` + `zmq_bind`) works, this is purely a slow-joiner race on the Python side" — that missed the actual failure mode, sorry.

### 🚨 B1 — CI still fails at `interface_zmq.py::_zmq_test_ipv6` (same line, same error)

`Linux x86_64` and `Linux ARM64` in the latest run ([job log](https://github.com/chaintope/tapyrus-core/actions/runs/28662381636/job/85005921357)) both fail with the same `zmq.error.Again` inside `socket6.recv_multipart()` after RCVTIMEO (60 s):

```
interface_zmq.py failed, Duration: 66 s
File "test/functional/interface_zmq.py", line 165, in _zmq_test_ipv6
    topic, body, seq = socket6.recv_multipart()
zmq.error.Again: Resource temporarily unavailable
```

Reading the combined-log timeline for that node — this is where my previous analysis was wrong — the publisher is working fine, and the "slow joiner over 5 blocks" fix isn't what's needed:

| Time (s) | Event |
|---|---|
| 13:21:51.506 | Python SUB calls `socket6.connect("tcp://[::1]:28340")` |
| 13:21:51.838 | node0 publisher: `Notifier pubhashblock ready (address = tcp://[::1]:28340)` |
| 13:21:51.960 | `generate(5)` RPC hits node0 |
| 13:21:51.961 – .964 | node0 actually publishes 5 hashblock notifications on `[::1]:28340` |
| 13:22:52 | SUB `recv_multipart()` raises `zmq.error.Again` (RCVTIMEO fired) |

If this were a slow-joiner race, generating 5 blocks would give the SUB multiple attempts to catch up. But the publisher emits all 5 within ~4 ms, and then the SUB sits idle for a full 60 s and receives nothing. The 5-message spacing is essentially instantaneous — batching more messages into the same instantaneous window doesn't change the outcome.

The real cause is one line short on the Python side of the test:

**The SUB socket is IPv4-only.** libzmq's `ZMQ_IPV6` defaults to `0` (IPv4 only). When you call `zmq_connect(sock, "tcp://[::1]:28340")` on such a socket, libzmq parses the endpoint, tries to resolve `::1` with `AF_INET`, `inet_pton` fails, the tcp_connecter enters its reconnect loop, and it retries forever without ever establishing a connection. No exception is raised — `zmq_connect` is asynchronous, so from the Python side the call succeeds and `recv` just quietly times out.

This is the exact symmetric of the C++ change this PR adds to `src/zmq/zmqpublishnotifier.cpp:81`:

```cpp
int ipv6 = 1;
zmq_setsockopt(psocket, ZMQ_IPV6, &ipv6, sizeof(ipv6));
zmq_bind(psocket, address.c_str());
```

The publisher does the right thing, but the test's SUB has been missing the corresponding `zmq.IPV6=1` from the start. That's why:

- publisher log shows the binds and publishes succeeding,
- SUB gets zero messages regardless of how many are published,
- the timeout is a deterministic 60 s on every runner (`fast-test Linux x86_64`, `fast-test Linux ARM64`), not the flaky race pattern I originally described.

**Fix — one line, right where the SUB is created:**

```python
socket6 = self.zmq_context.socket(zmq.SUB)
socket6.set(zmq.IPV6, 1)          # <-- add this before .connect() on an IPv6 endpoint
socket6.set(zmq.RCVTIMEO, 60000)
socket6.connect(ipv6_addr)
socket6.setsockopt(zmq.SUBSCRIBE, b"hashblock")
```

Once that's in place, the true slow-joiner window (SUB TCP handshake + SUBSCRIBE frame vs. the first `generate`) will still exist, but it becomes the small, sporadic race I originally described rather than an unconditional failure. The current 5-block generate + last-hash match approach is a reasonable mitigation for that; you can keep it. A stronger alternative if it ever proves flaky is to poll `generate(1)` inside a `wait_until` loop that stops when a message arrives — but I wouldn't do that pre-emptively.

I also want to flag the mistake in my previous review explicitly: I said the failure was "a SUB filter propagation race, not a C++ code bug" and pointed at generating multiple blocks as the fix. That was misdirected — the timeline in the CI log doesn't fit slow-joiner (60 s with a hot connection is far more than enough for SUBSCRIBE propagation), and I didn't check that the Python SUB even had IPv6 enabled. Apologies for the wasted round-trip.

### 🟢 N1 — InitBinds semantics restored

```cpp
fBound |= Bind(CService(inaddr6_any, ...), BF_NONE);
fBound |= Bind(CService(inaddr_any, ...),  !fBound ? BF_REPORT_ERROR : BF_NONE);
```

Back to the original "at least one of {IPv6, IPv4} must succeed; IPv4 failure only reports if IPv6 hasn't already bound." Comment updated. Good — IPv6-only hosts no longer trip a spurious `Cannot bind` dialog.

### 🟢 N2 — `netbase_islinklocal_boundary` added

Coverage is stronger than I asked for: the /10 lower boundary (`fe7f::`), an interior point (`fea0::`), the realistic host-bits case (`fe80::1`), and adjacent-prefix negatives (`fc00::` ULA, `ff02::1` multicast — separate mechanism, easy to conflate). Good addition.

### 🟢 N3 — `_ROUTABLE_IPV6` rationale corrected

New comment accurately describes RFC 3849 (routable, reserved for documentation) rather than claiming `IsValid()` rejects it. Good.

### Summary

Only B1 remains, and the specific fix is one `setsockopt(zmq.IPV6, 1)` on the SUB socket. Once that's in and CI is green, this is ready.

azuchi (changes requested):
Thanks for adding `socket6.set(zmq.IPV6, 1)` — that was necessary for the SUB to attempt an IPv6 connect in the first place. But CI is still failing, deterministically, on both `Linux x86_64` and `Linux ARM64` in [the latest run](https://github.com/chaintope/tapyrus-core/actions/runs/28776314597). Same line (`interface_zmq.py:166`), same `zmq.error.Again` after RCVTIMEO. I reproduced the failure locally and I now understand what's actually going on — closer to my very first diagnosis (slow joiner) than my second (missing IPv6), but with a wrinkle that makes the current mitigation ineffective.

### 🚨 B1 — Still failing: the "5 blocks in one burst" mitigation doesn't decouple `generate()` from SUBSCRIBE propagation

Timeline for `Linux x86_64` (`combined logs`):

| Time (s) | Event |
|---|---|
| 08:11:28.757 | Test: `socket6.set(IPV6,1); set(RCVTIMEO); connect([::1]:28340); SUBSCRIBE(hashblock)` — publisher is not up yet, libzmq starts its 100 ms reconnect loop |
| 08:11:28.954 | node0 shuts down the old IPv4 publisher (start of `restart_node`) |
| 08:11:29.087 | new node0: `Notifier pubhashblock ready (address = tcp://[::1]:28340)` — publisher is now listening |
| 08:11:29.212 – .216 | node0 publishes 5 hashblock messages on `[::1]:28340`, all 5 within 4 ms |
| 08:12:29.47 | SUB `recv_multipart()` raises `zmq.error.Again` (RCVTIMEO fired) |

The C++ side is doing everything right — `ZMQ_IPV6=1` is set before `zmq_bind` (`src/zmq/zmqpublishnotifier.cpp:81`), the bind succeeds, and 5 hashblock messages are sent. But the SUB receives zero of them.

### Why the mitigation doesn't work

The current test relies on the classic slow-joiner countermeasure: publish `num_blocks = 5` messages and hope that at least one arrives after the SUB's SUBSCRIBE has propagated to the publisher's filter table:

```python
num_blocks = 5
genhashes = self.nodes[0].generate(num_blocks, self.signblockprivkey_wif)

recv_hash = None
for _ in range(num_blocks):
    topic, body, seq = socket6.recv_multipart()
    ...
    if recv_hash == genhashes[-1]:
        break
```

That assumption holds for real chain progression, where blocks are spaced hundreds of milliseconds apart and the SUB has time to catch up. It doesn't hold here: `generate(5)` on a signet-style dev chain emits all 5 messages in a single ~4 ms burst (see 29.212 – 29.216 in the log). So either the SUBSCRIBE has propagated to the publisher's filter table before that 4 ms window (all 5 delivered) or it hasn't (all 5 dropped by the PUB, because a PUB with no matching subscription silently discards messages). There's essentially no interior state where "3 of 5 arrive".

On the GH runners the second case wins deterministically, which is why the failure looks identical on x86_64 and ARM64 and repeats round after round.

### Confirmed locally

I built a minimal reproducer that mirrors the exact timing (SUB starts, wait ~330 ms, PUB binds and after 125 ms sends 5 messages within a few ms). With `ZMQ_IPV6=1` on both sides and everything else nominal, the SUB receives 0 messages and times out. Replacing the "generate 5 in a burst" pattern with a "generate one → try recv → repeat" loop reliably converges in 1–2 iterations (~1 s wall-clock).

### Fix — decouple `generate()` from subscription propagation

Keep generating one block at a time and try to receive between calls. Each miss simply means "SUBSCRIBE hasn't landed on the publisher yet, publish another one and try again":

```python
socket6.set(zmq.IPV6, 1)
socket6.set(zmq.RCVTIMEO, 1000)   # short per-attempt, not 60 s
socket6.connect(ipv6_addr)
socket6.setsockopt(zmq.SUBSCRIBE, b"hashblock")

self.restart_node(0, extra_args=["-zmqpubhashblock={}".format(ipv6_addr)])

recv_hash = None
for _ in range(30):  # 30 * ~1 s worst case, well under the surrounding harness limits
    target = self.nodes[0].generate(1, self.signblockprivkey_wif)[0]
    try:
        topic, body, _ = socket6.recv_multipart()
        assert_equal(topic, b"hashblock")
        recv_hash = bytes_to_hex_str(body)
        # Any hashblock arriving means the SUBSCRIBE is now registered.
        # It may correspond to an earlier iteration if we're catching
        # up, so we don't require recv_hash == target on the first hit.
        break
    except zmq.Again:
        continue

assert recv_hash is not None, "no hashblock received on tcp://[::1]:{}".format(_ZMQ_IPV6_PORT)
```

On a healthy runner this converges in 1–2 attempts and terminates in ~1 s, so the total wall-clock is well below the previous 60 s hard timeout even in the worst case.

If you'd rather not restructure the loop, a `time.sleep(0.5)` after `restart_node` and before `generate(5)` would also work (it's what a few upstream Bitcoin functional tests do), but the loop-generate form is more honest about what's actually being defended against and doesn't hide the slow-joiner behind a magic constant.

### N1 / N2 / N3

Still 🟢 — no regressions from the previous round.

### Sorry — this is round three of me being wrong

For transparency about how much to trust my ZMQ diagnoses:

- **Round 1:** Called this a slow-joiner race and recommended "generate multiple blocks + match on last hash". The mechanism call was right, the mitigation call was wrong — I didn't check whether `generate(N)` staggers messages (it doesn't) or emits them in a single burst (it does).
- **Round 2:** Recanted round 1 and blamed missing `zmq.IPV6=1` on the SUB. That fix WAS necessary — without it the SUB couldn't even attempt IPv6 — but it wasn't sufficient. The CI timeline I quoted actually showed the publisher succeeding, which I should have taken as evidence that connectivity wasn't the whole story.
- **Round 3 (this one):** Both fixes needed. IPv6=1 is in. The slow-joiner mitigation needs to be a loop over `generate(1)`, not a batch `generate(5)`.

I verified round 3 with a local reproducer before writing this. Apologies for the wasted round-trips.

Once the loop-generate (or an equivalent fix that gives the SUBSCRIBE frame time to propagate before the messages start flowing) is in and CI is green, this is ready.

azuchi (changes requested):
### 🟢 B1 resolved — `interface_zmq.py::_zmq_test_ipv6` now passes

The loop-generate rewrite is exactly the shape I was asking for: `RCVTIMEO=1000`, one block per iteration, retry on `zmq.Again`, break on first arrival, assert `recv_hash is not None`. CI confirms `interface_zmq.py | ✓ Passed | 4 s` on both `Linux x86_64` and `Linux ARM64` in [run 28868424964](https://github.com/chaintope/tapyrus-core/actions/runs/28868424964). The updated comment block that captures why `generate(N)` in a burst doesn't work here is also a nice touch — future readers won't have to rediscover this.

### 🚨 B2 — `feature_ipv6.py::_test_p2p_via_ipv6` fails on both Linux runners

Now that `interface_zmq.py` passes, `--failfast` reaches the next test in `BASE_SCRIPTS`, and `feature_ipv6.py` fails at `feature_ipv6.py:62`:

```
AssertionError: Expected IPv6 peer addr, got: 127.0.0.1:11529
```

The same message on both `Linux x86_64` and `Linux ARM64`, both after 1 s. macOS jobs pass because `test_ipv6_local()` returns False on the runner and the whole test early-returns — so this is Linux-only and hasn't been visible in earlier CI runs.

The cause is in the test's assumption about the topology at `run_test` time, not in any of the C++ changes:

- `BitcoinTestFramework.setup_network()` (`test_framework/test_framework.py:263`) calls `connect_nodes_bi(self.nodes, 0, 1)` after `setup_nodes()`. `connect_nodes_bi` uses `127.0.0.1:<port>`.
- `IPv6Test` inherits the default `setup_network` and doesn't override it, so by the time `_test_p2p_via_ipv6` runs there is already one IPv4 peer on `node0`.
- `_test_p2p_via_ipv6` then does:
  ```python
  self.nodes[0].addnode("[::1]:{}".format(p2p_port(1)), "onetry")
  wait_until(lambda: len(self.nodes[0].getpeerinfo()) > 0, timeout=15)
  peer_addr = self.nodes[0].getpeerinfo()[0]["addr"]
  ```
  `wait_until` returns immediately because the IPv4 peer is already present, and `getpeerinfo()[0]` returns that IPv4 peer, not the [::1] one we just added.

A secondary issue in the same sub-test: the subsequent block-propagation check

```python
self.nodes[0].generate(1, self.signblockprivkey_wif)
wait_until(lambda: self.nodes[1].getblockcount() == 1, timeout=15)
```

does not actually verify that the block travelled over the IPv6 connection — it could equally well have travelled over the pre-existing IPv4 one. So even if we only patched the assertion, the "P2P via [::1]" claim in the log message would still be under-verified.

### Suggested fix

Skip the auto-connect for this test so each sub-test sets up its own topology explicitly. That way `_test_p2p_via_ipv6` is the sole owner of the connection it's testing:

```python
class IPv6Test(BitcoinTestFramework):
    def set_test_params(self):
        self.num_nodes = 2
        self.setup_clean_chain = True
        self.bind_to_localhost_only = False

    def setup_network(self):
        # Do NOT auto-connect via 127.0.0.1 — each sub-test sets up its own
        # IPv4/IPv6 topology, and _test_p2p_via_ipv6 in particular needs the
        # only peer on node0 to be the one it establishes over [::1].
        self.setup_nodes()
```

Then in `_test_p2p_via_ipv6`, waiting for `getpeerinfo() > 0` is already correct — the first (and only) peer will be the IPv6 one:

```python
def _test_p2p_via_ipv6(self):
    self.log.info("P2P: connect two nodes via [::1]")
    self.nodes[0].addnode("[::1]:{}".format(p2p_port(1)), "onetry")
    wait_until(lambda: len(self.nodes[0].getpeerinfo()) > 0, timeout=15)
    peer_addr = self.nodes[0].getpeerinfo()[0]["addr"]
    assert "::" in peer_addr, "Expected IPv6 peer addr, got: {}".format(peer_addr)
    self.nodes[0].generate(1, self.signblockprivkey_wif)
    wait_until(lambda: self.nodes[1].getblockcount() == 1, timeout=15)
    self.log.info("  block propagated over IPv6 P2P connection")
```

The other sub-tests are unaffected: `_test_addnode_ipv6` explicitly adds peers itself; `_test_connect_flag_ipv6`/`_test_externalip_*`/`_test_whitelist_*`/`_test_rpcallowip_*`/`_test_ban_ipv6_persistence` all restart node1 with their own args and reset its peer state.

If you'd rather keep the default topology, the minimal alternative is to look up the IPv6 peer explicitly rather than trusting the order:

```python
addr = "[::1]:{}".format(p2p_port(1))
self.nodes[0].addnode(addr, "onetry")
wait_until(lambda: any("::" in p["addr"] for p in self.nodes[0].getpeerinfo()), timeout=15)
```

but that still leaves the block-propagation check ambiguous about which path was used, so I'd lean toward the `setup_network` override.

### Summary

B1 (interface_zmq) is fully resolved. B2 (feature_ipv6) is a small, self-contained test bug — a `setup_network` override or an "any peer is IPv6" filter closes it. Once CI is green on both Linux runners, this is ready.

azuchi (approved):

PR #438, opened 2026-07-03
Edit weekly heavy test CI

Naviabheeman:
- Fix feature_pruning test by making the transaction standard (chunk the OP_RETURN payload into <=520-byte pushes, since Tapyrus's CScript::HasValidOps() rejects larger single push elements at tx-decode time, independent of mempool policy).
- Add a --test-timeout override to test_runner.py, independent of the other functional tests' 30-minute default, so feature_pruning can run for the hours it actually needs.
- Add a new Docker testnet sync job: builds the tapyrusd image and syncs a single node against the live public Chaintope testnet.

EDIT:
Add feature_dbcrash test to this test as that test is a long running test. In the daily CI it always failed after 30 minutes due to its long runtime.


azuchi (commented):
Overall this is a well-scoped fix + useful CI additions. The `blocktools.py` OP_RETURN chunking is correctly diagnosed and correctly implemented — I verified the arithmetic and encoding locally. The `--test-timeout` / `--heavy` additions to `test_runner.py` look clean. The new Docker testnet-sync job is a good idea but has a few shell-robustness gaps I'd like to see tightened before merge. Nothing here is a hard blocker.

### 🟢 G1 — `blocktools.py` OP_RETURN chunking

The diagnosis is right: `src/core_read.cpp:105` calls `CheckTxScriptsSanity` from every `DecodeHexTx`, which iterates `HasValidOps()` on each script — and `src/script/script.cpp:341` rejects any push element bigger than `MAX_SCRIPT_ELEMENT_SIZE` (520). A single 9990-byte OP_RETURN push is decoded as one PUSHDATA2 item of 9990 bytes; that fails the check before mempool policy ever runs. Chunking into ≤520-byte pushes fixes it because each push is now individually accepted, and the whole script (`OP_RETURN` + 19 × PUSHDATA2(520 B)) stays under `MAX_SCRIPT_SIZE`.

Verified locally that:

- Old script: `len(script)=9994`, single opcode `0x4d` (`OP_PUSHDATA2`) with `data_size=9990` — rejected.
- New script: `len(script)=9938`, `OP_RETURN` followed by 19 × `0x4d` with `data_size=520` — accepted.

Arithmetic check: `(MAX_SCRIPT_SIZE - 1) // (MAX_SCRIPT_ELEMENT_SIZE + 3) = 9999 // 523 = 19`; `1 + 19 * 523 = 9938 ≤ 10000` ✓. The `+= vs [...]` comment about `CScript.__coerce_instance` re-wrapping already-encoded bytes is accurate (`test/functional/test_framework/script.py:406-424`) and worth keeping — that's a genuine footgun.

Minor observation, not a change request: the `chunk_overhead = 3` (`OP_PUSHDATA2` = 1 byte opcode + 2 byte length) is only correct for pushes in the `[256, 65535]` range. If `MAX_SCRIPT_ELEMENT_SIZE` ever changed to a value below 256, `OP_PUSHDATA1` (1 byte length) would apply and the overhead would be 2. Given how load-bearing 520 is in the codebase this is a very theoretical concern, but you could either write `chunk_overhead = 3 if MAX_SCRIPT_ELEMENT_SIZE > 255 else 2` or add a one-line note that the number assumes PUSHDATA2. Optional.

**Note (informational):** `create_tx_with_large_script` has two other callers, and it's worth being explicit that neither is a covert regression:

- `feature_reindex.py:127` also calls `signrawtransactionwithwallet(tx_raw)`, so it would hit the same rejection — but that call site sits inside `test_longchain_reindex()` which is gated by `self.options.longchain` (see `feature_reindex.py:252`) and is not run by the daily suite. This PR's framework-level fix quietly re-enables that path for anyone running `--longchain` locally. Worth a line in the PR description.
- `feature_largeblocksize.py:103` submits via `node.p2p.send_txs_and_test(...)`, which deserialises through the P2P codec — `HasValidOps`/`CheckTxScriptsSanity` are only reachable from `DecodeHexTx`, not the P2P path. That's why this test has been quietly passing daily CI on the old script (single 9990 B push, standard `TX_NULL_DATA` under `-datacarriersize=10000`). After this PR the multi-push variant still classifies as `TX_NULL_DATA` (see `standard.cpp:184-186` — `OP_RETURN` + `IsPushOnly` matches regardless of chunk count) and is still ≤ `-datacarriersize=10000`, so no regression here either.

So the fix is correctly scoped at the framework level and lands with zero collateral risk on the two other callers.

### 🟢 G2 — `test_runner.py` `--test-timeout` / `--heavy`

Threading `test_timeout` through `run_tests` → `TestHandler.__init__` → `get_next` and replacing the module-level `TOTAL_TEST_DURATION` reference in the loop is the right shape. `--heavy` correctly plugs into the existing `HEAVY_SCRIPTS` list (added in #417) and slots into the `if tests / elif args.heavy / elif args.extended / else` chain without disturbing other paths. Retry path (`run_tests(..., retry=True)`) also receives the override, so a long test doesn't collapse back to 1800 s on its second try.

One small suggestion: since `HEAVY_SCRIPTS` today only holds `feature_pruning.py`, a future contributor adding an entry there will unknowingly opt into the weekly CI. A one-line comment on the list — "tests added here are exercised only by the weekly-heavy-tests workflow (`--heavy` in `test_runner.py`)" — would make that dependency explicit.

### ⚠️ N1 — `docker-testnet-sync`: shell robustness

The intent of this job is clear and I like the idea of a live-testnet canary. A few things to tighten:

**Empty-value hazard on numeric checks.** `cli getconnectioncount` / `cli getblockcount` return an empty string if the RPC fails (container not reachable, tapyrusd mid-restart, etc.). The subsequent `[ "$connections" -lt 1 ]` then evaluates `[ "" -lt 1 ]`, which under `bash`'s `[` errors with `integer expression expected` and aborts the step with a confusing message — the run fails, but not with the intended `FAILED: ...` line. Two ways to fix:

```bash
connections=$(cli getconnectioncount 2>/dev/null || echo 0)
height_prev=$(cli getblockcount 2>/dev/null || echo 0)
```

or explicitly:

```bash
if [ "${connections:-0}" -lt 1 ]; then
```

Same applies to the `height_now`/`height_prev` comparisons inside the checkpoint loop.

**Fragile JSON matching.** `grep -q '"initialblockdownload": false'` assumes tapyrus-cli's pretty-printed JSON format with a space after the colon. That's the current output, but coupling a CI health check to a format assumption is the kind of thing that quietly breaks and is hard to notice. If `jq` is acceptable as a runner-level dep (`ubuntu-latest` has it out of the box), consider:

```bash
ibd_complete() {
    [ "$(cli getblockchaininfo | jq -r .initialblockdownload)" = "false" ]
}
```

If you'd rather not add `jq`, keeping `grep -q` is fine but add a comment noting the format assumption.

**Inline genesis hex is a long-term hazard.** The 384-byte hex string in the workflow is authoritative in exactly one place — this workflow file. If we ever regenerate testnet (unlikely in the current cadence, but not impossible during a future fork test) the job silently syncs against a network that no longer matches. Either extract it into `contrib/docker/testnet-genesis-block-with-sig` (or similar) and load with `-e GENESIS_BLOCK_WITH_SIG=$(cat contrib/docker/testnet-genesis-block-with-sig)`, or at minimum add a comment linking to where the genesis is documented (looks like `doc/tapyrus/getting_started.md` per the workflow comment already).

**Long unconditional `sleep 1800` masks fast-fail conditions.** If tapyrusd crashes on startup (bad config, missing key, Docker mount issue), we discover it 30 minutes later. A cheap early check pays for itself:

```bash
sleep 30
if ! docker ps --format '{{.Names}}' | grep -q '^tapyrus_testnet_sync$'; then
    docker logs tapyrus_testnet_sync
    echo "FAILED: container exited within 30 seconds of startup"
    exit 1
fi
```

**Nit — workflow file rename.** `.github/workflows/heavy-functional-tests.yml` → `weekly-heavy-tests.yml` is fine for the schedule (GitHub re-registers cron on push), but any README badge, PR template link, or doc reference pointing at the old name will 404. Quick grep to be sure:

```
git grep -n 'heavy-functional-tests'
```

If nothing matches, ignore this.

### Summary

`blocktools.py` and `test_runner.py` changes look good to me — the OP_RETURN diagnosis is correct, the arithmetic checks out, and the timeout override is threaded through cleanly. The `docker-testnet-sync` job is a welcome addition but I'd like the empty-value hazards and the JSON-format assumption cleaned up before merge, and a fast-fail startup check would meaningfully improve the debugging story. Optional nits (`chunk_overhead` comment, HEAVY_SCRIPTS docstring, genesis externalisation) can be follow-ups.

azuchi (changes requested):
Thanks — everything I flagged last round is addressed:

- `feature_dbcrash.py` (both variants) added to `HEAVY_SCRIPTS`; `HEAVY_SCRIPTS`'s docstring now spells out that entries here are exercised only by the weekly workflow.
- `chunk_overhead = 3` gained the "assumes `MAX_SCRIPT_ELEMENT_SIZE ∈ [256, 65535]`" note.
- `docker-testnet-sync`: `2>/dev/null || echo 0` on every `cli` numeric read, `jq -r .initialblockdownload` instead of the grep-against-pretty-JSON assumption, genesis hex externalised to `contrib/docker/testnet-genesis-block-with-sig`, and the 30-second post-startup container-alive check with a log dump. Each of the accompanying comments explains *why* the guard is there, which is genuinely nicer than the code alone.

One remaining blocker on the test-runner half, then a couple of coordinating notes.

### 🚨 R1 — This PR now runs `feature_dbcrash.py` in **both** `--extended` and `--heavy`; daily CI on this branch fails as a result

The change adds `feature_dbcrash.py` (and its `--scheme SCHNORR` variant) to `HEAVY_SCRIPTS`, but leaves the same two entries in `EXTENDED_SCRIPTS`. So on this branch:

```console
$ git show pr/438:test/functional/test_runner.py | grep -B1 -A2 feature_dbcrash
    'feature_xfield_maxblocksize.py --scheme SCHNORR',
    'feature_dbcrash.py',
    'feature_dbcrash.py --scheme SCHNORR',
--
    'feature_pruning.py',
    'feature_dbcrash.py',
    'feature_dbcrash.py --scheme SCHNORR',
]
```

Which means the daily workflow (`test_runner.py --extended`) still picks it up. On the latest run against `1f9fa3a4b5` ([run 28859037969](https://github.com/chaintope/tapyrus-core/actions/runs/28859037969)), the same five Release+Wallet jobs I audited on PR #437 fail with the same signature:

```
feature_dbcrash.py                                     | ✖ Failed  | 1802 s
```

At iteration 37 the `restart_counts` are `[11, 4, 0]` — same failure mode as `master`, `restart_counts[2] > 0` never fires inside the 1800 s per-test cap.

**Fix.** Remove `feature_dbcrash.py` and `feature_dbcrash.py --scheme SCHNORR` from `EXTENDED_SCRIPTS` in the same commit. Then `--extended` (daily) skips it and `--heavy` (weekly) picks it up, which is what the PR title / job rename already claim to be doing:

```diff
     'feature_xfield_maxblocksize.py',
     'feature_xfield_maxblocksize.py --scheme SCHNORR',
-    'feature_dbcrash.py',
-    'feature_dbcrash.py --scheme SCHNORR',
 ]
```

Once that's in, this branch's daily CI should go green (nothing else on the failing jobs points to a different cause).

### 🟡 R2 — This PR fully supersedes PR #437; #437 should be closed

The two PRs are asymmetric halves of the same "move to weekly" operation:

- **PR #437** removes from `EXTENDED_SCRIPTS` but doesn't add to `HEAVY_SCRIPTS` — merged alone, `feature_dbcrash.py` is orphaned (I flagged this on #437 as a blocker).
- **PR #438** adds to `HEAVY_SCRIPTS` but doesn't remove from `EXTENDED_SCRIPTS` — merged alone, daily still fails (R1 above).

Once #438 also removes `feature_dbcrash.*` from `EXTENDED_SCRIPTS` (the diff in R1), it becomes a self-contained end-to-end "move to weekly" change and fully covers what #437 was trying to do. Recommend closing #437 as superseded.

Two small things to check when you do:

1. **`cc82b1aafb`'s `4/8/12` tiers + `DBCRASH_RESTART_TIMEOUT` are lost when #437 closes.** That's actually fine: the code comment on those tiers explicitly says the change was to fit inside the 1800 s daily budget, and once the test only runs under `--test-timeout=18000` (5 h) that rationale evaporates — the original `8/16/24` tiers converge easily with 10× the budget. If you *do* want to keep the aggressive tiers for tighter weekly coverage, port them into #438 as a follow-up commit with a *new* comment reflecting the weekly context (the current "to fit 1800 s" wording would be actively misleading in the new setup). Absent that, letting them go with #437 is the smallest-diff outcome.

2. **Mention in #438's PR description that it supersedes #437.** Just so whoever handles the merge queue doesn't try to land both.

### 🟢 Everything else

All previously noted G1/G2 (`blocktools.py` OP_RETURN chunking, `--test-timeout` / `--heavy` plumbing) still looks correct. The docker-testnet-sync hardening is a genuine improvement over the previous revision — none of the fragile spots I called out remain.

### Summary

Only R1 stands between this and green: also drop `feature_dbcrash.*` from `EXTENDED_SCRIPTS` in this PR so the "move to weekly" is atomic, then this branch's daily CI passes and merging doesn't leave master broken. R2 is just a coordination note with #437.

PR #440, opened 2026-07-08
Daily smoke benchmark ci

Naviabheeman:
Improvements to CI:
smoke test - run jobs in parallel
CI on PR - do not run benchmark test to save time
Daily CI - run benchmark.  plot a graph using the benchmark data using matplotlib python library

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -22,7 +22,7 @@ on:
       - '**'
 
 concurrency:
-  group: ${{ github.workflow }}-daily
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 env:
@@ -57,6 +57,7 @@ jobs:
             enable_gui: true
             enable_usdt: true
             enable_wallet: true
+            collect_benchmark_history: true
           - name: L-X86-Rel-No_GUI+USDT+Wallet
             os: ubuntu-latest
             arch: x86_64
@@ -114,6 +115,7 @@ jobs:
             enable_gui: true
             enable_usdt: false
             enable_wallet: true
+            collect_benchmark_history: true
           - name: M-ARM-Rel-No_GUI+Wallet
             os: macos-latest
             arch: arm64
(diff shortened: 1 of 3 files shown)


azuchi:
Two things I'd like to see folded in before this lands. Both are small; R2 in particular is directly on-topic for this PR's title.

### R1 — Move the benchmark-history opt-in from three literal `matrix.config.name` comparisons to a matrix field

The two canonical config names are hardcoded into `daily-test.yml` at:

- `:202` — `Restore Benchmark History cache` step's `if:`
- `:445` — `Save Benchmark History cache` step's `if:`
- `:778` — the in-line shell guard inside "Build and run daily tests" that decides whether to append to `history.csv` and render the chart

Renaming either config touches all three sites, and the failure mode of missing one is subtle (e.g. cache keeps persisting while the chart renders against stale data, or vice versa). Please add a matrix field:

```yaml
matrix:
  config:
    - name: L-X86-Rel-GUI+USDT+Wallet
      ...
      collect_benchmark_history: true
    - name: M-ARM-Rel-GUI+Wallet
      ...
      collect_benchmark_history: true
    # others: field omitted → falsy
```

then all three `if:`s read `matrix.config.collect_benchmark_history == true` (and the shell guard becomes `if [ "${{ matrix.config.collect_benchmark_history }}" = "true" ]; then`). Adding/removing/renaming a canonical config becomes a one-field change.

### R2 — Make `daily-test.yml`'s concurrency group per-ref, matching `smoke-test.yml`

Current `daily-test.yml:24-26`:

```yaml
concurrency:
  group: ${{ github.workflow }}-daily
  cancel-in-progress: true
```

The group name doesn't include `github.ref`, so every run of Cmake Daily Test — across every branch, PR, and schedule tick — competes for a single global slot. This is what caused this PR's daily run to cancel PR#439's mid-compile last week, and it will keep doing that whenever two PRs land daily runs within an hour of each other.

`smoke-test.yml:16-18` already uses the correct pattern:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

Applying the same one-line change to `daily-test.yml` isolates runs per-branch, so a push to PR X only cancels PR X's own in-flight daily run (which is the actual intent of `cancel-in-progress`). This is squarely inside this PR's stated scope ("Daily smoke benchmark ci") and a one-line change, so I'd rather see it here than as a follow-up.

Other than these two, everything else in the current revision looks good — the lint fix works, smoke parallelism is confirmed by the green matrix, and the single `L-X86-Deb-GUI+USDT+Wallet` red on the latest run is `feature_tokencreation.py` (7-second functional-test failure, passes on `L-ARM-Deb`), which this PR's diff can't have caused. A re-run should clear it.

Naviabheeman:
Apart from adding the review recommendations, the last commit also fixes the core dump creation and debugging steps.

azuchi (approved):

Issue #405, opened 2026-04-16
Refactor tapyrus-core and remove witness code and irrelevant wallet versioning.


Naviabheeman:
# Dead Code Analysis: Witness & Wallet Version

## 1. Witness Dead Code

Tapyrus does not support SegWit/witness, but the codebase retains extensive witness scaffolding inherited from Bitcoin Core. The findings are split into two tiers.

---

### Tier 1 — Entirely Unreachable (safest to delete)

| Item | File(s) | Notes |
|---|---|---|
| `BlockWitnessMerkleRoot()` | `consensus/merkle.cpp:80-89`, `merkle.h:28` | No callsites in production code |
| `if (tx->HasWitness())` block rejection | `validation.cpp:1423-1424` | Always false; the reject path is dead |
| `HasWitness()` branch in miner | `miner.cpp:205` | Dead branch — `scriptWitness` is never populated |
| `MSG_WITNESS_TX` / `MSG_WITNESS_BLOCK` / `MSG_FILTERED_WITNESS_BLOCK` cases | `net_processing.cpp:1097,1121,1244,1263,1406` | No peer sends these message types |
| `IsWitnessStandard()` | `policy/policy.cpp:183-232`, `policy.h:71` | Only callsite is inside `#ifdef DEBUG` gated on `HasWitness()` — never reached |
| Witness execution path in `VerifyScript` | `script/interpreter.cpp` (~lines 1625-1749) | Guarded by `SCRIPT_VERIFY_WITNESS` which is never included in production `flags` |
| `CountWitnessSigOps()` inner body | `script/interpreter.cpp:1727+` | Returns 0 immediately at the `(flags & SCRIPT_VERIFY_WITNESS) == 0` check; the call in `tx_verify.cpp:154` is a no-op |
| `"WITNESS"` service label in GUI | `qt/guiutil.cpp:879` | No Tapyrus node ever has `NODE_WITNESS` set |

---

### Tier 2 — Present but Always No-Op (clean up for clarity)

| Item | File(s) | Notes |
|---|---|---|
[...]

PR #441, opened 2026-07-09
Witness cleanup

Naviabheeman:
Remove witness related code completely:(issue #405)
 1 —Remove unreachable code (zero behavior change) 
 2 — Remove Hash/serialization (Tier 2 core: m_witness_hash, SERIALIZE_TRANSACTION_NO_WITNESS, scriptWitness)
 3 — Fix script-verification witness path
 4 — Fix network protocol constants and rename NODE_WITNESS to NODE_UNSUPPORTED_SERVICE and keep the peer disconnect behavior 
 5 — Fix tests and benchmarks
 6 — Edit documentation/comments


azuchi:
Two small in-scope items I'd like to see folded into this PR before it lands — both are within the spirit of the sweep and small enough that a separate PR feels like overhead.

### `rpc/mining.cpp` — dead `fSupportsSegwit` / `default_witness_commitment` block

`bool fSupportsSegwit = false;` at `rpc/mining.cpp:502` is hardcoded and never assigned true anywhere, so every downstream branch collapses:

- `:513` `fLastTemplateSupportsSegwit != fSupportsSegwit` — always `fLastTemplateSupportsSegwit != false`
- `:522` `fLastTemplateSupportsSegwit = fSupportsSegwit;` — always writing `false`
- `:526` `CreateNewBlock(scriptDummy, fSupportsSegwit)` — always passes `false`
- `:605-606`:
  ```cpp
  if (!pblocktemplate->vchCoinbaseCommitment.empty() && fSupportsSegwit) {
      result.pushKV("default_witness_commitment", HexStr(...));
  }
  ```
  never fires — `getblocktemplate` will never emit the `default_witness_commitment` field, so miners polling this RPC see an unreachable-in-practice key documented in our output.

Please strip `fSupportsSegwit` and its two bookkeeping variables, drop the `fSupportsSegwit` argument from `CreateNewBlock` (`miner.h` / `miner.cpp`), delete the `default_witness_commitment` push, and remove the now-stale `// to select witness transactions, after segwit activates ...` comment at `:500`. All of this falls under the same "Tapyrus never activated segwit, so this is unreachable" rationale as the commits already here.

### `net_processing.cpp:262` — stale doc comment

`fSupportsDesiredCmpctVersion`'s doc comment still opens with

```
/**
 * If we've announced NODE_WITNESS to this peer: whether the peer sends witnesses in cmpctblocks/blocktxns,
 * otherwise: whether this peer sends non-witnesses in cmpctblocks/blocktxns.
 */
```

which is factually wrong now — we don't announce NODE_WITNESS (it's `NODE_UNSUPPORTED_FEATURE` and used purely for peer rejection), and the "sends witnesses in cmpctblocks/blocktxns" concept doesn't apply to Tapyrus. Just needs a one-shot rewording — commit 6 caught `txmempool.h` and `transaction.h` doc comments but missed this one.

The rest of what I noted (the `SigVersion::WITNESS_V0`/BIP143-sighash branch, `SCRIPT_VERIFY_MINIMALIF`, the `IsWitnessProgram()` CLEANSTACK bypass, and the `CScriptWitness&` test-helper parameter) is fine as a follow-up "witness residuals" PR — those either touch semantics worth discussing separately or are pure noise that doesn't affect correctness.

azuchi (approved):

PR #442, opened 2026-07-13
Run tests in  parallel in weekly CI

Naviabheeman:
Last weekly CI job ended in timeout as Github action can run only for 6 hours. Change the job a little to improve parallelism:
- Collapsed feature-pruning/feature-dbcrash into one heavy-tests job with strategy.matrix — build pipeline lives once, matrix-driven test_args/name differentiate the two.
- Added step to Print full combined logs on failure: runs combine_logs.py directly over every test tmpdir and writes the full, untruncated output (all nodes + test framework log) to  — works even for a Timeout result, which test_runner.py itself never surfaces to the console.
- Retained the dead-timeout removal, chmod cleanup, ulimit fix, and dangling-process kill from the earlier split.

### DIFF
--- a/.github/workflows/weekly-heavy-tests.yml
+++ b/.github/workflows/weekly-heavy-tests.yml
@@ -25,7 +25,7 @@ on:
         required: false
 
 concurrency:
-  group: ${{ github.workflow }}-heavy
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 env:
@@ -206,7 +206,10 @@ jobs:
       #     fi
 
       - name: Print full combined logs (on failure)
-        if: failure()
+        # A job hitting the 6h GitHub-hosted-runner ceiling ends with
+        # conclusion "cancelled", not "failure" -- exactly the case this step
+        # exists for, so it must run on both.
+        if: failure() || cancelled()
         run: |
           # test_runner.py only ever prints the last 4000 lines of the
           # combined per-node logs to the console, and only for a genuine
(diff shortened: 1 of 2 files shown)


azuchi:
Matrix split shape is right. A few items worth resolving before merge:

### Recommended (in-scope)

1. **`if: failure()` on the combined-logs step won't fire on the failure mode this PR targets.** A job hitting the 6h runner ceiling terminates with `conclusion: cancelled`, not `failure`, so the log-dump step is skipped in exactly the case that motivated it. Use `if: failure() || cancelled()`.

2. **`$GITHUB_STEP_SUMMARY` has a 1 MB per-step cap.** `combine_logs.py` on a heavy test easily produces tens of MB and GitHub silently drops the excess, so the "full untruncated" claim doesn't hold in practice. Either `tail -c 950000` before appending, or write the full log to a file and upload it as an artifact with the summary linking to it.

3. **`concurrency.group: ${{ github.workflow }}-heavy` (line 28) is a single global slot.** Same shape as `daily-test.yml`'s old group that PR #440 fixed to `${{ github.workflow }}-${{ github.ref }}`. Weekly contention is rarer but a `workflow_dispatch` during a schedule run still cancels the earlier one unnecessarily. Apply the same one-line fix here for consistency.

### Nit (fine as follow-up)

4. **`Debug core dump (on failure)` is the pre-#440 shape** — single most-recent dump only, hardcoded `tapyrusd` binary, no summary output. PR #440 rewrote the daily version to loop over every `core.*`, resolve the crashing binary from `file` output, and emit `bt full` into `$GITHUB_STEP_SUMMARY`. Copy that block verbatim.

5. **`HEAVY_SCRIPTS` in `test_runner.py:255` is now a second source of truth for "what runs weekly."** Before: workflow ran `--heavy`, list lived in one place. After: the matrix drives it, and `HEAVY_SCRIPTS` is only kept alive by the completeness check at `test_runner.py:663`. Add a one-line comment on `HEAVY_SCRIPTS` saying "kept for the completeness check only; the weekly workflow drives what actually runs" so a future editor knows both places need updating together.

Also — the weekly workflow has no runs on this SHA (only `schedule` + `workflow_dispatch` triggers). A `workflow_dispatch` against this branch before merging would exercise the matrix split + `combine_logs.py` path for real.

PR #444, opened 2026-07-17
Witness CScriptWitness cleanup

Naviabheeman:
Earlier witness cleanup #441 didn't remove its support from script related code.  `CTxIn` still has `CScriptWitness`. This PR removes CScriptWitness and all code using it layer by layer.


azuchi:
Reads to me like this branch got left based on old master and hasn't been rebased — `git diff origin/master pr444` shows zero new content, and the 8 commits it lists are all already in `master` via #441's squash-merge (including the descriptor cleanup commit `568690c9b`). PR description says "CTxIn still has CScriptWitness" but that field was already removed in #441 Tier 2 (`00c8c04a`).

If the intent is Tier 3 — removing the `CScriptWitness` type itself along with the dead-but-kept parameter in `script_tests.cpp` / `sigopcount_tests.cpp`'s `BuildSpendingTransaction` / `DoTest`, the doc-only mentions in `script/sign.h`, and the Python-side residuals in `test_framework/{blocktools,messages}.py` — that code doesn't appear to be pushed yet. Rebase onto current `master` first and push the actual changes on top; then this PR will show a clean diff of only the new work.

azuchi:
Rebased and repushed — now this is a real 7-commit series and reviewing it as such. Overall the layering is thoughtful: each commit picks up exactly what became provably dead when the previous one landed, and the commit messages carry the reasoning (BIP143 midstate cache never wired to legacy sighash; `GetBlockScriptFlags` never includes CLEANSTACK; `PrecomputedTransactionData::ready` is never written so `cacheready` is always false; etc.). I spot-checked the load-bearing claims against `origin/master` and they hold up.

## Blocker — CI: Linux/Windows Release builds fail to link

All 7 Linux/Windows Release jobs die with the same symbol:

```
sign.cpp:(.text): undefined reference to
  uint256 SignatureHash<CMutableTransaction>(CScript const&,
    CMutableTransaction const&, unsigned int, int, long const&)
```

Debug builds and macOS pass — this is a Release-only optimization interaction. `MutableTransactionSignatureCreator::CreateSig` (`src/script/sign.cpp:25`) calls `SignatureHash` with a `CMutableTransaction`. The out-of-line template body lives only in `interpreter.cpp` and the file now has no explicit instantiations for `SignatureHash<T>` — only for `GenericTransactionSignatureChecker<T>` (lines 1440–1441). In Debug builds the checker's `CheckSig` is emitted as a real function that transitively pins `SignatureHash<T>`; in Release GCC inlines `CheckSig` inside `interpreter.cpp`, no external `CheckSig` body is emitted, and the transitive instantiation of `SignatureHash<T>` disappears with it — so `sign.cpp`'s reference is unresolved.

Add explicit instantiations at the same place the checker's are, e.g.:
```cpp
template uint256 SignatureHash(const CScript&, const CTransaction&, unsigned int, int, const CAmount&);
template uint256 SignatureHash(const CScript&, const CMutableTransaction&, unsigned int, int, const CAmount&);
```
(Removing the sigversion default arg widened this exposure — with the default, `sign.cpp`'s call previously produced the same effective signature as call sites inside `interpreter.cpp`, so any implicit instantiation there covered it. Now `sign.cpp` is the only TU that needs the `CMutableTransaction` overload and can no longer freeload.)

## Behavior changes worth flagging in the PR description / release notes

Two commits are not pure dead-code removal even though the surrounding series is:

- **`d34bfd52` — cleanstack bypass removal.** The bypass had no consensus effect (`GetBlockScriptFlags` doesn't include `SCRIPT_VERIFY_CLEANSTACK` — verified). But it did let `OP_0 <push>`-shaped scriptPubKey spends into the mempool. After this, such spends fail the CLEANSTACK standardness check. In practice these UTXOs shouldn't exist because `Solver()` already flags them `TX_NONSTANDARD` at creation, but if any ever landed on-chain (e.g. mined directly), they'd now need a miner to include them directly — no relay path. Worth calling out in release notes as a mempool-policy tightening rather than pure cleanup.
- **`3ae569fb` — `bitcoinconsensus_SCRIPT_FLAGS_VERIFY_WITNESS` removed.** The commit correctly identifies this as a real external ABI break. `bitcoinconsensus_version()` currently returns `BITCOINCONSENSUS_API_VER` unchanged — consider bumping so downstream `libtapyrusconsensus` consumers can detect the change instead of silently getting a no-op where they previously got `ERR_AMOUNT_REQUIRED`.

## Minor residuals (non-blocking)

Small cosmetic loose ends after the sweep:
- `src/test/data/tx_invalid.json:488,495` still reference `SigVersion::BASE` in description strings.
- `src/test/script_standard_tests.cpp:98,105` have `// TX_WITNESS_V0_KEYHASH` / `_SCRIPTHASH` comments describing test cases.
- `test/functional/test_framework/messages.py` still defines `CScriptWitness` and sets `scriptWitness` on `CTxIn`; `test/functional/test_framework/segwit_addr.py` still exists. These are the Python-side residuals from #441 — they don't seem to break anything today, but the C++/Python asymmetry is worth a follow-up (either wire it up or drop it).

Once the linker error is fixed and CI is green, this looks good to me.

azuchi:
nit: `test/functional/p2p_compactblocks.py:574` still has an `assert(tx.wit.is_null())` — same shape as the one you removed at line 332, and same reasoning applies (this file never sets `.wit`, and `tx` here is populated from a `blocktxn` message received from the daemon, so it's trivially null too). Either drop it for consistency with the line 332 removal, or leave a comment explaining why this one is kept as a smoke check.

Not a blocker — the rest of the update addresses every previous review point cleanly and CI is green. LGTM once this is sorted (or explicitly deferred).

azuchi (approved):

PR #445, opened 2026-07-24
Run heavy tests individually without using test runner list

Naviabheeman:
The weekly test run from last week did not run feature_pruning or feature_dbcache tests because these tests are not in the test_runner test list. This pr invokes the tests as individual python scripts bypassing the list.

### DIFF
--- a/.github/workflows/weekly-heavy-tests.yml
+++ b/.github/workflows/weekly-heavy-tests.yml
@@ -9,8 +9,14 @@ name: Weekly Heavy Test
 # loosest dbcrashratio/dbcache tier to reliably crash at least once).
 # Each test runs in its own job (feature-pruning, feature-dbcrash) so a slow
 # run of one doesn't eat into the other's share of the 6h GitHub-hosted-runner
-# execution ceiling -- combining them previously caused a TIMEOUT-retry of
-# feature_pruning to blow through that ceiling mid-job.
+# execution ceiling. Each script is invoked directly (test/functional/<script>.py),
+# not via test_runner.py: these scripts are deliberately excluded from
+# test_runner.py's ALL_SCRIPTS list (see HEAVY_SCRIPTS there) so they never run
+# as part of the daily base/extended suites, but that also means
+# test_runner.py can't be given their names positionally -- its lookup only
+# matches ALL_SCRIPTS, so passing e.g. "feature_pruning.py" on the command
+# line used to silently no-op (WARNING + empty test list + exit 0), making
+# this job appear to pass without ever running anything.
 # Scheduled weekly; can also be triggered manually or after a successful daily build.
 on:
   schedule:
@@ -43,9 +49,12 @@ jobs:
       matrix:
         config:
           - name: feature_pruning
-            test_args: 'feature_pruning.py'
+            scripts:
+              - 'feature_pruning.py'
           - name: feature_dbcrash
-            test_args: 'feature_dbcrash.py "feature_dbcrash.py --scheme SCHNORR"'
+            scripts:
+              - 'feature_dbcrash.py'
+              - 'feature_dbcrash.py --scheme SCHNORR'
     name: ${{ matrix.config.name }} test (Linux x86_64 RelWithDebInfo)
     runs-on: ubuntu-latest
     # No timeout-minutes set: GitHub-hosted runners hard-cap job execution at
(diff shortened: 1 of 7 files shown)


azuchi:
Diagnosis is spot-on — the previous `${{ matrix.config.test_args }}` was being passed to `test_runner.py` as positional args, which only match against `ALL_SCRIPTS` (test_runner.py:346), so heavy names hit the WARNING branch and the job exited 0 without running anything. The switch to direct `python3 test/functional/<script>.py` invocation is the right fix, and the `--configfile`/`--cachedir`/`--portseed` mirror of what `test_runner.py` auto-injects is well-justified in the comment.

A few nits, plus one bit of scope worth considering:

### Nits

1. **Stale `--heavy` reference in `weekly-heavy-tests.yml:198`** — the comment reads "this matches test_runner.py's own --heavy behavior, which also runs both variants against a single shared cache," but `--heavy` is deleted in this same PR, so the reader can no longer verify the claim. Reword to something like "matches how `test_runner.py` itself shares one per-run cache across a batch (see `test_runner.py:449–453`)."

2. **Stale field name in `test_runner.py:253-257`** — the `HEAVY_SCRIPTS` docstring still says "matrix.config.test_args," but the new field is `matrix.config.scripts`. Please update to keep the pointer accurate.

### Improvement worth considering (same-PR or follow-up)

The `--heavy` removal closes the CI-side foot-gun, but leaves the CLI-side one open: a developer running `python3 test_runner.py feature_pruning.py` by hand still hits the same silent-pass path (`WARNING! Test not found in full test list.` → empty `test_list` → `sys.exit(0)`) — exactly the trap this PR exists to eliminate for CI. A cheap fix at `test_runner.py:346`:

```python
if test in ALL_SCRIPTS:
    test_list.append(test)
elif any(test == h.split()[0] for h in HEAVY_SCRIPTS):
    print("{}ERROR!{} Test '{}' is a heavy test — invoke it directly with "
          "python3 test/functional/{}, not via test_runner.py.".format(BOLD[1], BOLD[0], test, test))
    sys.exit(1)
else:
    print("{}WARNING!{} Test '{}' not found in full test list.".format(BOLD[1], BOLD[0], test))
```

Not a blocker for this PR — happy to see it as a follow-up if you prefer to keep this PR narrowly scoped to the CI workflow.

### Pre-merge checks

Since `weekly-heavy-tests.yml` triggers on `schedule`/`workflow_dispatch` only (not `pull_request`), none of the new invocation path has actually run in this PR's CI. Recommend a manual `workflow_dispatch` against this branch (`test_branch=WeeklyHeavyTestCI`) to confirm the direct-invocation path, cache/config/portseed handling, and the log-collection step all work end-to-end before merging.

The one failing daily check (`L-X86-Deb-GUI+USDT+Wallet` — `feature_tokencreation.py` wallet balance flake) is unrelated to this PR.

azuchi:
Before merging, could you dispatch the `Weekly Heavy Test` workflow once against this branch on your fork? Since PR#445 modifies `weekly-heavy-tests.yml` itself, and that workflow only triggers on `schedule` / `workflow_dispatch` (not `pull_request`), the new direct-invocation path hasn't actually been exercised in any CI run yet — the 20 green checks above are the daily/smoke workflows, which don't run this file.

Base-repo `workflow_dispatch` UI doesn't list branches from forks, so it needs to be triggered from your fork:

1. Go to https://github.com/Naviabheeman/tapyrus-core/actions/workflows/weekly-heavy-tests.yml (enable Actions on the fork first if it's disabled).
2. Click "Run workflow", select branch `WeeklyHeavyTestCI`, leave `test_branch` empty.
3. Wait for both matrix jobs (`feature_pruning` and `feature_dbcrash`) to finish — this takes several hours (feature_pruning ~90 min – 4 h, feature_dbcrash longer since it runs twice for ECDSA + SCHNORR).

What to verify from the run logs, to confirm the tests actually executed rather than silently no-op'ing again:

- Each matrix job's step output shows `---------------- Start feature_pruning.py ----------------` / `---------------- Start feature_dbcrash.py ... ----------------` and then `... passed`.
- `feature_dbcrash` job runs both `feature_dbcrash.py` and `feature_dbcrash.py --scheme SCHNORR` (two `Start ... / passed` pairs).
- Total step duration is on the order of hours, not seconds (the pre-fix silent-pass ran in ~30s).

Once both matrix jobs are green with real execution times, this is ready to merge. Happy to take a look at the run URLs if anything surprises you.

azuchi:
The new commit's motivation is directly validated by your fork's earlier `workflow_dispatch` run [30338166600](https://github.com/Naviabheeman/tapyrus-core/actions/runs/30338166600): Depends build passed, `feature_dbcrash` completed cleanly in ~2.5 h, and `feature_pruning` reached the exact call site this commit changes:

```
11:03:09  TestFramework (INFO): Syncing node 5 to test wallet
11:07:13  TestFramework (WARNING): Timeout. None                ← ~240s (TAPYRUSD_SYNC_TIMEOUT)
11:07:16  feature_pruning.py FAILED
```

Node 5 has been disconnected since `setup_network()`, so at this point it's doing initial block download of the whole chain the test just built — a duration that scales with chain size, not a fixed budget. Swapping in stall-detection is the right fix.

### Design of `sync_blocks_with_stall_detection`

Looks solid. Notes:

- `getbestblockhash()` for the equality check + `getblockcount()` for progress tracking is a clean split — height changes reset the stall timer, best-hash convergence is the exit condition.
- 60s progress-log interval keeps CI output non-silent during a long sync — useful for the 90 min – 4 h IBD.
- First-iteration bootstrap is correct: `last_heights = None` on entry, first observed heights become baseline, stall clock starts from 0. A sync that hasn't started yet won't trip stall immediately.
- Retaining `sync_blocks()` unchanged is the right non-breaking choice; other tests are unaffected.

### Small suggestion — docstring wording

The current caveat ("at least one node already synced to the latest, stable tip") is correct, but leaves "stable" implicit. A general-purpose caller could easily misuse this helper by passing a set that includes an active block producer, in which case heights keep advancing forever and the loop never exits (no wall-clock ceiling to catch it). Consider adding one line spelling out that "stable" means "not producing new blocks":

> The rpc_connections set must include at least one node that is synced to the target tip **and not producing further blocks**; otherwise the progress-based stall detector never fires and this function loops indefinitely.

Optional; the feature_pruning call site trivially satisfies this since node 0 stops mining before `connect_nodes` is called.

### Scope question

This commit expands PR#445 beyond "run heavy tests individually" into a general test-framework helper. Given the empirical link (the sync timeout would immediately re-fail feature_pruning the moment the CI change actually runs the test), bundling is defensible. Splitting into a standalone "add stall-detection sync helper + use it in feature_pruning" PR merged first would keep this PR narrower and reviewer-friendlier for the CI-only change — up to you. Not a blocker either way.

### CI status — 4 failures unrelated

The four failing daily checks (`L-X86-Deb-GUI+USDT+Wallet`, `L-X86-Rel-GUI+USDT+Wallet`, `L-ARM-Deb-GUI+USDT+Wallet`, `L-ARM-Rel-GUI+Wallet`) all die at the Depends-build step with:

```
curl failed to verify the legitimacy of the server
Depends build failed
```

Every `+GUI` variant fails (they build Qt via `depends`), every `+No_GUI` and cross-compile variant passes. Same symptom on your fork's most recent `workflow_dispatch` at 13:53Z. This is a global GitHub-hosted-runner infra flake today (likely upstream cert/mirror), unrelated to any code change in this PR. Recommend retrying once the flake clears rather than treating the reds as PR-blocking.

### Remaining gate before merge

The one thing not yet confirmed end-to-end: a `workflow_dispatch` run against this branch that includes commit `3244b2a5` and reaches the test-execution phase (not blocked by Depends-build infra). Once the infra flake clears, one more dispatch is worth doing — success criteria being that `feature_pruning` gets past `Syncing node 5 to test wallet` (which it couldn't previously) and both matrix jobs finish with genuine execution times (hours, not the pre-fix ~30s silent-pass or the 4-min sync-timeout failure).

### Follow-up (out of scope, not blocking)

Worth grepping for other `sync_blocks(..., timeout=TAPYRUSD_SYNC_TIMEOUT)` call sites in tests that also do IBD-of-a-long-chain patterns — same latent flake, and same helper would fit. Not needed for this PR.

Once (a) the infra flake clears and daily CI is green, and (b) one `workflow_dispatch` completes with `feature_pruning` past the sync step, this is ready to merge.

azuchi:
`5064e23b` cleanly addresses the docstring caveat I flagged on `sync_blocks_with_stall_detection`. LGTM.

`e8767cfa` extends the stall-detection pattern to three more sites (`feature_federation_management.py`, `feature_xfield_maxblocksize.py`, `feature_pruning.py:test_height_min`). Each replacement is well-motivated (same root cause as `feature_pruning`'s Node 5 sync: operation duration scales with something unbounded by a fixed budget). Two observations:

### 1. Duplication — four near-identical stall-detection loops now

After this commit the codebase has:
- `sync_blocks_with_stall_detection()` (util.py, pre-existing)
- `wait_for_node3_sync()` in `feature_federation_management.py:669-689`
- `wait_for_node3_sync()` in `feature_xfield_maxblocksize.py:605-624`
- `wait_for_prune()` in `feature_pruning.py:124-156`

All four are the same shape: track a progress metric → reset `last_progress_time` on change → raise on `stall_timeout` elapsed → log every `progress_log_interval` → `time.sleep(1)`. The two `wait_for_node3_sync` implementations are literally the same code modulo the peer-count expression (`< 2` vs `== 0`) and the reconnect target list. This cries out for a single reusable helper in `util.py` — something like:

\`\`\`python
def wait_with_stall_detection(done_condition, progress_snapshot, *,
                              stall_timeout, progress_log_interval,
                              before_check=None, log_line_fn=None):
    ...
\`\`\`

Suggest as a follow-up rather than blocking this PR — the extraction is a pure refactor and cleaner in a standalone commit.

### 2. `wait_for_prune`: `calc_usage()` TOCTOU (existing, not introduced here)

`calc_usage` (feature_pruning.py:67) does `os.listdir` → `os.path.isfile` → `os.path.getsize`; if a file is unlinked between `listdir` and `getsize`, `getsize` throws `FileNotFoundError` and `wait_for_prune` propagates it as a test failure. Pruning is tied to block-connection cadence rather than a concurrent GC thread so the race is very unlikely in practice, but the new `wait_for_prune` calls `calc_usage` in a tight loop while pruning is actively happening — the first place in the test that does so. Not a blocker; worth guarding with a `try/except FileNotFoundError: continue` if the flake ever shows up.

azuchi (approved):

PR #447, opened 2026-08-07
Add PSTT (TIP-174) Phase 1: core data structures and Signer

Naviabheeman:
Implements the Partially Signed Tapyrus Transaction format that replaces the inherited Bitcoin-shaped PSBT (BIP-174 v0): no global unsigned tx, per-input outpoint fields instead, Colored Coin aware, ECDSA/Schnorr aware. Full specification in doc/tapyrus/pstt.md.

- src/pstt.h / src/pstt.cpp: PartiallySignedTapyrusTransaction, PSTTInput, PSTTOutput structs with full wire-format Serialize/Unserialize, IsNull/ Merge/IsSane (incl. mixed ECDSA/Schnorr rejection and the per-field-class Combiner conflict policy), ComputeLocktime, GetIdentifier, the private MaterializeTransaction helper, xpub keydata helpers with a strict single-network prefix policy, and SignPSTTInput (ahead of its originally planned phase, since it was natural to land alongside the structs it depends on).
- src/test/pstt_tests.cpp: structural round-trip/format tests, verified against TIP-174's own published fixtures (all valid stages parse, all parse-stage invalid vectors throw, all rule-stage vectors parse-fine as the spec intends); plus mempool/consensus-level valid & invalid transaction tests in the style of txvalidation_tests.cpp (ECDSA spend, Schnorr spend, wrong-key rejection, colored-coin issue+transfer across a two-input/two-key PSTT, and a real bad-txns-token-without-fee rejection).
- test/functional/rpc_pstt.py: functional test ported from rpc_psbt.py's operational coverage to the PSTT RPC surface, plus a TIP-174-fixture- driven section (stage decoding, identifier stability, extract+broadcast on each entry's final stage, parse/rule dispatch for invalid vectors). Not yet runnable or registered in test_runner.py -- the RPCs it drives land in later phases.
- test/functional/data/tip174_valid.json, tip174_invalid.json, tip174_SOURCE.md: TIP-174's own test vectors, pinned to the exact chaintope/tips commit fetched.

### DIFF
--- a/doc/tapyrus/pstt.md
+++ b/doc/tapyrus/pstt.md
@@ -1,0 +1,231 @@
+Partially Signed Tapyrus Transaction (PSTT)
+============================================
+
+PSTT is Tapyrus's own partially-signed-transaction format, defined by
+[TIP-174](https://github.com/chaintope/tips/blob/master/tip-0174.md) and
+implemented in `src/pstt.h`/`src/pstt.cpp`. It replaces the Bitcoin-shaped
+PSBT (BIP-174 "v0") that tapyrus-core previously inherited unmodified: PSBT
+has no counterpart for Tapyrus's malleability-fixed txid, no notion of
+Colored Coins, and hardcodes ECDSA-only signing.
+
+PSTT is built on BIP-370's constructable (per-input/output) data model
+instead of BIP-174's single embedded unsigned transaction: there is no
+global unsigned tx record. Each input carries its own previous-txid/output-index
+pair directly, which is what lets a PSTT be built up incrementally across
+multiple round trips between parties (see [Fee Provider workflow](#fee-provider-workflow)
+below) rather than requiring one party to fix the whole set of inputs and
+outputs up front.
+
+Key differences from BIP-174 PSBT:
+* No global unsigned transaction; per-input outpoint fields instead.
+* No witness fields (Tapyrus has no segwit) and no Taproot fields.
+* Colored Coin aware: `PSTT_OUT_SCRIPT` may carry the `<color id> OP_COLOR`
+  prefix, and CP2SH redeem-script verification accounts for it.
+* Signer supports both ECDSA and Schnorr, and rejects mixing the two
+  schemes within one input's signature set.
+
+---
+
+Wire format
+-----------
+
+A PSTT is `PSTT_MAGIC_BYTES` followed by a global key-value map, then one
+key-value map per input, then one per output. Every map is terminated by an
+empty key (a single `0x00` length byte) — a map with no terminator is a
+parse error, not an implicit end-of-data.
+
+```
+magic bytes: 0x70 0x73 0x74 0x74 0xFF  ("pstt" + 0xFF)
+```
+
+Each record in a map is `<key><value>`, where both `<key>` and `<value>`
+are themselves length-prefixed byte strings (a CompactSize byte count
+followed by that many bytes) — this holds even for fixed-width values like
+a `uint32_t` or a `uint256`; those still get wrapped in the same
+length-prefix envelope as every other value. `<key>`'s first byte is the
+record's type; any remaining key bytes are that type's keydata (e.g. a
+pubkey for `PSTT_IN_PARTIAL_SIG`).
+
+### Global types
+
+| Name | Value | Keydata | Value | Notes |
+|---|---|---|---|---|
+| `PSTT_GLOBAL_XPUB` | `0x01` | `<xpub>` — the 78-byte BIP32 extended public key (4-byte version prefix + `CExtPubKey`'s 74-byte encoding) | `<fingerprint><32-bit uint>*` — master fingerprint + derivation path | Prefix must equal exactly `Params().Base58Prefix(EXT_PUBLIC_KEY)` for the running node's own network; any other prefix, including the other Tapyrus network's, is rejected at parse time |
+| `PSTT_GLOBAL_TX_FEATURES` | `0x02` | none | `int32_t` | **Required.** Any value is accepted and round-tripped verbatim into `CMutableTransaction::nFeatures` on extraction — no validation is imposed here beyond what the transaction layer itself imposes |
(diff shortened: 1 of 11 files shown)


azuchi (changes requested):
Thanks for the substantial Phase 1 work — the wire-format validation is thorough, the constexpr linker fix is minimal and targeted, and the end-to-end colored-coin test is a good smoke of the whole PSTT pipeline. I'd like to flag a few items before this lands.

## Must-fix

### 1. `PartiallySignedTapyrusTransaction::Merge()` has no size guard (pstt.cpp:573–593)

The loop iterates `inputs.size()` and dereferences `other.inputs[i]` via `operator[]`, with no check that `other.inputs.size() >= inputs.size()`. Same for outputs. The docstring on `PSTTInput::Merge` says "HasSameIdentifierAs() is checked by the Combiner before Merge() is ever called," but nothing at the API level enforces that; a caller that merges two independently-constructed PSTTs with mismatched shapes gets UB. Suggest:

```cpp
if (inputs.size() != other.inputs.size() || outputs.size() != other.outputs.size()) {
    throw std::invalid_argument("PSTT Merge() with mismatched input/output counts");
}
```

or a `HasSameIdentifierAs(other)` precondition assert inside `Merge()` itself.

### 2. TIP-174 fixture verification is present in the repo but not actually run in Phase 1

The PR body describes `pstt_tests.cpp` as "verified against TIP-174's own published fixtures," but `grep tip174 src/test/pstt_tests.cpp` is empty; the only consumer of `data/tip174_valid.json` / `data/tip174_invalid.json` is `test/functional/rpc_pstt.py`, which the PR body itself notes is "not yet runnable or registered in `test_runner.py`." So today no CI job actually validates the wire format against TIP-174 vectors — a silent drift wouldn't be caught. For Phase 1, either wire a minimal `pstt_tests.cpp` case that parses `tip174_valid.json`'s stages, or make the fixture-driven section of `rpc_pstt.py` runnable standalone.

## Recommended

### 3. Comment vs. code: UTXO / PREVIOUS_TXID mismatch is not enforced at Unserialize (pstt.cpp:968–971)

`SignPSTTInput`'s comment claims "it's also enforced at Unserialize() time on the whole PSTT," but `PSTTInput::IsSane()` (called at the end of `Unserialize()`) doesn't check `input.utxo->GetHashMalFix() == input.previous_txid`. Either add the check to `IsSane()` (defense in depth) or drop the comment; today it's misleading.

### 4. `sighash_type = 0` is folded into "absent" on wire (pstt.cpp:216–218, 313–318, 994)

`sighash_type` is a plain `int` defaulting to 0. Serialize skips it if `sighash_type == 0`, but Unserialize accepts an explicit-wire `PSTT_IN_SIGHASH_TYPE = 0` and sets the field to 0 with no way to distinguish "wire had 0" from "wire absent" on a subsequent re-serialize. `Merge()` and `SignPSTTInput` also treat 0 as absent. Since 0 isn't a valid sighash in Tapyrus this is defensible, but it breaks strict round-trip and diverges from every other optional field in the struct. Suggest `boost::optional<int32_t>` for parity.

### 5. `prev_out_index >= utxo->vout.size()` reports `UTXO_TXID_MISMATCH` (pstt.cpp:975–977)

The txid matches; the *index* is out of range. Consider a distinct enum value (e.g. `PREV_OUT_INDEX_OOB`) so RPC callers can surface an accurate error to users.

## Coverage gaps

None of these are correctness bugs on their own, but together they leave a lot of the added surface without a regression net:

- `PSTTSignResult`'s 7 non-OK values (MISSING_UTXO, UTXO_TXID_MISMATCH, REDEEM_SCRIPT_HASH_MISMATCH, SIGHASH_CONFLICT, SCHEME_CONFLICT, LOCKTIME_INVALID, SIGHASH_SINGLE_OOB) are never exercised; only OK is reached in the pipeline tests. The `wrong-key` case still returns OK from `SignPSTTInput`.
- `PSTTInput::Merge()`'s five throw sites (UTXO / FINAL_SCRIPTSIG / PARTIAL_SIG conflicts, REDEEM_SCRIPT / BIP32_DERIVATION mismatches) have no test.
- Output-side reserved keytype rejection (0x01, 0x05, 0x06, 0x07) has no analog to `pstt_rejects_reserved_input_keytype`.
- The `< 500000000` / `== 0` / `>= 500000000` parse-time throws on `PSTT_IN_REQUIRED_*_LOCKTIME` are untested.
- `PSTT_TXMOD_RESERVED_MASK` (bits 3–7) rejection via `IsSane()` is untested.

Given how much of `IsSane` / `Merge` / `SignPSTTInput` this leaves uncovered, I'd expect at least a first-round expansion here in Phase 1 rather than deferring to a later phase.

## Follow-up nit

### `CTxIn::SEQUENCE_FINAL` still has the same ODR-use trap (primitives/transaction.h:73)

The linker fix converted `CURRENT_FEATURES` / `MAX_STANDARD_FEATURES` to `static constexpr`, but the neighbouring `static const uint32_t SEQUENCE_FINAL = 0xffffffff` (and the sequence-lock constants below it) are the same shape. The PR sidesteps by writing `input.sequence.value_or(0xFFFFFFFFu)` with an inline comment, but the next contributor to write `optional<uint32_t>::value_or(CTxIn::SEQUENCE_FINAL)` will hit the identical linker error on Linux GCC. Worth converting these to `constexpr` in the same patch (or a follow-up) to remove the trap entirely.

## Positive observations for the record

Wire-format validation is genuinely thorough: magic, missing separator, duplicate keys, count mismatch, `0x00` global rejection, `version > 0` rejection, xpub keydata size, XPUB value length divisibility, the 37/57 CP2SH-hash extraction offsets, the DER lower-bound sanity check, and the CompactSize-value trailing-byte guard are all in the right places. The constexpr fix is correctly diagnosed (Linux/GCC's strict ODR-use enforcement vs. clang/ld64 tolerance) and matches `pubkey.h`'s existing pattern.

azuchi (approved):

PR #448, opened 2026-08-10
Add PSTT (TIP-174) Phase 2+3: Signer, Combiner, Finalizer, Extractor RPCs

Naviabheeman:
- Phase 2 — Signer RPCs:
  - decodepstt (full JSON breakdown, including computed locktime and identification_txid) and signpsttwithkey (signs with an explicit key list via FlatSigningProvider), plus DecodePSTT/EncodePSTT in core_io/core_read/core_write, mirroring the existing PSBT equivalents.
  - Wallet-side FillPSTT (src/wallet/pstt.{h,cpp}): combined Updater+Signer pass — attaches PSTT_IN_UTXO from mapWallet where missing, signs via SignPSTTInput against the real wallet or a PublicOnlySigningProvider (for redeem-script/BIP32-path discovery without producing signatures), fills output BIP32 keypaths via a throwaway dummy transaction (PSTT has no embedded global tx to piggyback on, unlike PSBT). Also adds ComputePsttColorBalances.
  - New wallet RPCs: walletupdatepstt (Updater only), walletsignpstt (Signer only), walletprocesspstt (combined, mirrors walletprocesspsbt).
- Phase 3 — Combiner, Finalizer, Extractor:
  - New public ExtractPSTT(): requires PSTT_IN_FINAL_SCRIPTSIG on every input (throws naming the first one missing it), computes the final locktime, materializes the signed transaction.
  - combinepstt (decodes all inputs, verifies pairwise HasSameIdentifierAs(), merges via Merge(), checks IsSane() on the result), finalizepstt (refuses while either modifiable flag is set; derives sighash/scheme from what each input already carries rather than hardcoding, so SignPSTTInput's conflict checks are structurally unreachable here; strips signing-only fields on success; extracts to hex when complete), and extractpstt (thin wrapper over ExtractPSTT).
- Test coverage: removed the Phase-1 ExtractForTest stand-in now that a real ExtractPSTT exists; added Merge()'s refuse-tier (conflicting partial sig) and pick-first-tier (sighash conflict) coverage; a full end-to-end 2-of-2 P2SH multisig test (fund → two parties sign independently → Merge() → completeness check → ExtractPSTT → mempool/mined); and a new RPC-dispatch layer (CallPsttRPC helper via tableRPC, mirroring rpc_tests.cpp's CallRPC) exercising combinepstt/finalizepstt/extractpstt through their real registered actors rather than just the underlying pstt.cpp calls.

### DIFF
--- a/src/core_io.h
+++ b/src/core_io.h
@@ -15,6 +15,7 @@ class CScript;
 class CTransaction;
 struct CMutableTransaction;
 struct PartiallySignedTransaction;
+struct PartiallySignedTapyrusTransaction;
 class uint256;
 class UniValue;
 
@@ -26,12 +27,14 @@ bool DecodeHexBlk(CBlock&, const std::string& strHexBlk);
 uint256 ParseHashStr(const std::string&, const std::string& strName);
 std::vector<unsigned char> ParseHexUV(const UniValue& v, const std::string& strName);
 bool DecodePSBT(PartiallySignedTransaction& psbt, const std::string& base64_tx, std::string& error);
+bool DecodePSTT(PartiallySignedTapyrusTransaction& pstt, const std::string& base64_tx, std::string& error);
 int ParseSighashString(const UniValue& sighash);
 
 // core_write.cpp
 UniValue ValueFromAmount(const CAmount& amount);
 std::string FormatScript(const CScript& script);
 std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
+std::string EncodePSTT(const PartiallySignedTapyrusTransaction& pstt);
 std::string SighashToStr(unsigned char sighash_type);
 void ScriptPubKeyToUniv(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
 void ScriptToUniv(const CScript& script, UniValue& out, bool include_address);
--- a/src/core_io.h
+++ b/src/core_io.h
@@ -18,6 +18,7 @@ struct PartiallySignedTransaction;
 struct PartiallySignedTapyrusTransaction;
 class uint256;
 class UniValue;
+enum class SignatureScheme;
 
 // core_read.cpp
 CScript ParseScript(const std::string& s);
@@ -29,6 +30,7 @@ std::vector<unsigned char> ParseHexUV(const UniValue& v, const std::string& strN
 bool DecodePSBT(PartiallySignedTransaction& psbt, const std::string& base64_tx, std::string& error);
 bool DecodePSTT(PartiallySignedTapyrusTransaction& pstt, const std::string& base64_tx, std::string& error);
 int ParseSighashString(const UniValue& sighash);
+SignatureScheme ParseSigSchemeString(const UniValue& sigscheme);
 
 // core_write.cpp
 UniValue ValueFromAmount(const CAmount& amount);
--- a/src/core_read.cpp
+++ b/src/core_read.cpp
@@ -6,6 +6,7 @@
 
 #include <primitives/block.h>
 #include <primitives/transaction.h>
+#include <pstt.h>
 #include <script/script.h>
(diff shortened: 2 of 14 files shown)


azuchi:
@Naviabheeman please fix conflict.

azuchi (commented):
Phase 2+3 layers cleanly onto Phase 1: role separation follows the spec, the DUMMY_SIGNING_PROVIDER trick in finalizepstt makes the sighash/scheme conflict paths structurally unreachable from the finalizer (any failure there means a genuine problem, not a spurious mismatch against our own derivation), and the last commit correctly propagates the boost::optional<int32_t> sighash_type change from Phase 1 to every consumer added here. The tableRPC-dispatch test pattern is a nice touch — it exercises argument parsing and error wrapping in the RPC layer, not just the pstt.cpp calls underneath.

A few small points worth addressing:

1. `ParseSigSchemeParam` / `signpsttwithkey` silent-fallback on unknown sigscheme (rpc/rawtransaction.cpp:520-521, wallet/rpcwallet.cpp:26): only exact `"SCHNORR"` selects Schnorr; anything else (`"Schnorr"`, `"schnorr"`, a typo, `"ECDSA"` misspelled) silently falls back to ECDSA. `ParseSighashString` rejects unknown strings with a clear error — please match that pattern here so a user typo doesn't quietly sign with the wrong scheme.

2. `extractpstt` / `ExtractPSTT` don't call `IsSane()` (src/pstt.cpp:928, rpc/rawtransaction.cpp:437): only `final_script_sig` presence and locktime are checked. A PSTT with an output missing `PSTT_OUT_AMOUNT` will silently materialize with `amount=0` (via `output.amount.value_or(0)` in `MaterializeTransaction`). Mempool acceptance would reject the resulting tx so it doesn't produce a wrong signed tx in practice, but a defensive `if (!pstt.IsSane()) throw ...` in `ExtractPSTT` would fail earlier with a clearer message and match the invariant that `combinepstt`/`finalizepstt` already maintain upstream.

3. Magic number in `walletupdatepstt` (wallet/rpcwallet.cpp:69): `FillPSTT(pwallet, pstt, 1, /*sign=*/false, bip32derivs)` passes literal `1` for `sighash_type`. It's SIGHASH_ALL and effectively unused when `sign=false` (the check at `wallet/pstt.cpp:36` is guarded by `sign`), but `SIGHASH_ALL` reads more clearly.

4. Wallet RPCs don't pass `fAllowNull=true` to `RPCTypeCheck` (wallet/rpcwallet.cpp:60, 115, 196), while the rawtx PSTT RPCs do (rawtransaction.cpp:274, 356, 490). The `.isNull()`-then-default branches in the wallet RPCs look prepared for null but explicit `null` for an optional trailing param will fail at the typecheck before reaching them. Please align — either add `true` to the wallet RPCs, or drop the unreachable `.isNull()` branches.

5. `combinepstt` catches only `std::invalid_argument` from `Merge()` (rawtransaction.cpp:298). Current `Merge()` only throws that type, so this is correct today; if a future addition throws `std::runtime_error` (used elsewhere in pstt.cpp) it'd surface as an internal RPC error instead of the friendly "Cannot combine PSTTs: ..." message. Widening to `std::exception` would be defensive.

Test coverage gaps to fill:

- No RPC-dispatch test for `signpsttwithkey` (the other new rawtx RPC gets the tableRPC treatment but this one only via its underlying library calls).
- No unit/functional test for the three new wallet RPCs (`walletupdatepstt`/`walletsignpstt`/`walletprocesspstt`).
- No test asserting `finalizepstt`'s signing-only-field strip actually happens on completion.
- No test for `finalizepstt` on a partially-completable PSTT (mix of completable inputs plus `MISSING_UTXO` inputs).
- No shape/format test for `decodepstt`'s JSON output.

azuchi (approved):

PR #450, opened 2026-08-14
 PSTT (TIP-174): Creator/Constructor RPCs

Naviabheeman:
**Phase 4 — Creator and Constructor RPCs.**
  Implement all RPCs needed for creation and modification of PSTT:
- `createpstt` Creator role: builds a new PSTT from a list of inputs/outputs, with no signatures, ready for Constructor/Updater/Signer round-tripping.
- `converttopstt` Creator role: converts an existing raw (network-serialized) transaction into an unsigned PSTT.
- `addinputtopstt` Constructor role: appends one input to a PSTT whose Inputs-Modifiable flag is still set.
- `addoutputtopstt` Constructor role: appends one output to a PSTT whose Outputs-Modifiable flag is still set.
- `addinputoutputpairtopstt`  Constructor role: appends one input and one output together in a single call, required once a SIGHASH_SINGLE signature is already present.
- `walletcreatefundedpstt` Creator+Updater role: builds a PSTT from the given inputs/outputs, funds it from the wallet if needed, and attaches known UTXOs.
- `walletupdatepstt` Updater role: fills in UTXO (and optionally BIP32 derivation) data for a PSTT's inputs/outputs from the wallet, without signing.
- `decodepstt`  Diagnostic RPC: returns a PSTT's contents (globals, inputs, outputs) as human-readable JSON.

Add `rpc_pstt.py` test and move previous json valid and invalid input tests to unit test.


azuchi:
@Naviabheeman please fix conflicts and CI.

azuchi (changes requested):
The rebase is clean and CI is green — the tree is byte-identical to the pre-rebase head, so this is the same code as before. A few things worth fixing before merge.

### 1. `src/rpc/client.cpp` has no PSTT entries at all

`vRPCConvertParams` registers every PSBT equivalent (`createpsbt` 0–3, `walletcreatefundedpsbt` 0–4, `combinepsbt` 0, `finalizepsbt` 1, `converttopsbt` 1), but there is not a single `pstt` row. Without them `tapyrus-cli` passes arguments through as JSON strings and the RPCs throw on `RPCTypeCheck`:

```
$ tapyrus-cli createpstt "[]" "[]" 0 true true
error: Expected type array, got string
```

That makes every `HelpExampleCli` line this PR adds non-functional as written. The Phase 1–3 RPCs are missing too, so this PR is the natural place to add the whole set:

```cpp
{ "createpstt", 0, "inputs" },
{ "createpstt", 1, "outputs" },
{ "createpstt", 2, "fallback_locktime" },
{ "createpstt", 3, "inputs_modifiable" },
{ "createpstt", 4, "outputs_modifiable" },
{ "createpstt", 5, "has_sighash_single" },
{ "converttopstt", 1, "permitsigdata" },
{ "converttopstt", 2, "inputs_modifiable" },
{ "converttopstt", 3, "outputs_modifiable" },
{ "addinputtopstt", 2, "output_index" },
{ "addinputtopstt", 3, "sequence" },
{ "addoutputtopstt", 1, "output" },
{ "addinputoutputpairtopstt", 2, "output_index" },
{ "addinputoutputpairtopstt", 3, "output" },
{ "addinputoutputpairtopstt", 4, "sequence" },
{ "finalizepsttconstruction", 1, "clear_inputs_modifiable" },
{ "finalizepsttconstruction", 2, "clear_outputs_modifiable" },
{ "walletcreatefundedpstt", 0, "inputs" },
{ "walletcreatefundedpstt", 1, "outputs" },
{ "walletcreatefundedpstt", 2, "fallback_locktime" },
{ "walletcreatefundedpstt", 3, "options" },
{ "walletcreatefundedpstt", 4, "bip32derivs" },
{ "walletcreatefundedpstt", 5, "inputs_modifiable" },
{ "walletcreatefundedpstt", 6, "outputs_modifiable" },
```

plus the Phase 1–3 ones (`combinepstt` 0, `finalizepstt` 1, `walletprocesspstt` 1/3, …). The functional test only exercises the Python RPC path, which sends typed JSON directly, so it cannot catch this.

### 2. `createpstt` / `walletcreatefundedpstt` help still documents `txid` / `vout`

Input parsing moved to `ParsePsttInputEntries`, which reads `previous_txid` / `output_index` (rawtransaction.cpp:481, 484), but the help text and CLI examples were not updated:

- src/rpc/rawtransaction.cpp:1900, 1907–1908, 1931
- src/wallet/rpcwallet.cpp:4420, 4428–4429, 4478

Copying the documented example verbatim fails with `previous_txid must be hexadecimal string`. rpc_pstt.py already uses the new names throughout, so only the docs drifted.

### 3. rpc_pstt.py picks Constructor UTXOs without filtering out colored coins

test/functional/rpc_pstt.py:318–320:

```python
unspent_list = self.nodes[0].listunspent()
assert len(unspent_list) >= 4, "test needs at least 4 spendable TPC UTXOs on node0 at this point"
unspent_a, unspent_b, unspent_c, unspent_d = unspent_list[0:4]
```

`listunspent` returns colored UTXOs unless filtered, and node0 holds colored change from earlier in the test (the 85-token output around line 118). If a colored UTXO lands in the first four, `unspent_a`/`unspent_b` become inputs to a PSTT whose only output is TPC, and the `sendrawtransaction` at line 374 fails on the color balance rule. The assertion message already says "TPC UTXOs" — make the code match, using the idiom the rest of the suite uses (feature_cp2sh_softfork.py):

```python
unspent_list = [u for u in self.nodes[0].listunspent() if u.get('token') == 'TPC']
```

### Nits

- src/wallet/rpcwallet.cpp:4537 — `FillPSTT(pwallet, pstt, 1, false, bip32derivs)` re-introduces the magic literal that #448 replaced in `walletupdatepstt`; please use `SIGHASH_ALL, /*sign=*/false` for consistency with rpcwallet.cpp:4149.
- src/rpc/rawtransaction.cpp:2087 and 2148 — the `pstt.tx_modifiable &&` half of the HAS_SIGHASH_SINGLE guard is unreachable; the preceding `if (!pstt.tx_modifiable || ...) throw` already guarantees the optional is engaged.
- `finalizepsttconstruction` always writes `tx_modifiable` even when the resulting mask is 0, while `createpstt` omits the field in that case (`if (modifiable != 0)`). Worth making the two agree so a fully frozen PSTT does not carry a redundant global.

### Looks good

- The global field reorder is mirrored exactly on both `Serialize` and `Unserialize`, with the required/optional split spelled out in comments.
- Sharing tip174_valid.json / tip174_invalid.json from test/functional/data via the CMake header generator, rather than copying them under src/test/data, is the right call — no drift between the unit and functional layers.
- The Constructor functional coverage (modifiable-flag enforcement, count increments, Has-SIGHASH_SINGLE paired-add, multi-party round trip, then actually broadcasting the result) goes beyond what rpc_psbt.py does.
- Routing `fallback_locktime` through `ConstructTransaction`'s locktime parameter to inherit the `nSequence = max-1` coupling, with the reasoning written down, is a good catch.

azuchi:
Re-reviewed at `e93cd1d`. Point 1 is resolved: I checked the 30 new `vRPCConvertParams` rows against every PSTT `argNames` entry and the set is complete and correctly named, with the string-only arguments (`decodepstt`, `extractpstt`, `walletsignpstt`, and `hexstring`/`sighashtype`/`sigscheme`) correctly left out. The two CLI-path regression tests are more than I asked for.

Points 2 and 3 and the three nits are untouched — the line numbers below are unchanged from the previous review.

### Still open

**2. `createpstt` / `walletcreatefundedpstt` help still documents `txid` / `vout`**

`ParsePsttInputEntries` still reads `previous_txid` / `output_index` (`src/rpc/rawtransaction.cpp:480,482`), so copying the documented example verbatim still fails.

- `src/rpc/rawtransaction.cpp:1900, 1907-1908, 1931`
- `src/wallet/rpcwallet.cpp:4420, 4428-4429, 4478`

**3. `rpc_pstt.py` selects UTXOs without filtering out colored coins**

Three unfiltered positional `listunspent()` calls remain — `test/functional/rpc_pstt.py:133`, `:250`, `:318` — while node0 still holds the 90- and 85-token colored change created earlier in the test.

**Nits**

- `src/wallet/rpcwallet.cpp:4537` — `FillPSTT(pwallet, pstt, 1, false, bip32derivs)`; please use `SIGHASH_ALL, /*sign=*/false` as at `:4149`.
- `src/rpc/rawtransaction.cpp:2087` and `:2148` — the `pstt.tx_modifiable &&` half of the guard is redundant after the preceding early throw.
- `createpstt` (`:1985`) and `converttopstt` (`:2051`) omit `tx_modifiable` when the mask is 0, while `finalizepsttconstruction` (`:2255`) always writes it.

### On the new tests

**`rpc_pstt_params` cannot catch point 2 as written.** It calls `createpstt [] []` with an empty inputs array, so it never exercises an input object's field names — which is exactly the gap that let the help/code divergence survive. Adding one CLI-path case with a populated inputs array using `previous_txid`/`output_index` would guard the regression and double as executable documentation of the correct field names.

**The `BOOST_CHECK_THROW` assertions are not regression guards.** `CallRPC` translates the RPC's `JSONRPCError` into `std::runtime_error` (`src/test/rpc_tests.cpp:37-40`), so an unregistered argument's `RPCTypeCheck` failure surfaces as the same exception type as a JSON parse failure. Those assertions therefore pass with or without the `client.cpp` fix. No harm, but only the `BOOST_CHECK_NO_THROW` cases are actually guarding the table — worth knowing before relying on them.

**`CallPsttWalletRPCFromCli` does not mirror `CallRPC` in the way its comment claims** (`src/wallet/test/pstt_wallet_tests.cpp:78-91`). It omits `CallRPC`'s `catch (const UniValue&) { throw std::runtime_error(...); }` translation. The current assertions happen to pass because they all fail at the JSON parse stage, but any error raised by the RPC body itself would escape as a `UniValue`, slip past `BOOST_CHECK_THROW(..., std::runtime_error)`, and abort the case as an unknown exception rather than failing cleanly. Please add the same catch.

Naviabheeman:
item 2, createpstt / walletcreatefundedpstt help still documents txid / vout is already addressed in #451. All roc parameters and help text are updated to use txid/vout. So this item can be skipped in this review.

azuchi (approved):

PR #451, opened 2026-08-18
PSTT (TIP-174): Fee Provider workflow, edge-case coverage, legacy PSBT removal, and multi-party network test

Naviabheeman:
Finishes the PSTT (TIP-174) implementation on top of Phase 4
(Creator/Constructor RPCs): adds the Fee Provider role, closes out
remaining Phase 6 edge-case coverage, removes the legacy PSBT
(BIP-174) implementation now that PSTT fully replaces it, and adds an
end-to-end multi-party network test exercising the whole pipeline
together.

- Phase 5: Fee Provider workflow (walletfundpsttfee, both
  noninteractive and interactive coin-selection variants), covered by
  the new pstt_fee_provider.py.
- Phase 6 remainder: closes out the TIP-174 spec's §9d edge cases not
  yet covered by Phase 5's tests.
- Phase 7: removes the legacy PSBT (BIP-174) RPCs and supporting code
  (~1,000 lines across rawtransaction.cpp, sign.h/sign.cpp,
  keystore.cpp) now superseded by PSTT; drops rpc_psbt.py,
  data/rpc_psbt.json, and wallet/test/psbt_wallet_tests.cpp in favor
  of wallet/test/pstt_wallet_tests.cpp.
- pstt_network.py: new multi-party round-trip test -- six nodes each
  with a fixed pipeline role (Creator, three Contributors, Signer, Fee
  Provider; see doc/tapyrus/pstt.md). Ten colored coins move between
  peers across randomized multi-color rounds, exercising Creator ->
  Constructor -> Signer -> Combiner -> Finalizer -> Extractor -> Fee
  Provider together rather than RPC-by-RPC, plus tamper simulations
  for each PSTTSignResult error and a Combiner-level check that
  removing an already-signed input is rejected via
  HasSameIdentifierAs(). Registered for both the default and
  --scheme SCHNORR signature schemes.


azuchi (changes requested):
### 1. The PSBT removal is incomplete — several places still reference the deleted RPCs

- `src/rpc/client.cpp:93-106` still carries 14 `vRPCConvertParams` rows for `createpsbt` / `walletcreatefundedpsbt` / `walletprocesspsbt` / `combinepsbt` / `finalizepsbt` / `converttopsbt`, and still has **zero** PSTT rows (point 1 of my #450 review), including none for this PR's new `walletfundpsttfee` `fee_input` object. Net effect of Phase 7 as it stands: `tapyrus-cli` loses its only working partially-signed-transaction surface, because the PSTT replacements were never registered.
- `src/wallet/test/CMakeLists.txt:11` still lists the deleted `psbt_wallet_tests.cpp`. That file is dead (never `add_subdirectory`'d, so the build does not break), but leaving it in a PR whose job is removing PSBT is the wrong end state — update or delete it.
- `doc/bips.md:38` still states BIP 174 PSBT RPCs "are present as of v0.17.0". That line was added deliberately on the Tapyrus side (`1dbad82f6 Add BIP174 to list of implemented BIPs`), so this PR should retract it.
- `doc/tapyrus/fix_transaction_malleability.md:71` still has a `createpsbt, walletcreatefundedpsbt` row. Since the PSTT RPCs now use the same `txid`/`vout` vocabulary, it maps over directly.
- `doc/release-notes.md:190-258` is a whole PSBT feature section. That text is inherited from Bitcoin Core 0.17, so the call on it is yours.

### 2. `walletfundpsttfee` interactive mode has no Has-SIGHASH_SINGLE guard

The noninteractive branch guards it explicitly (rpcwallet.cpp:4370-4372), with the right reasoning: a standalone input add is exactly what `addinputtopstt` refuses once a SIGHASH_SINGLE signature exists. The interactive branch (rpcwallet.cpp:4415-4417) only checks Outputs-Modifiable.

The implicit justification is "interactive adds an input *and* an output, so the pairing is preserved" — but that is not guaranteed. When `FundTransaction`'s change is dust it is dropped into the fee instead (`wallet.cpp:2885-2888`: `mapChangePosInOut[colorId] = -1; nFeeRet += nChange;`), and an exact-amount selection produces no change either. In those cases `rawTx.vout.size() == original_output_count`, no output is appended, and interactive mode performs precisely the standalone input add that noninteractive rejects. Either apply the same guard in both branches, or assert an output was actually added before allowing it under Has-SIGHASH_SINGLE.

### 3. The cited spec rules do not exist in the doc they cite

`rule 22` / `25` / `31` / `32` are cited in eight places (`src/wallet/pstt.h:33,40`, `src/wallet/rpcwallet.cpp:4090`, `src/wallet/test/pstt_wallet_tests.cpp:316,374,384`, `test/functional/p2p_pstt.py:558`, `test/functional/wallet_pstt.py:91`), and `src/wallet/pstt.h:33` names `doc/tapyrus/pstt.md` as the source — but that document has no numbered rules at all.

Related: pstt.md's Signer bullet (lines 163-170) does not describe the post-sign `PSTT_GLOBAL_TX_MODIFIABLE` mutation this PR implements, and `walletfundpsttfee` is never named in the doc even though the Fee Provider workflow section describes exactly what it does. Please either add these to pstt.md or cite TIP-174 itself.

### 4. Still open from my #450 review

- `src/rpc/client.cpp` PSTT rows (above).
- `test/functional/rpc_pstt.py:318-320` still takes `listunspent()[0:4]` without filtering colored coins, on a node that holds colored change.
- The `FillPSTT(..., 1, false, ...)` magic literal is now at two sites: `rpcwallet.cpp:4297` (walletcreatefundedpstt) and `4483` (walletfundpsttfee).
- The unreachable `pstt.tx_modifiable &&` half of the Has-SIGHASH_SINGLE guard is now at `rawtransaction.cpp:1606` and `1667`.
- Writing `tx_modifiable` even when the mask is 0 now happens in `ApplyPsttPostSignModifiableRules` too, alongside `finalizepsttconstruction`, while `createpstt` still omits the field in that case.

Moving the whole RPC surface to `txid`/`vout` here resolves point 2 of that review, but #450 on its own is still broken — it needs the same fix, or the two need to land together.

### Nits

- The `GetKeyForDestination` `CColorKeyID` handling in `src/keystore.cpp` (commit 8afe9239) is a real user-visible fix for `dumpprivkey` / `getaddressinfo` on colored addresses, but it has no regression test and is not mentioned in the PR title or release notes.
- `test_runner.py` placement: `p2p_pstt.py` (six nodes, multi-round) sits under "Tests less than 60s" and `wallet_pstt.py` under "less than 30s". Worth measuring and re-placing.
- `wallet_pstt.py` is the only one of the three PSTT functional tests with no `--scheme SCHNORR` variant registered.
- `p2p_pstt.py` seeds its RNG (`random.Random(1174)`) but generates keys with `os.urandom(32)` (lines 238-239), so a failure is not fully reproducible from the seed alone.

azuchi:
#450 is merged, so this branch needs a rebase. Notes on the conflict and what's still open, checked against `7bed8c7`.

## The conflict, and how to resolve it

#450 was **squash-merged** — `c348ed10` on master has a single parent (`baf6175`), and none of #450's individual commits are ancestors of master. This branch was rebased onto #450's *branch* head (`6f8603f1`), so it still carries its own copies of those commits. Every line #450 touched now exists twice, once inside master's squashed commit and once in the inherited copies, which is what GitHub is reporting as `CONFLICTING`.

Replaying only this branch's own commits onto master resolves it:

```
git fetch origin
git rebase --onto origin/master 6f8603f11
```

I ran exactly this locally: **9 commits, zero conflicts**, and the resulting tree is byte-identical to `7bed8c7`'s. `6f8603f11` is the boundary — it is the last commit belonging to #450, so everything after it is this branch's own work.

Please don't resolve this by merging master in or by re-applying #450's commits by hand; the `--onto` form is what drops the duplicated history cleanly.

## Two things this rebase settles

**`7bed8c7` (reverting `ParsePsttInputEntries` to `previous_txid`/`output_index`) was the right call.** Without it, `src/test/rpc_tests.cpp:114` — which #450 added at my request, specifically to catch the help/parser divergence — would have sent `previous_txid`/`output_index` into a parser expecting `txid`/`vout`, and `ParseHashO` → `ParseHashV` would have thrown `must be hexadecimal string (not '')` (`src/rpc/server.cpp:135`), failing `BOOST_CHECK_NO_THROW` and then throwing again on `r.get_str()`. Git would not have flagged that file at all, since it only exists on one side.

**The `KeyError: 'tx_modifiable'` CI failure is fixed by the rebase alone.** #450's merged `decodepstt` change now emits the object unconditionally via `value_or(0)`, so `rpc_pstt.py:373-374`, `wallet_pstt.py:95` and `wallet_pstt.py:163-164` all pass with no test edits. Please don't patch those assertions — the omit-when-zero encoding on the wire stays as it is, and only the RPC's diagnostic output is normalised.

## One question on naming

After `7bed8c7` the two vocabularies coexist:

- `createpstt` / `walletcreatefundedpstt` inputs-array object keys → `previous_txid` / `output_index` (`ParsePsttInputEntries`, rawtransaction.cpp:479-481)
- `addinputtopstt` / `addinputoutputpairtopstt` positional arguments → `txid` / `vout` (argNames at :2186/:2188, `client.cpp:102,105`, help text and error strings)

Each RPC is internally consistent, so nothing is broken. But they sit side by side in the same test:

```python
# rpc_pstt.py:221
createpstt([{"previous_txid": txid1, "output_index": vout1}, ...], ...)
# rpc_pstt.py:339
addinputtopstt(bare, unspent_a['txid'], unspent_a['vout'])
```

Naming the same two concepts differently within one feature is worth a deliberate decision rather than an artifact of the back-and-forth. If the split is intentional (PSTT's own field vocabulary for wire-level keys, the familiar `txid`/`vout` for positional arguments), a sentence in `doc/tapyrus/pstt.md` saying so would save the next reader the trip through both parsers. Also, `rpc_tests.cpp:133` and `:136` still describe these as `output_index(2)`, which no longer matches the `vout` entries registered in `client.cpp`.

## Still open from my last review

**`walletfundpsttfee` interactive mode: the Has-SIGHASH_SINGLE guard misses the multi-input case** (`src/wallet/rpcwallet.cpp:4463`).

```cpp
if (rawTx.vin.size() > original_input_count && rawTx.vout.size() == original_output_count && ...)
```

This only fires when no output was added at all. Coin selection pushes one `CTxIn` per selected coin (`src/wallet/wallet.cpp:2926`, `:3047`), so two inputs plus one change output slips through and adds unpaired inputs to a SIGHASH_SINGLE-marked PSTT. `addinputoutputpairtopstt` establishes the invariant as strictly 1:1, so the condition should compare the deltas:

```cpp
(rawTx.vin.size() - original_input_count) != (rawTx.vout.size() - original_output_count)
```

The same single-input assumption is baked into `doc/tapyrus/pstt.md`'s new "adds a TPC input" wording and the comment at `rpcwallet.cpp:4489` ("Both an input and an output were just added together").

**`test/functional/wallet_pstt.py:81` still calls `listunspent()[0]` unfiltered.** It is safe today only because the colored transaction has not been broadcast at that point. The comment directly above it is about colour being verified from the scriptPubKey, and the other two sites in `rpc_pstt.py` were filtered for exactly this reason — worth making this one consistent so a future reordering doesn't turn it into a confusing "fee_input must reference a TPC output" failure.

(`rpc_pstt.py:134`'s unfiltered call is fine as-is — the out-of-range prevout rejection happens before any colour check.)

Naviabheeman:
>After 7bed8c7 the two vocabularies coexist:

>createpstt / walletcreatefundedpstt inputs-array object keys → previous_txid / output_index (ParsePsttInputEntries, rawtransaction.cpp:479-481)
addinputtopstt / addinputoutputpairtopstt positional arguments → txid / vout (argNames at :2186/:2188, client.cpp:102,105, help text and error strings)

This shouldn't have happened. I had updated everything to txid/vout in this branch. But later rebases changed them. ITs cleaned up now.

azuchi (approved):

PR #452, opened 2026-08-25
Bump libXau to 1.0.12, update dependencies.md

Naviabheeman:
Daily CI failed for 4 straight days on all Linux GUI jobs because xorg.freedesktop.org is intermittently unreachable while depends downloads xcb_proto/xproto/libXau.  These packages are upgraded
- libXau-1.0.12.tar.gz (this branch)
-  fontconfig-2.16.0.tar.xz
- ibxkbcommon-1.13.2.tar.gz
- expat-2.8.3.tar.xz
- systemtap-5.5.tar.gz

Also replaces doc/dependencies.md's version table with one listing the exact file name, sha256, and download URL depends expects for each package, and documents the FALLBACK_DOWNLOAD_PATH S3 mirror (depends/Makefile) that depends falls back to.


azuchi:
This does not fix the failure described in the PR description. Details below.

## 1. The package that actually fails is `xcb_proto`, not `libXau`

From the last master daily-test run (32764348553, job 97550293464, `L-ARM-Deb-GUI+USDT+Wallet`):

```
Fetching xcb-proto-1.17.0.tar.gz from https://xorg.freedesktop.org/archive/individual/proto
curl: (28) Failed to connect to xorg.freedesktop.org port 443 after 30001 ms: Timeout was reached
    ... 4 attempts, all timed out ...
Fetching xcb-proto-1.17.0.tar.gz from https://s3-ap-northeast-1.amazonaws.com/repo.tapyrus.chaintope.com
curl: (22) The requested URL returned error: 403
make: *** [funcs.mk:340: .../.stamp_fetched-xcb_proto-xcb-proto-1.17.0.tar.gz.hash] Error 1
Depends build failed
```

The build dies at `xcb_proto` and never reaches `libXau`. On top of that, `libXau.mk`'s `$(package)_download_path` is still `https://xorg.freedesktop.org/releases/individual/lib/` after this change, so libXau's own reachability is unchanged too. Bumping the version does not affect either failure mode.

## 2. The S3 fallback mirror does not hold the xorg/xcb family

I fetched each file in `doc/dependencies.md` from the exact URL `depends` constructs (`$(FALLBACK_DOWNLOAD_PATH)/$(file_name)`, `depends/funcs.mk:39`). Present files return 200/206; these all return `403 AccessDenied`:

| File | Primary host |
| --- | --- |
| `xcb-proto-1.17.0.tar.gz` | xorg.freedesktop.org |
| `xproto-7.0.31.tar.gz` | xorg.freedesktop.org |
| `libXau-1.0.9.tar.bz2` **and** `libXau-1.0.12.tar.gz` | xorg.freedesktop.org |
| `libxcb-1.17.0.tar.gz` | xcb.freedesktop.org |
| `xcb-util-0.4.1.tar.gz` | xcb.freedesktop.org |
| `xcb-util-cursor-0.1.6.tar.gz` | xcb.freedesktop.org |
| `xcb-util-image-0.4.1.tar.gz` | xcb.freedesktop.org |
| `xcb-util-keysyms-0.4.1.tar.gz` | xcb.freedesktop.org |
| `xcb-util-renderutil-0.3.10.tar.gz` | xcb.freedesktop.org |
| `xcb-util-wm-0.4.2.tar.gz` | xcb.freedesktop.org |

Also missing: `db-4.8.30.NC.tar.gz`, `libevent-2.1.12-stable.tar.gz`, `miniupnpc-2.3.3.tar.gz`, `systemtap-4.7.tar.gz`, `zeromq-4.3.5.tar.gz`, and the three Qt CMake files. Only 9 of the 25 files are actually served: boost, expat, fontconfig, freetype, qrencode, qtbase, qttranslations, qttools, libxkbcommon.

The mechanism itself works — in the same failing job, `fontconfig` timed out on `www.freedesktop.org` and then recovered from the mirror, because that object exists. So the fix for the CI breakage is to upload the missing tarballs, starting with `xcb-proto-1.17.0.tar.gz` and `xproto-7.0.31.tar.gz`. Uploading only `xcb_proto` just moves the failure to the next package in the list.

Two notes on placement:

- The fallback URL is a flat concatenation, so a copy under a subdirectory would never be found. I checked `depends-sources/`, `depends/` and `sources/` — all 403.
- The mirror must carry whichever filename ends up pinned. If libXau is bumped here, the object to upload is `libXau-1.0.12.tar.gz`, not `libXau-1.0.9.tar.bz2`.

(The bucket denies `ListBucket`, so a 403 cannot strictly distinguish "absent" from "present but not public". Either way `depends` fails identically, and the required action is the same: publish the objects.)

Because of this, the new sentence in `doc/dependencies.md` —

> The mirror is flat (no subdirectories) and must serve each file under the exact "File name" listed in the table above.

— reads as a description of the current state but is aspirational. As written, a reader would conclude the fallback covers every package.

## 3. The green CI on this PR never built libXau 1.0.12

From this PR's own run (32825543191, job 97732592830):

```
Cache restored from key: depends-v6-qt6-linux-arm64-fc9ba5cc6dd512d0f2e518c7ed726f5a587df3c275ba7e03e811d8ade8006d5f
...
Fetching libXau-1.0.9.tar.bz2 from https://xorg.freedesktop.org/releases/individual/lib/
.../depends/work/build/x86_64-pc-linux-gnu/libXau/1.0.9-503c523e56a
```

The string `1.0.12` does not appear anywhere in that job log.

The cause is that the depends cache path is the whole `depends/` directory (`.github/workflows/daily-test.yml:172` and `:193`), which contains the version-controlled `depends/packages/*.mk`. The restore step runs after checkout and overwrites the modified `libXau.mk` with the cached copy. Since the exact key changed (it is derived from `hashFiles('depends/packages/*', ...)`), the `restore-keys` prefix matched an older cache and reverted the change.

That is a pre-existing CI defect rather than something this PR introduced, but it means:

- 14/14 green here is not evidence that 1.0.12 builds; nothing verified this change.
- The run is green only because `xorg.freedesktop.org` happened to be reachable that time — `xcb-proto` and `xproto` were fetched from it successfully at 08:19:36 and 08:19:38.
- Any future `depends/` change is silently reverted in CI whenever a prefix-matched cache restores.

Please fix the cache scope (exclude `depends/packages` from the cached path, or drop the `restore-keys` prefix for depends), or show a local verification, before merging the bump.

Given all three points, I would suggest splitting this: upload the missing objects first (that alone stops the red builds), fix the cache scope second, and land the libXau bump separately once it can actually be verified.

## 4. `doc/dependencies.md`

- **Dropping the "Minimum required" column loses real information.** The build enforces minimums: `CMakeLists.txt:366` `find_package(Libevent 2.1.8 MODULE REQUIRED)`, `cmake/module/AddBoostIfNeeded.cmake:85` `find_package(Boost 1.73.0 REQUIRED)`, `CMakeLists.txt:124` `find_package(BerkeleyDB 4.8 MODULE REQUIRED)`. The old values (Boost 1.55.0, libevent 2.0.22) were stale, so please correct the column rather than delete it.
- **The "(Linux only)" labels are inconsistent.** Per `depends/packages/packages.mk`, `qt_linux_packages` also contains `expat`, `freetype`, `fontconfig`, `libXau`, `xproto` and `xcb_proto`, but those six rows are unlabelled while their siblings from the same list are labelled. The other conditionals are not represented either: `bdb` is wallet-only, `miniupnpc` is upnp-only, `systemtap` is USDT + Linux only.
- **Two download targets are missing from the table.** The three Qt CMake files appear only in prose, with no filename/hash/URL row, even though the table claims to list what `depends` expects for each package. The macOS SDK is absent entirely; `daily-test.yml:238-240` downloads it from bitcoincore.org and falls back to `https://s3.ap-northeast-1.amazonaws.com/repo.tapyrus.chaintope.com` — note the different host spelling (`s3.` rather than `s3-`) from the one documented here.
- **The rationale for not mirroring the Qt CMake files is backwards.** `qt_details.mk:19` pins a sha256 against a mutable branch ref (`refs/heads/6.10.1`). That is precisely the case where a mirror helps: if upstream changes the file, the build fails on a hash mismatch with no fallback. Also, "intentionally" asserts an intent that is not recorded anywhere. Please state the fact without the rationale, or fix the underlying pin.
- The table now duplicates the sha256 and URL data that lives in `depends/packages/*.mk`, so it will drift. Consider generating it or linking to the `.mk` files instead. `depends/README.md:90` already documents `FALLBACK_DOWNLOAD_PATH`.

## Nits

- `libXau.mk` carries a comment asking the updater to check the default of `--disable-xthreads`. For the record, 1.0.12's `configure.ac:56-59` still defaults `xthreads=yes`, so no change is needed — but since the comment explicitly asks, please note the check in the PR description.
- The branch is named `fix/depends-xorg-mirror`, but nothing about the mirror is changed.

Naviabheeman:
Thanks for the review — addressed below.

"Minimum required" column dropped — This was a deliberate call, not an oversight: keeping a hand-maintained minimum alongside the pinned build version is exactly what let it go stale in the first place (the old Boost 1.55.0 / libevent 2.0.22 values you caught). We're keeping just the "Version" column — the version actually built and tested — as the single source of truth going forward.

"(Linux only)" labels inconsistent — Rather than complete the labeling (wallet-only, upnp-only, USDT+Linux-only, etc.), we removed all platform/feature-conditional labels from the table. That information already lives in the build-*.md guides per platform; keeping it in two places was how it drifted out of sync here.

Two download targets missing —
- macOS SDK: added as a table row (Xcode-26.2-17C52-extracted-SDK-with-libcxx-headers.tar.gz, sourced from depends/hosts/darwin.mk + daily-test.yml), noted as unhashed since the workflow itself doesn't verify a checksum either.
- Qt CMake files: intentionally left out of the table. They're files living in the qt/qt5 GitHub repo, not a standalone release artifact — there's no independent "download URL/hash for this artifact" to document in the same sense as the other rows, and no mirror to speak of.

Backwards Qt-mirroring rationale — Rewritten to state the fact plainly (not mirrored) without the "intentional" framing, since as you noted a mutable branch-ref pin is the opposite of "no fallback needed."

Table duplicates .mk data / will drift — Added an explicit note that depends/packages/*.mk and depends/hosts/darwin.mk are the source of truth and this table is a convenience copy. Also removed the FALLBACK_DOWNLOAD_PATH mechanism re-explanation and pointed to depends/README.md's existing docs instead of duplicating it.

One more from cross-checking the build guides while in there: doc/build-unix.md's Ubuntu/Debian section had zeromq-devel (the Fedora/RHEL package name) instead of libzmq3-dev — fixed.

And: standardized the S3 fallback bucket hostname on one spelling (s3.ap-northeast-1.amazonaws.com) in depends/Makefile, matching what daily-test.yml's macOS SDK fallback already used, rather than documenting the inconsistency and leaving both in place.

azuchi:
Re-reviewed at `a01ac8c`.

The mirror gap from my earlier comment is resolved — I re-probed the bucket and every file now returns 200/206, old and new file names alike (`xcb-proto`, `xproto`, `libXau`, `libxcb`, all six `xcb-util-*`, bdb, libevent, miniupnpc, zeromq, plus the four new tarballs this PR pins). Only the three Qt CMake files are still absent, which matches what the doc now says. The `depends/description.md` note about keeping the mirror in sync with `packages/*.mk` is the right place for that rule.

## Blocking: CI built none of the five version bumps

All 20 jobs are green, but from job 98144416588's log:

| Package | New version occurrences | Old version occurrences |
| --- | --- | --- |
| libXau | `1.0.12` — **0** | `1.0.9` — 3 |
| fontconfig | `2.16.0` — **0** | `2.12.6` — 3 |
| libxkbcommon | `1.13.2` — **0** | `1.7.0` — 3 |
| expat | `2.8.3` — **0** | `2.4.8` — 7 |
| systemtap | `5.5` — **0** | `4.7` — 3 |

This is the depends-cache scope problem from my earlier comment, still unfixed. The log shows `Cache restored from key: depends-v6-qt6-linux-arm64-fc9ba5cc6dd512d0f2e518c7ed726f5a587df3c275ba7e03e811d8ade8006d5f` — the same stale key as the previous run — and the cached `depends/` directory is extracted over the checkout, reverting the version-controlled files under it.

The clearest proof is this line:

```
Fetching qrencode-4.1.1.tar.bz2 from https://s3-ap-northeast-1.amazonaws.com/repo.tapyrus.chaintope.com
```

This PR changes `depends/Makefile:27` to `s3.ap-northeast-1`, yet CI used the old spelling. So it is not just the five `.mk` files being reverted — `depends/Makefile` and `depends/patches/fontconfig/gperf_header_regen.patch` go with them. The meson 1.12.0 and gperf that the workflow now installs were never used for their intended purpose either: no libxkbcommon 1.13.2 Meson build and no fontconfig 2.16.0 gperf regeneration ever ran.

That leaves five version bumps, a patch rebase, a new build-system version requirement, and a new build dependency all completely unverified. This is a pre-existing CI defect rather than something this PR introduced, but it has to be fixed (exclude `depends/packages`, `depends/patches` and `depends/Makefile` from the cached path, or drop the `restore-keys` prefix for depends) before any of this can be trusted.

Since CI didn't check these, I verified what I could locally, and all of it is fine:

- All four new sha256 hashes match the real upstream files (downloaded and compared).
- `gperf_header_regen.patch` applies cleanly to fontconfig 2.16.0 (`patch -p1 --dry-run`, no errors). The rule really is at line 1027, matching the rebase note.
- `xkbcommon.org/download/` genuinely stops at 1.7.0 — 1.8.0 and everything after 404s there — and the GitHub releases carry no assets, so the tag archive is upstream's only remaining channel. 1.13.2 is the current stable (1.14.0-beta1 is next).
- The macOS SDK row matches `depends/hosts/darwin.mk:4-5`, and the `depends/README.md:84` reference is correct.

## Findings

**1. `libxkbcommon` now depends on a GitHub tag archive.** GitHub's auto-generated `archive/refs/tags/*.tar.gz` files are not guaranteed byte-stable; GitHub changed how they are produced in 2023 and invalidated pinned checksums across the ecosystem. As noted above, upstream leaves no alternative, so this isn't avoidable — but it makes libxkbcommon the only pinned package whose primary source has that property. Worth a line in the `.mk` saying so, and treating the mirrored copy as the dependable one.

**2. The Meson install is unpinned.** `sudo pip3 install --break-system-packages --no-cache-dir "meson>=1.4.0"` resolves to whatever is newest at run time (1.12.0 in this run). Everything under `depends` is sha256-pinned; leaving the build tool floating works against the stability this PR is trying to buy. Please pin an exact `meson==X.Y.Z`.

**3. The "Linux only" markers are gone entirely.** Master's table carried them in its last column, and an earlier revision of this PR carried them in the package name. My earlier point was that six packages (`expat`, `freetype`, `fontconfig`, `libXau`, `xproto`, `xcb_proto`) sit in the same `qt_linux_packages` list as their labelled siblings and so were inconsistently unlabelled. Dropping the information for every package instead loses it: nothing in the new table shows that these are conditional, nor that `bdb` is wallet-only, `miniupnpc` upnp-only, or `systemtap` USDT+Linux-only.

**4. The "Minimum required" column is still removed.** The build does enforce minimums: `CMakeLists.txt:366` `find_package(Libevent 2.1.8 MODULE REQUIRED)`, `cmake/module/AddBoostIfNeeded.cmake:85` `find_package(Boost 1.73.0 REQUIRED)`, `CMakeLists.txt:124` `find_package(BerkeleyDB 4.8 MODULE REQUIRED)`. The old values were stale, which is a reason to correct the column rather than delete it.

**5. `gperf` and Meson are missing from the local build instructions.** `doc/build-unix.md` gains the zeromq fix but says nothing about the new requirements. A developer building the GUI through `depends` now needs `gperf` and Meson >= 1.4.0. The note at the end of `dependencies.md` is addressed at CI; the build instructions need it too.

**6. The `zeromq-devel` → `libzmq3-dev` change is out of scope.** It is correct (`zeromq-devel` is the RPM name), but unrelated to this PR's purpose and would be easier to review on its own.

azuchi (changes requested):
Checked at `6dda0edb`. The Meson pin, the restored `Minimum required` column, the local-build docs for gperf/Meson, and dropping the fontconfig patch are all good, and this run genuinely built the five bumps — `depends/built/` shows `expat-2.8.3`, `fontconfig-2.16.0`, `libXau-1.0.12`, `libxkbcommon-1.13.2`, `systemtap-5.5`. (`meson==1.4.0` is exactly right: libxkbcommon 1.13.2's `meson.build:9` asks for `>= 1.4.0`.) Two things left before merge.

## 1. The new "Print built dependency versions" step can't detect what it documents

Both sides of the comparison read state that a cache restore has already rewritten:

- left: `depends/built/$HOST/*/*.tar.gz` — comes from the restored cache
- right: `depends/packages/*.mk` — the working tree, which `actions/cache/restore` extracted the cached `depends/` over

So when the reversion happens, both sides show the old versions and agree. The step reports "no mismatch" precisely in the case it was written to catch.

The `a01ac8cc` run is the concrete example: it built `libXau-1.0.9` while the branch pinned `1.0.12`, and it fetched from `https://s3-ap-northeast-1.amazonaws.com/...` even though this PR changes `depends/Makefile:27` to `s3.ap-northeast-1` — so `depends/Makefile` and `depends/packages/*.mk` were both reverted, not just the build artifacts. The step's comment attributes this to the `if [ ! -d $DEPENDS_PREFIX ]` build-skip, but that run did not skip: it logged `Depends verification failed - 8 files missing` and rebuilt, and still fetched the old versions.

There is also a permanent false positive. The `sed` extraction doesn't expand make variables, so the right-hand list prints

```
native_qt-$(qt_details_version)
qt-$(qt_details_version)
```

against `native_qt-6.10.1` / `qt-6.10.1` on the left. Two rows will mismatch on every single run.

A diagnostic that is silent in the failing case and noisy in the passing one is worse than none — it will be read as evidence that nothing is stale. Please either compare against the committed files rather than the working tree (`git show HEAD:depends/packages/<pkg>.mk`, or simply `git status --porcelain depends/` to detect that the restore dirtied the tree) and handle or skip the Qt rows, or drop the step.

For the record on why this run is green: the log says `Cache not found for input keys: depends-v6-qt6-linux-arm64-1eff2c1b…, depends-v6-qt6-linux-arm64-` — even the `restore-keys` prefix missed, so nothing was restored and the checkout survived. That is cache eviction, not a fix. Once a master run repopulates the cache, the silent reversion returns. Fixing the cache scope itself is a pre-existing CI issue and belongs in its own PR, not this one — but that makes it more important that the diagnostic left behind here actually works.

## 2. The "Linux only" markers are still gone

Master's table carries them in its last column; this PR's table has none (`grep -c "Linux only" doc/dependencies.md` → 0). That is a straight loss of information relative to master: nothing in the new table shows that these packages are conditional at all.

My earlier point was about consistency, not about the markers themselves — `expat`, `freetype`, `fontconfig`, `libXau`, `xproto` and `xcb_proto` sit in the same `qt_linux_packages` list (`depends/packages/packages.mk`) as the labelled `libxcb*` rows and were unlabelled. Please restore the markers with all of `qt_linux_packages` covered. Extending it to `bdb` (wallet only), `miniupnpc` (upnp only) and `systemtap` (USDT + Linux only) would be nice but isn't required.

## Not blocking

The libxkbcommon GitHub tag-archive note, the `zeromq-devel` → `libzmq3-dev` change being out of scope, and `gperf` sitting in `depends/README.md`'s `#### Common` apt line rather than a Linux-GUI-specific one — all fine to leave as they are or fold in if convenient.

azuchi:
Checked at `b65e171c`. Both blockers are resolved. Two corrections left before I flip the review.

## The version-print fix works — it caught the real thing on this very run

Reading the expected side from `git show HEAD:$mk` is the right fix, and the Qt indirection now resolves. I ran the extraction logic locally and `native_qt-6.10.1` / `qt-6.10.1` come out correctly, so the permanent false positive is gone.

Better than that, this run restored a stale cache (`Cache restored from key: depends-v6-qt6-linux-arm64-fc9ba5cc…`), so the step got tested against exactly the case it exists for. From job 98787303237:

| Package | `depends/built/` | git HEAD |
| --- | --- | --- |
| expat | **2.4.8** | 2.8.3 |
| fontconfig | **2.12.6** | 2.16.0 |
| libXau | **1.0.9** | 1.0.12 |
| libxkbcommon | **1.7.0** | 1.13.2 |
| systemtap | **4.7** | 5.5 |

The pre-fix version would have printed the old numbers on both sides and agreed. Worth noting what that also means: this run's 20/20 green did *not* build the five bumps either. The content is still verified — the `a01ac8cc`-era run at `6dda0edb` missed the cache entirely and built all five — but the cache-scope bug is clearly live and reproducible, which makes fixing it (separately) more urgent than it looked.

## The `Platform` column is missing the GUI gate

`depends/Makefile:112,117`:

```make
qt_packages_$(NO_QT) = $(qt_packages) $(qt_$(host_os)_packages) $(qt_$(host_arch)_$(host_os)_packages)
packages += $($(host_arch)_$(host_os)_packages) $($(host_os)_packages) $(qt_packages_) ...
```

With `NO_QT=1`, `qt_packages_` is empty, so `qrencode`, the three Qt rows, and the whole of `qt_linux_packages` drop out of the build. The table leaves `qrencode` and the Qt rows blank, and the legend says a blank cell means "the package is built the same way regardless of platform/feature flags" — which isn't true for any of them.

- `qrencode` and Qt (qtbase/qttranslations/qttools) → `GUI only`
- the xcb family, `expat`, `freetype`, `fontconfig`, `libXau`, `xproto`, `xcb_proto` → `Linux + GUI only` rather than plain `Linux only`

`depends/README.md` already states this correctly ("Neither is needed for a `NO_QT=1` build"); it's only the table that hasn't followed.

## `doc/build-unix.md`'s zeromq line is now duplicated with #453

#453 carries the same `zeromq-devel` → `libzmq3-dev` edit as its own commit (`e65e21fd4`), split out after the scope comment on this PR. Whichever merges second will conflict or no-op — please drop it here and let #453 own it.

## Optional

The step prints two 24-line lists and neither compares them nor fails. On this run it emitted a five-package divergence and the job still finished green, with nothing but the raw lists to notice it by. A `diff` of the two, plus a `::warning::` (or a hard failure) when they differ, would turn this from something a reader has to eyeball into something that announces itself. Not blocking, but it's cheap and this run is a good argument for it.

## Already folded in

`gperf` moving out of `depends/README.md`'s `#### Common` section into a Linux-GUI-specific one, and the note explaining why libxkbcommon's URL is a GitHub tag archive — both were listed as non-blocking last time, thanks for picking them up anyway.

azuchi (approved):

PR #453, opened 2026-08-27
Fix ARM64 runner mislabeling, add minimum-supported-build testing, version-scoped depends patches

Naviabheeman:
CI infrastructure :
- Fixes daily-test.yml's L-ARM-* jobs, which were silently building x86_64 on ubuntu-latest despite the label — now on ubuntu-24.04-arm with HOST read from uname -m directly.
- Adds a minimum-dependency-versions job to weekly-heavy-tests.yml: a from-scratch (no cache) build across Linux x86_64/arm64/macOS arm64 that verifies doc/dependencies.md's documented floors still build, including a version-print step that diffs built-vs-expected and warns on mismatch instead of requiring a human to eyeball two lists.

Minimum-supported-build:
Depends patches infrastructure :
- depends/patches/<pkg>/ now requires patches to live in a <version>/ subdirectory, resolved from the package's actual pinned version — enforced at build time. Migrates 8 packages (bdb, libevent, libxcb, qrencode, qt/native_qt, systemtap, zeromq, fontconfig) to the convention.
- Deletes 17 confirmed-dead orphaned patch files.
- Run the minimum-dependency-versions build, each library tested against the version set in doc/dependencies.md's "Minimum required" column. Legend updated to match this narrower meaning.
- Scopes weekly-heavy-tests.yml's push trigger to branches: [master] so it no longer double-fires alongside pull_request when pushing to a PR branch.
- Fixes two real bugs along the way: zeromq's patch-application order didn't match its $(package)_patches list (6 patches touch the same CMakeLists.txt sequentially); a naive loop for qt would have applied qttools_skip_dependencies.patch unconditionally instead of only when host==build.

Version bumps:
- boost 1.81.0→1.92.0, freetype 2.11.0→2.14.3, libevent 2.1.12-stable→2.1.13-stable.

bison fix:
- libxkbcommon's Meson build needs bison — Fixed in daily-test.yml, both GUI jobs in weekly-heavy-tests.yml, and depends/README.md.


azuchi:
Reviewed at `d2a69dc`. The ARM64 diagnosis checks out — I confirmed it in this PR's own run (job 98590779685): `Image: ubuntu-24.04-arm` and `Building depends for HOST=aarch64-linux-gnu`, where master's `L-ARM-*` entries got `x86_64-pc-linux-gnu` because HOST is derived from the runner's `uname -m` and the `arch:` label was never cross-checked against it. Deriving `env.ARCH` from `uname -m` instead of a hand-maintained label is the right fix; `smoke-test.yml` keeps its own `config.arch` and is untouched, which is consistent since it has a separate matrix. The HOST triples the new weekly job hardcodes (`aarch64-linux-gnu`, `arm64-apple-darwin`) match what daily-test.yml derives.

## The new 260-line job has never run

`weekly-heavy-tests.yml` triggers on `schedule` and `workflow_dispatch` only — there is no `pull_request` trigger, so `minimum-dependency-versions` was not executed by this PR's CI. The 20/20 green says nothing about it.

That leaves a lot riding on a first execution that will happen on a Sunday schedule against master:

- the `sed -i.bak` rewriting of `depends/packages/*.mk`, and the `sha256sum`/`shasum` fallback, both written for GNU+BSD portability but never exercised on either
- **whether Boost 1.73.0 even builds here**. It's a 2020 release, this repo compiles with `-std=c++20` under GCC 13, and `boost.mk`'s recipe was written against the 1.81 series. Same question for Libevent 2.1.8-stable.
- the whole path on ARM64 and macOS

Please trigger it once via `workflow_dispatch` against this branch before merging — that answers all of the above at once.

Worth stating explicitly in the job or the PR description: if it fails on Boost 1.73.0, the correct response is to raise `doc/dependencies.md`'s `Minimum required` to whatever actually builds, not to relax the job. Detecting exactly that is the job's purpose. `fail-fast: false` is set, which is right.

## Timeout risk

The job does a from-scratch depends build (no cache — deliberate and well justified in the comment), a full GUI+wallet+tests build, then unit, util and functional tests with `-j 1`. For scale, daily's `L-X86-Deb-GUI+USDT+Wallet` takes 2h50m. GitHub's hard cap is 6 hours, as `weekly-heavy-tests.yml:60-61` already notes, and no `timeout-minutes` is set on this job.

If the `-j 1` isn't load-bearing, `-j $CPU_COUNT` would buy a lot of headroom — the dependency-version question doesn't change with parallelism. Otherwise an explicit `timeout-minutes` would at least make the failure mode legible.

## A comment that contradicts this PR

`weekly-heavy-tests.yml:385-392`:

> `ubuntu-24.04-arm`, a native GH-hosted runner, not the x86_64 ubuntu-latest + arch:arm64 label the daily matrix's `L-ARM-*` entries use -- worth checking separately whether those are actually native

`186326572`, in this same PR, changes those entries to `ubuntu-24.04-arm`. The comment describes a state that no longer exists and asks the reader to investigate something this PR already fixed. Presumably an artifact of the commit order.

## `meson>=1.4.0` vs the pin in #452

The new job installs `pip3 install "meson>=1.4.0"`, which resolves to whatever is newest at run time. #452 pins `meson==1.4.0` for the same install in the same file, for the reason that a floating build tool works against CI stability. Whichever of the two merges second will leave one of the two call sites unpinned — worth matching here.

## The computed hashes are circular

`patch_min_version` downloads the tarball and derives the sha256 from that same download, then patches it into the `.mk`, so depends' integrity check verifies the file against a hash taken from the file. The comment is admirably explicit that this is trust-on-first-use, but it doesn't have to be: Boost 1.73.0 and Libevent 2.1.8-stable are fixed historical releases whose hashes never change, so they can simply be written into the script as constants. I confirmed both URLs are live (HTTP 206), so pinning the values is straightforward.

## Nits

- `depends/hosts/linux.mk:39`'s aarch64 block is commented `# AArch64 (ARM64) cross-compilation`, but from this PR onward it is used for native builds too. Green CI confirms `aarch64-linux-gnu-gcc` exists on a native arm64 Ubuntu runner, so only the comment is now wrong.
- #452 carries the same `zeromq-devel` → `libzmq3-dev` edit to `doc/build-unix.md`. Whichever merges second will conflict or no-op; dropping it from #452 is the cleaner split, since that's where it was out of scope.

azuchi:
Checked at `3eeea5d0`.

## The libevent failure is a limitation of the version swap, not a floor violation

All three `minimum dependency versions` jobs die at the same place:

```
Preprocessing libevent...
can't find file to patch at input line 5
|--- a/CMakeLists.txt
make: *** [funcs.mk:340: .../libevent/2.1.8-stable-1386c1b4dd6/.stamp_preprocessed] Error 1
```

`depends/packages/libevent.mk` applies two patches:

```make
$(package)_patches=cmake_fixups.patch
$(package)_patches += winver_fixup.patch
```

Both were written against 2.1.12-stable. The swap step replaces only `_version` and `_sha256_hash`, so those patches stay in place and don't apply to the 2.1.8 tarball.

This matters for how the failure gets read. The note added in this commit says:

> If this fails on a specific package's minimum version, that means the documented floor doesn't actually build against this repo's current code/toolchain -- the correct response is to raise doc/dependencies.md's "Minimum required" for that package to whatever version genuinely does build, not to relax or skip this job.

Following that here would produce the wrong fix: raising libevent's documented minimum to 2.1.12 when nothing has shown 2.1.8 to be unbuildable. The job's own matrix comment already applies the correct reasoning to Qt —

> Qt's own depends recipe (qt.mk, qt_details.mk, its bundled patches) was written against the currently-pinned Qt series and isn't verified compatible with a 6.0.0 source tree, so it's left pinned

— and libevent carries version-specific patches for exactly the same reason; it just wasn't covered. Either exclude it the way Qt and Berkeley DB are (with that reason recorded alongside them), or make the swap patch-aware.

More generally, the job needs to be able to distinguish "the documented floor is wrong" from "our depends recipe is pinned-version-specific", otherwise the next failure invites the same misreading. Even a `$(package)_patches=` line being non-empty for a swapped package would be enough to warn about up front.

Worth recording the part that did work: Boost 1.73.0 built through to `Caching boost...`, so that floor is valid at least at the depends layer. Boost carries no patches, which is precisely why it got that far.

(The other red check, `macOS ARM64`, is unrelated — `p2p_addr_relay.py:272`'s `assert_greater_than_or_equal(2, len(nodes_received_addr))` failing `2 < 3`, a probabilistic destination-rotation assertion. Nothing here touches P2P, the same suite passes on Linux and Windows in the same run, and master's smoke-test had an intermittent failure of its own on 2026-08-27. No action needed on that one.)

## Nit: `push` and `pull_request` both fire on a PR branch

```yaml
  push:
    paths:
      - '.github/workflows/weekly-heavy-tests.yml'
  pull_request:
    paths:
      - '.github/workflows/weekly-heavy-tests.yml'
```

Pushing to a PR branch that touches this file triggers both, and `concurrency: ${{ github.workflow }}-${{ github.ref }}` doesn't collapse them because the two events carry different refs (`refs/heads/<branch>` vs `refs/pull/N/merge`). For a job budgeted at `timeout-minutes: 330` across three runners that's a lot to duplicate. Adding `branches: [master]` under `push:` keeps the post-merge validation while leaving PRs to the `pull_request` trigger alone.

azuchi:
@Naviabheeman Fix conflicts.

Naviabheeman:
With the new commits, minimum version build can be executed successfully. The patching step in depends build is extended to be version aware, with separate patch directory according to library version.

azuchi:
Re-reviewed at `5ae7cf58`. The rebase onto the merged #452 is clean (merge-base `75082c425`, none of #452's commits duplicated), CI is 26/26, and everything I raised last time is addressed. Three small things left, none of them blocking.

## The libevent finding got fixed at the root

Rather than excluding libevent from the swap, the patches are now version-scoped and enforced at build time (`funcs.mk:43-45`):

```make
$(1)_patches_path?=$(PATCHES_PATH)/$(1)/$($(1)_version)
$(if $(wildcard $(dir $($(1)_patches_path))*.patch),$(error Patch file(s) found directly under ... -- patches must live in a version subdirectory))
```

with `libevent.mk` binding its patches to the version via `ifeq ($($(package)_version),2.1.12-stable)` and a generic `$(foreach patch,...)` preprocess step. Swapping the version in the minimum-versions job now re-applies exactly the right patches instead of forcing 2.1.12-era ones onto whatever tarball turns up. That's the distinction I asked for — "the documented floor is wrong" vs "our recipe is version-specific" — turned into something the build itself enforces. `qt_details_patches_path` is version-scoped too, so the `$(dir ...)` guard works correctly for the qt/native_qt pair that override `_patches_path`.

And the three jobs now pass for real:

```
Linux x86_64 - minimum dependency versions   pass  54m24s
Linux arm64  - minimum dependency versions   pass  48m44s
macOS arm64  - minimum dependency versions   pass  55m48s
```

so Boost 1.81.0, Expat 2.4.8, FreeType 2.11.0, libevent 2.1.12-stable and systemtap 4.7 are demonstrably buildable, and the runtimes land comfortably inside `timeout-minutes: 330` — the timeout concern from last time is settled by measurement.

Things I checked independently, all fine: the three new hashes match upstream (boost 1.92.0 downloaded in full at 235 MB, freetype 2.14.3, libevent 2.1.13-stable); and the eight deleted patch files were all dead — none appear in any `_patches` list on master, and `qt.mk`/`zeromq.mk` keep their lists unchanged, so the new `$(error)` rule forcing the cleanup didn't drop anything live.

## 1. The title no longer describes the PR

"Fix arm64 runner and build doc" now also covers Boost 1.81→1.92 (eleven minor versions), FreeType 2.11→2.14, libevent 2.1.12→2.1.13, and a full restructure of `depends/patches/` with a build-time rule enforcing the new layout. That's a dependency-upgrade and build-system change sitting under a title about ARM runners — worth retitling so the history reads honestly, even if the content stays together.

## 2. `63ec62886`'s subject says 8, the job tests 5

> CI: test 8 recently-bumped packages' pre-bump versions as documented minimums

The step's own comment says "each of the 5 packages" and lists Boost, Expat, FreeType, libevent and systemtap, with fontconfig, libXau and libxkbcommon excluded and the reason recorded. Eight is the number bumped, not the number given a tested floor.

## 3. `Minimum required` now means two different things in one document

In the first table (CMake/Clang/GCC/Python) it's the floor the build enforces — `cmake_minimum_required(VERSION 3.22)`, `find_package(Python3 3.10 ...)`. In the second it's the oldest version CI actively builds and tests, which the new paragraph is admirably explicit about, including that `find_package(Boost 1.73.0 REQUIRED)` and `find_package(Libevent 2.1.8 MODULE REQUIRED)` are older and untouched. The prose scopes it with "in the table below", so it isn't ambiguous on a careful read — but the same column heading carrying two definitions a few lines apart is easy to trip over. Renaming one of them (e.g. `Tested floor` for the package table) would remove the trap.

azuchi:
Re-reviewed at `a66ad0d3`.

## The bumped dependency versions are never built by CI

`boost 1.92.0`, `freetype 2.14.3` and `libevent 2.1.13-stable` are not compiled by any daily-test job. This PR's own diagnostic step reports it, on `L-ARM-Rel-GUI+Wallet` (same on `L-X86-Rel-GUI+USDT+Wallet` and `L-X86-Deb-GUI+USDT+Wallet`):

```
== depends/built/aarch64-linux-gnu: actually-built/cached artifact versions ==
boost-1.81.0 / expat-2.4.8 / fontconfig-2.12.6 / freetype-2.11.0 / libXau-1.0.9 / libevent-2.1.12-stable
== MISMATCH: depends/built/aarch64-linux-gnu doesn't match depends/packages/*.mk at HEAD ==
-boost-1.92.0 / -expat-2.8.3 / -fontconfig-2.16.0 / -freetype-2.14.3
```

Two mechanisms combine:

```yaml
- name: Save Depends cache
  if: github.ref == 'refs/heads/master'     # daily-test.yml:544-545
```
```bash
if  [ ! -d "$DEPENDS_PREFIX" ]; then        # daily-test.yml:419-425
  make -C depends $DEPENDS_OPTS -j$CPU_COUNT install_cmake
fi
```

The cache is only saved on master, so a PR's exact key can never exist and `restore-keys` always serves master's tree. That restore recreates `depends/<host>/`, and the shell guard then skips `make` entirely — before depends' build-id logic can notice a `.mk` version changed.

Master has the same problem. Its scheduled run on 2026-09-06 (`34057312203`) reports built `expat-2.4.8`/`libXau-1.0.9` against expected `expat-2.8.3`/`libXau-1.0.12`, so #452's merged bumps have never been built either — and master then saves that stale tree under the new hash key, so later runs get an exact-key hit on mislabelled content and the staleness sustains itself.

The defect predates this PR, but this is the PR bumping three more packages, so without a fix its content merges unverified. Dropping the `[ ! -d "$DEPENDS_PREFIX" ]` guard and running `make -C depends` unconditionally is enough — depends decides per package from its build ids, so an unchanged tree is a fast no-op and only the bumped packages rebuild. Removing `restore-keys` would also work but costs a full ~50-minute depends build every run.

The mismatch warning doubles as the check that the fix worked: it should stop firing.

## The three failing jobs

None look attributable to the diff, but two need a re-run before merge.

**`L-ARM-Rel-GUI+Wallet`** — `WalletTests::walletTests()` at `src/qt/test/wallettests.cpp:592`, `r.status` is 2 (`IssueTokenResult::RpcError`, `src/qt/walletmodel.h:166`) instead of `OK`, on the NFT `issueToken(3, 1)` step. This PR changes no `src/` file, master is green on this job on 09-04/05/06, this branch passed it on 09-02, and per the section above the binaries were built against the same dependency tree as those green runs. Looks flaky.

**`feature_pruning`** — gives up after 3h57m waiting for the first block file to be pruned (`wait_for_prune: usage=113, blk00000.dat still present`, five times). The mining rate degrades from ~18 to ~79 minutes per 2000 blocks within the run, which points at runner contention. (`abf3842a9`'s `pull_request` trigger is path-filtered to this workflow file, so it only gates PRs editing it — that part looks right.)

**`macOS x86_64`** — `brew install berkeley-db@4` fails and falls back to 18.1.40: #458, unrelated, with #459 open against it.

azuchi (changes requested):
Reviewed at `6595d35b`. The change is right and should stay, but it doesn't fix the problem — and that's on me: my previous diagnosis stopped one layer short.

`make -C depends` now runs unconditionally and really does run (`Depends verification failed - 8 files missing`, prefix removed, Qt rebuilt from scratch). But the mismatch is unchanged:

```
== depends/built/x86_64-pc-linux-gnu: actually-built/cached artifact versions ==
boost-1.81.0 / expat-2.4.8 / freetype-2.11.0 / libXau-1.0.9 / libevent-2.1.12-stable
== MISMATCH: depends/built/x86_64-pc-linux-gnu doesn't match depends/packages/*.mk at HEAD ==
```

Two lines from the same run say why:

```
Fetching freetype-2.11.0.tar.xz from https://download.savannah.gnu.org/releases/freetype
Fetching qrencode-4.1.1.tar.bz2 from https://s3-ap-northeast-1.amazonaws.com/repo.tapyrus.chaintope.com
```

It downloaded freetype **2.11.0**, which cannot happen if the `.mk` in effect said 2.14.3. And it used the `s3-ap-northeast-1` host form, which #452 replaced with `s3.ap-northeast-1` on master.

So the tree `make` reads is not the one that was checked out. The depends cache is stored and restored with `path: ${{ env.DEPENDS_DIR }}`, i.e. the whole `depends/` directory, so restoring it writes the cached copies of version-controlled files — `depends/packages/*.mk`, `depends/Makefile`, `funcs.mk`, `patches/` — straight over the checkout. `make` then faithfully follows the old recipes.

I previously pointed at the `[ ! -d "$DEPENDS_PREFIX" ]` guard as the cause. It was one necessary condition (the fix here is needed, since with the guard in place fixing the real cause still wouldn't rebuild), but the actual cause is the cache overwriting tracked sources. Sorry for sending you at the narrower target.

## What's still needed

Keep the version-controlled part of `depends/` out of the restore. Either:

1. **Narrow `path:`** to generated content only — `depends/built`, `depends/sources`, `depends/work`, and the per-host prefix directory — never `depends/packages`, `depends/patches`, `depends/hosts`, `depends/builders`, `depends/Makefile` or `depends/funcs.mk`. The `Save Depends cache` step at `daily-test.yml:544` uses the same `path:` and has to match.

2. **Or restore the tracked files after the cache step** with `git checkout -- depends/`, which resets tracked files to HEAD and leaves the untracked generated directories alone. One line, and it stays correct as files are added under `depends/` later.

Same verification as before, plus a positive signal this time: the mismatch warning should stop firing, and the log should show `boost_1_92_0.tar.gz` and `freetype-2.14.3` being fetched.

## Unrelated

`macOS ARM64` and `macOS x86_64` fail at `brew install berkeley-db@4` falling back to 18.1.40 — #458, with #459 open against it and already green on `berkeley-db@5`.

azuchi:
Reviewed at `34ae695c`. The root cause is fixed and the bumped versions are finally being built:

```
== depends/built/x86_64-pc-linux-gnu: actually-built/cached artifact versions ==
boost-1.92.0 / expat-2.8.3 / freetype-2.14.3 / libXau-1.0.12 / libevent-2.1.13-stable
== OK: depends/built/x86_64-pc-linux-gnu matches depends/packages/*.mk at HEAD ==
```

`9f7d4c76f` also covers something I missed: `git checkout` only restores tracked files' content, so a stale cache entry can still leave behind a file that has since been deleted — and this PR is exactly the change that makes that fatal, since a leftover flat patch under `depends/patches/<pkg>/` now trips the new `$(error)`. Scoping `git clean -fdx` to `patches/`, `packages/`, `hosts/` and `builders/` is the right call: per `depends/.gitignore`, everything generated (`SDKs/`, `work/`, `built/`, `sources/`, `config.site`, the per-host prefixes) lives outside those four, so `-x` has nothing legitimate to remove there.

`34ae695c7` checks out too — the 2.1.13 `cmake_fixups.patch` is byte-identical to the 2.1.12 one, and `libevent.mk` records why `winver_fixup.patch` isn't carried over (one hunk fails against `evutil.c`, and it's Windows-only, unrelated to the `cmake_minimum_required` failure being fixed). The macOS cross jobs are green, so it applies.

Two non-blocking things.

## 1. The mismatch warning is now mostly false positives

11 of the 14 daily-test jobs report `MISMATCH`; only the three `GUI+USDT+Wallet` Linux jobs report `OK`. The diffs aren't staleness — they're packages that the job's configuration never builds:

```
L-ARM-Rel-GUI+Wallet (USDT off):
-systemtap-5.5

M-ARM-Rel-GUI+Wallet (macOS):
-expat-2.8.3 -fontconfig-2.16.0 -freetype-2.14.3 -libXau-1.0.12
-libxcb-* -libxkbcommon-1.13.2 -systemtap-5.5 -xcb_proto-1.17.0 -xproto-7.0.31
```

The expected side enumerates every `.mk` unconditionally (`daily-test.yml:527`):

```bash
MK_FILES=$(git ls-tree -r --name-only HEAD -- depends/packages | grep '\.mk$')
```

with no regard for the gating in `depends/packages/packages.mk` — `qt_linux_packages`, `usdt_linux_packages`, `wallet_packages`, `upnp_packages`, and `NO_QT` in `depends/Makefile`. So any job that turns a feature off, or isn't Linux, mismatches by construction.

This check is what caught the real bug, so it's worth keeping sharp rather than leaving it crying wolf on 11 of 14 jobs. The direction that actually matters is "a package was built at a version other than HEAD's" — packages absent from `built/` are usually just not part of this configuration. Intersecting the expected list with the package names present in `built/` before diffing would do it, and keeps the check honest about what it can see.

## 2. `weekly-heavy-tests.yml` still has the same pattern

The `heavy-tests` job restores the depends cache exactly the same way and got neither fix:

```yaml
:99-103   actions/cache/restore@v5 with path: ${{ env.DEPENDS_DIR }}
          restore-keys: depends-v6-qt6-linux-x86_64-       # prefix, no content hash
:125-132  if [ -d "$DEPENDS_PREFIX" ] && [ -f ... libdb_cxx-4.8.a ]; then
            echo "Depends cache hit — skipping rebuild"
          else
            make -C depends ...
```

No `git checkout -- depends/` after the restore, and the build-skip guard is still there. The stakes are lower — these jobs aren't validating dependency versions — but it's the same latent bug, in a workflow this PR newly runs on pull requests.

`minimum-dependency-versions` is unaffected: it builds depends from scratch with no cache (`:585`), which is also why those three jobs are the only ones currently proving anything about a specific dependency version.

azuchi (approved):

PR #454, opened 2026-08-28
PSTT - Implement RPC gaps identified during a review comparing with Bitcoin Core's PSBT RPC surface

Naviabheeman:
1. No **analyzepsbt** equivalent. Bitcoin's analyzepsbt tells a caller, without them having to reason about it manually: which role should run next, per-input readiness, and — the part Tapyrus genuinely lacks — an estimated fee/vsize once every input has a UTXO attached. decodepstt already exposes tx_modifiable directly (arguably better than Bitcoin's PSBT, which has no wire-level modifiable bit at all and has to infer it), but it computes no size/fee estimate. A wallet UI building on PSTT has no way to ask "what would this cost if extracted right now?" without reconstructing the transaction itself. Suggest: add analyzepstt (or extend decodepstt) with per-input has_utxo/is_final, an overall next role hint, and an estimated-fee/vsize once computable.

2. No **non-wallet Updater** (utxoupdatepsbt equivalent). walletupdatepstt requires a loaded wallet. Since Tapyrus has no witness-UTXO shortcut — every input needs the full previous transaction attached to PSTT_IN_UTXO — this gap arguably matters more here than in Bitcoin: a coordinator, hardware-wallet bridge, or watch-only tool with no wallet loaded has no way to attach known previous transactions to a PSTT at all. Suggest: a updatepstt "pstt" ["rawtx",...] RPC that fills PSTT_IN_UTXO from caller-supplied raw transactions (or the node's own tx index) purely by matching previous_txid, no wallet required.

3. No **joinpsbts** equivalent. Given PSTT's whole point is the constructable (BIP-370) model specifically because it makes composition easier than BIP-174's embedded-global-tx shape, this is a case where PSTT is better positioned to support it than old-style PSBT ever was — and the hard part (locktime-kind intersection across inputs) is already implemented and reusable (ComputeLocktime). Genuinely low-hanging: merge N independent PSTTs' inputs/outputs into one, re-run ComputeLocktime over the union, reject if the intersection is empty. Suggest: joinpstts ["pstt",...].

Suggested test scenarios (regardless of whether the RPC gaps above get filled)

- Fee/size estimation (once added): a PSTT with a subset of inputs UTXO-attached should report "not yet estimable"; fully attached should match extractpstt's actual materialized size within tolerance.
- PSTT_GLOBAL_TX_MODIFIABLE bit-transition races: two independent Signer rounds (different parties) each closing a different bit concurrently, then combinepstt — confirm the merged modifiable state is the intersection/union rule you'd expect, not last-writer-wins.
- Large-PSTT stress: hundreds of inputs/outputs, many-preimage inputs (RIPEMD160/SHA256/HASH160/HASH256 all populated) — confirms the TLV parser doesn't have an accidental O(n²) (the unknown map dedup and xpub canonical-bytes dedup are the two spots I'd specifically check under load).
- combinepstt idempotency: merging a PSTT with itself, or merging A into B then B into A, should be a no-op/order-independent for every field-merge tier (refuse/must-match/pick-first) — currently only tested pairwise-conflict, not commutativity.
- Cross-tool interop: if any other TIP-174 implementation exists (even a reference/test one from the chaintope/tips repo), a round-trip test — decode there, re-encode here, byte-identical — would catch subtle serialization-order assumptions the current test suite can't, since it only ever round-trips through this one implementation.
- Once updatepstt/joinpstts land: the obvious construction/rejection pairs, same style as everything already in pstt_tests.cpp and rpc_pstt.py.

### DIFF
--- a/doc/tapyrus/pstt.md
+++ b/doc/tapyrus/pstt.md
@@ -227,18 +227,22 @@ construction across separate parties, not a single call).
 Inspecting a PSTT
 -----------------
 
-`decodepstt` reports, alongside the raw field breakdown, which role above
-should run next for this PSTT (`"next"`: `constructor`/`updater`/`signer`/
-`finalizer`/`extractor`), determined by the same completeness check
-`finalizepstt` itself uses (a dry-run `SignPSTTInput` against a keyless
-provider — this can only ever confirm signatures *already collected* are
-enough, never fabricate new ones). Once every input carries `PSTT_IN_UTXO`,
-it also reports `"estimated_size"` (bytes) and `"estimated_fee"` (TPC only,
-per the Fee Provider note above) for the transaction this PSTT would
-currently extract to — using each unfinalized input's already-known redeem
+`decodepstt` reports, alongside the raw field breakdown, `"next"`: a JSON
+array of every role above that could act on this PSTT right now, in ladder
+order (`constructor`/`updater`/`signer`/`finalizer`/`extractor`) — these are
+independent, not mutually exclusive, e.g. a still-modifiable but
+already-fully-funded PSTT reports both `constructor` and `signer`. Role
+completeness is determined by the same check `finalizepstt` itself uses (a
+dry-run `SignPSTTInput` against a keyless provider — this can only ever
+confirm signatures *already collected* are enough, never fabricate new
+ones). Once every input carries `PSTT_IN_UTXO`, it also reports
+`"estimated_size"` (bytes) for the transaction this PSTT would currently
+extract to — using each unfinalized input's already-known redeem
 script/public keys to size a plausible completion, real
-`PSTT_IN_FINAL_SCRIPTSIG` where already present. Both estimates can still
-change while `"next"` is `constructor` or `signer`.
+`PSTT_IN_FINAL_SCRIPTSIG` where already present — and `"estimated_fee"`
+(TPC only, per the Fee Provider note above) alongside it, unless it would be
+negative, which just means this PSTT is still gathering inputs. Both can
+still change while `"next"` contains `constructor` or `signer`.
 
 ---
 
--- a/doc/tapyrus/pstt.md
+++ b/doc/tapyrus/pstt.md
@@ -90,7 +90,7 @@ Reserved and must be rejected, not silently treated as unknown: `0x01`
 from BIP-371 (`0x13`-`0x18`) — none have a Tapyrus counterpart.
 
 `PSTT_IN_PREVIOUS_TXID`/`PSTT_IN_OUTPUT_INDEX` are this wire format's own
-field names (mirrored by `PSTTInput`'s `previous_txid`/`prev_out_index`
+field names (mirrored by `PSTTInput`'s `previous_txid`/`prevout_index`
 members in C++), distinct from the RPC surface: every PSTT RPC that takes
 this pair as a user-facing argument (`createpstt`, `walletcreatefundedpstt`,
 `addinputtopstt`, `addinputoutputpairtopstt`) uses `txid`/`vout` uniformly,
@@ -161,9 +161,26 @@ construction across separate parties, not a single call).
 * **Constructor** — appends inputs/outputs to an existing PSTT. Only ever
   appends; never removes or reorders what is already present. Refuses to
   add an input whose required locktime would make the PSTT's already-signed
-  content locktime-invalid.
+  content locktime-invalid. `addinputtopstt`/`addoutputtopstt`/
+  `addinputoutputpairtopstt` do this one input/output at a time;
+  `joinpstt` instead concatenates the inputs and outputs of several
+  independently-built PSTTs in one call, for assembling a large PSTT (e.g.
+  one filling most of a block) out of pieces built separately without a
+  single caller having to drive every individual `addinputtopstt` call.
+  Every joined PSTT must still have both Inputs/Outputs Modifiable set, must
+  not have Has-`SIGHASH_SINGLE` set, and must carry no signature on any
+  input — joining shifts every subsequent piece's input/output positions,
+  which would silently invalidate a `SIGHASH_SINGLE` commitment to a
+  specific position; join first, add any `SIGHASH_SINGLE` input/output pair
+  afterwards. The same outpoint may not be spent by more than one of the
+  joined PSTTs.
 * **Updater** — attaches externally known data (UTXOs, redeem scripts,
-  BIP32 derivation paths) without altering which inputs/outputs exist. Must
+  BIP32 derivation paths) without altering which inputs/outputs exist.
+  `walletupdatepstt` sources this from a loaded wallet (also discovering
+  redeem scripts/BIP32 paths from it); `updatepstt` instead takes
+  caller-supplied raw transactions directly and only ever attaches
+  `PSTT_IN_UTXO` (no wallet, so no redeem script/BIP32 discovery), for
+  contexts with no wallet loaded at all. Must
   not change an input's sequence number once that input carries a
   signature, or once any other input carries a `SIGHASH_ALL`-without-
   `SIGHASH_ANYONECANPAY` signature (either case would silently invalidate
@@ -184,7 +201,10 @@ construction across separate parties, not a single call).
   inputs and outputs together, never one without the other. This only ever
   clears Inputs/Outputs Modifiable or sets Has-`SIGHASH_SINGLE` — it never
   reopens a bit an earlier signing round already closed.
-* **Combiner** — merges two or more PSTTs that share an identifier.
+* **Combiner** — merges two or more PSTTs that share an identifier (i.e.
+  alternate signing progress on the *same* underlying transaction — not to
+  be confused with `joinpstt`, which assembles *different* inputs/outputs
+  from multiple PSTTs into one new, larger transaction).
   Per-field conflict policy on a mismatch: `PSTT_IN_PARTIAL_SIG`,
   `PSTT_IN_FINAL_SCRIPTSIG`, and `PSTT_IN_UTXO` refuse (signature/UTXO
   disagreements are exactly the disagreements that must not be silently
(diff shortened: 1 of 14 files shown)


azuchi:
Reviewed at `fe52dcf`. The three gaps from the description are covered, `joinpstt`'s rejection guards are thorough, and `EstimatePsttFee` being TPC-only is the right call for colored coins. Findings below, with the test additions folded into each.

## 1. `joinpstt` can produce a PSTT that no longer decodes

`src/rpc/rawtransaction.cpp:2262`:

```cpp
for (const auto& entry : pstt.xpubs) joined.xpubs.push_back(entry);
```

There's no dedup, but `Unserialize` rejects a repeated `PSTT_GLOBAL_XPUB` keydata outright (`src/pstt.cpp`, `"Duplicate Key, PSTT_GLOBAL_XPUB already provided"`). So joining two PSTTs that share an xpub succeeds, and the result then fails to parse in `decodepstt`, `combinepstt`, or anything else that reads it back — in this same implementation. `IsSane()` doesn't look at xpubs, so nothing catches it in between.

`Merge()` already gets this right for the `combinepstt` path, deduping via `XpubEntryCanonicalBytes()` (`src/pstt.cpp:589-591`). That helper is `static` in `pstt.cpp`, so it needs exposing, or the dedup belongs alongside `Merge`'s.

Nothing in the tree currently sets global xpubs — only `joinpstt` copies them and `decodepstt` reads them — so this arrives via externally-produced PSTTs. Given PSTT is an interchange format and the description itself calls for cross-tool interop tests, that's the case that matters.

**Test:** join two PSTTs carrying the same xpub, round-trip the result through `DecodePSTT`. That's the regression guard.

## 2. `EstimatePsttFee` dereferences an unset `optional`

```cpp
for (const PSTTOutput& output : pstt.outputs) {
    if (GetColorIdFromScript(output.script).type == TokenTypes::NONE) out_tpc += *output.amount;
}
```

`amount` is `boost::optional<CAmount>` (`src/pstt.h:162`), and `boost::optional::operator*` on a disengaged optional is undefined behaviour. The only guard is `AllInputsHaveUtxo()`, which looks at inputs only.

Neighbouring code already defends this: `EstimatePsttSize` uses `output.amount.value_or(0)`, and `ComputePsttColorBalances` does `if (!output.amount || output.script.empty()) continue;`.

**Test:** decode a PSTT whose inputs all have UTXOs and one output has a script but no `PSTT_OUT_AMOUNT`.

## 3. `decodepstt` now throws on PSTTs it used to display

Both `EstimatePsttFee` and `EstimateInputScriptSig` do:

```cpp
input.utxo->vout.at(input.prevout_index)
```

`AllInputsHaveUtxo()` only checks that `input.utxo` is non-null, never that the index is in range. The range check lives in `IsSane()`, which `decodepstt` deliberately does not call — its own comment on `identification_txid` says so:

> decodepstt is an inspection tool and this is exactly the kind of malformed-but-parseable PSTT it should still be able to show

So a PSTT with a UTXO attached and an out-of-range `prevout_index` now raises `std::out_of_range` where it previously rendered fine. That's a regression in what this RPC promises. The RPC paths that attach UTXOs (`walletupdatepstt`, the new `updatepstt`) both reject out-of-range indices, so this too comes in via an externally-produced PSTT.

Skipping the estimate (as with `identification_txid`) rather than throwing would keep the existing contract.

**Test:** `decodepstt` on that shape should return a result without `estimated_size`/`estimated_fee`, not error.

## 4. `joinpstt` silently drops `fallback_locktime` (and `version`)

```cpp
if (!joined.fallback_locktime) joined.fallback_locktime = pstt.fallback_locktime;
```

If PSTT A carries 500 and B carries 900, B's value disappears with no indication. The `unknown` map's pick-first is at least labelled as matching `combinepstt`'s tier; this isn't mentioned in the help at all. Please either take the maximum or refuse on a mismatch, and say which in the help text. `version` isn't carried over from any input PSTT either — intentional or not, it's worth stating.

**Test:** join two PSTTs with differing `fallback_locktime` and assert whichever rule you pick.

## 5. `next` can point at the wrong role

```cpp
if (pstt.tx_modifiable && (*pstt.tx_modifiable & (INPUTS|OUTPUTS)) != 0) return "constructor";
```

Any open modifiable bit short-circuits to `constructor`. A PSTT from `walletcreatefundedpstt(..., inputs_modifiable=true)` has its UTXOs attached and is unsigned, so the user's actual next step is signing — but `next` says `constructor`.

"Construction is still permitted" and "construction is what should happen next" aren't the same thing, and the help promises the latter ("which role should run next"); Bitcoin's `analyzepsbt` reports the required next role. Either report the required role, or state explicitly in the help that an open modifiable bit takes precedence.

## 6. The estimate's fidelity depends on what key material happens to be attached

`EstimateInputScriptSig` builds a keystore from `hd_keypaths` and `partial_sigs` pubkeys and dummy-signs with `DUMMY_SIGNATURE_CREATOR`. With BIP32 data attached that approximates the signed size well; for a PSTT built by `createpstt` + `updatepstt` with no wallet involved, nothing goes into the keystore and the number is essentially the unsigned size.

"the transaction this PSTT would currently extract to" isn't wrong, but it reads as a cost estimate, which is what the description asks for ("what would this cost if extracted right now?"). Worth saying which of the two it is.

Relatedly, `estimated_fee` is `in_tpc - out_tpc`, so a PSTT still gathering inputs reports a **negative** fee. `analyzepsbt` omits the field when it isn't meaningful; the same would avoid a confusing number here.

## 7. Unrelated rename inflates the diff

`prev_out_index` → `prevout_index` and `PREV_OUT_INDEX_OOB` → `PREVOUT_INDEX_OOB` span six files in a PR that's otherwise about three new RPCs. Landing right after #451 merged, that's avoidable rebase surface — better as its own commit or PR.

One part of it is a readability regression: `utxo_out` → `utxo`. In `SignPSTTInput`, `ComputePsttColorBalances` and `EstimateInputScriptSig` you now have `input.utxo` (a `CTransactionRef`, the whole previous transaction) and `utxo` (a `CTxOut`, one of its outputs) live in the same scope. The old name kept those apart.

## Test coverage

The 15 new unit tests cover `joinpstt`'s rejection paths well — modifiable bits, Has-SIGHASH_SINGLE, already-signed inputs, duplicate outpoints. What's missing maps onto the findings above: duplicate xpubs across joined PSTTs (1), an output with no amount (2), an out-of-range `prevout_index` on a UTXO-attached input (3), and conflicting `fallback_locktime` (4). Each is a small addition and each guards a case that currently either corrupts the output or throws.

azuchi (changes requested):
Marking this as changes-requested to match the findings in https://github.com/chaintope/tapyrus-core/pull/454#issuecomment-5489444561 — the detail is all there rather than repeated here.

The three that make this blocking:

1. `joinpstt` concatenates `xpubs` without dedup (`src/rpc/rawtransaction.cpp:2262`), while `Unserialize` rejects duplicate `PSTT_GLOBAL_XPUB` keydata — so the RPC can emit a PSTT that this same implementation can no longer decode.
2. `EstimatePsttFee` dereferences `output.amount` unconditionally; it's a `boost::optional` and `operator*` on a disengaged one is undefined behaviour.
3. `EstimatePsttFee`/`EstimateInputScriptSig` use `.at(input.prevout_index)` guarded only by a non-null UTXO check, so `decodepstt` now throws on malformed-but-parseable PSTTs it previously displayed — against its own stated contract.

Items 4–7 in that comment (dropped `fallback_locktime`/`version`, the `next` short-circuit, the estimate's dependence on attached key material plus negative fees, and the unrelated `prev_out_index`/`utxo_out` rename) are worth addressing too but wouldn't block on their own.

The four missing tests listed there map one-to-one onto findings 1–4.

azuchi (changes requested):
Reviewed at `d0e520cc`. All seven items are addressed — thanks, and in particular for the reachability analysis on findings 2 and 3.

You're right and I was wrong there: `PartiallySignedTapyrusTransaction::Unserialize()` enforces `IsSane()` unconditionally at the end of parsing (`src/pstt.cpp:830`, and it predates this PR), so neither the missing-`PSTT_OUT_AMOUNT` output nor the out-of-range `prevout_index` can reach `decodepstt`'s body — `DecodePSTT()` rejects both first. My "regression in what this RPC promises" framing was incorrect. The `value_or(0)` and range-check guards are still worth keeping as defense in depth, and pinning that down with tests at both the C++ and the RPC boundary, with the unreachability spelled out in the comments, is the right way to close it.

Two things left.

## 1. CI is red: the new dedup test fails

`pstt_tests` fails on both Linux x86_64 and Linux ARM64:

```
test/pstt_tests.cpp(1914): fatal error: in "pstt_tests/pstt_rpc_joinpstt_dedups_shared_xpub":
critical check joined.xpubs.size() == 1U has failed [2 != 1]
```

I believe the dedup implementation is correct and the test is what's wrong — it leaves `CExtPubKey::vchFingerprint` uninitialized:

- The preceding `BOOST_REQUIRE(DecodePSTT(...))` passes. `Unserialize()` rejects a repeated `PSTT_GLOBAL_XPUB` keydata outright (`src/pstt.cpp:764-766`), so the fact that the joined PSTT decodes at all proves its two entries have *different* keydata.
- keydata is the 4-byte prefix plus `CExtPubKey::Encode()`: `nDepth`, `vchFingerprint[4]`, `nChild`, `chaincode`, `pubkey` (`src/pubkey.cpp:265-273`).
- The test sets `pubkey`, `nDepth`, `nChild` and `chaincode`. `CExtPubKey` has no default member initializers (`src/pubkey.h:220-225`), so `vchFingerprint` is the one field left indeterminate, and it goes straight into the keydata via `Encode()`.

So the two entries aren't actually the same xpub, and `joinpstt` correctly declines to merge them.

Fix by initializing it — `memset(xpub.vchFingerprint, 0, sizeof(xpub.vchFingerprint));` — or by deriving the xpub properly (`CExtKey::SetSeed()` + `Neuter()`), which also gives a realistic fingerprint.

Worth noting that until this is fixed the dedup path is effectively untested, so please confirm the test goes green rather than just stops failing.

(The macOS ARM64 failure in the same run is #458 — Homebrew dropped `berkeley-db@4` — and is unrelated to this PR.)

## 2. `doc/tapyrus/pstt.md` still documents the old single-role `next`

The RPC help text was updated for item 5, but the doc wasn't:

- `doc/tapyrus/pstt.md:230-231` still describes "which role above should run next for this PSTT (`"next"`: `constructor`/`updater`/`signer`/`finalizer`/`extractor`)" — singular.
- `:241`: "Both estimates can still change while `"next"` is `constructor` or `signer`" — reads as a string comparison, not array membership.

While you're there, `:235-237` reads as unconditional ("Once every input carries `PSTT_IN_UTXO`, it also reports `"estimated_size"` ... and `"estimated_fee"`"), which no longer holds now that a negative fee is omitted. The help text covers this; the doc should say the same.

azuchi (approved):

PR #455, opened 2026-09-01
Upgrade upload artefact GitHub action

Naviabheeman:
Upgrade upload artefact actions in the CI from v5 to v6 to fix the CI deprecated warnings.

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -957,7 +957,7 @@ jobs:
 
       - name: Upload test results
         if: always()
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: daily-test-results-${{ matrix.config.name }}-${{ github.run_id }}
           path: |
@@ -969,7 +969,7 @@ jobs:
 
       - name: Upload coverage data
         if: matrix.config.build_type == 'Debug'
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: coverage-data-${{ matrix.config.name }}-${{ github.run_id }}
           path: |
@@ -979,7 +979,7 @@ jobs:
 
       - name: Upload Windows binaries
         if: matrix.config.target_platform == 'windows' && always()
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: tapyrus-windows-${{ matrix.config.target_arch }}-binaries-${{ github.run_id }}
           path: |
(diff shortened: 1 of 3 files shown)


azuchi:
The bump itself is clean and I have nothing to flag in the diff. One correction to the description.

## This doesn't fix the macOS arm64 CI failure

The most recent macOS ARM64 failure (run 33385036219, job 99465765013) failed in the functional test step, not in an upload:

```
p2p_addr_relay.py     | ✖ Failed  | 13 s
ALL                   | ✖ Failed
Fast functional tests failed
##[error]Process completed with exit code 1.
...
##[group]Run actions/upload-artifact@v5      <- runs after, completes fine
```

The assertion is `p2p_addr_relay.py:272`:

```python
assert_greater_than_or_equal(2, len(nodes_received_addr))   # AssertionError: 2 < 3
```

— a probabilistic addr destination-rotation check, unrelated to artifact upload. The `upload-artifact` steps are `if: always()`, so they ran *after* that failure and produced no error of their own. The only upload-related output in that job is the warning:

```
##[warning]Node.js 20 is deprecated. The following actions target Node.js 20
but are being forced to run on Node.js 24: actions/upload-artifact@v5
```

So the second half of the description holds — this clears the deprecation warning — but the first half doesn't. Merging it as written would leave a real flaky test believed fixed. Could you drop that clause from the description and open a separate issue for the `p2p_addr_relay.py` flake? It has failed at least twice recently (this run, and master's smoke-test on 2026-08-27).

## What I checked on the change itself

- All 10 occurrences bumped; no `upload-artifact@v5` left anywhere under `.github/workflows/`.
- No `actions/download-artifact` in the repo, so there's no counterpart action whose major version has to move in step.
- v6 is exercised for real, not just on empty paths: this PR's smoke run (33511848360) uploaded `tapyrus-windows-x86_64-binaries` at 89.6 MB. The `No files were found ... No artifacts will be uploaded` lines on some jobs are the pre-existing `if-no-files-found` behaviour, not a v6 change.
- The Node 20 warning is gone entirely — zero occurrences in this PR's macOS log. `actions/checkout@v5` and `actions/cache@v5` already target Node 24, so nothing else needs bumping to finish the job.

Naviabheeman:
The was about the job https://github.com/chaintope/tapyrus-core/actions/runs/33447295888/job/99669141073 which is a daily test failure in the master branch. It does not have any functional test failure, just upload artefact failure.

Created issue [10](https://github.com/chaintope/tapyrus-integration-tests/issues/10) to remember this flakiness. it doesn't seem to have a direct fix.

PR #456, opened 2026-09-01
AI-assisted fuzz-testing pipelines with initial set of tests setup in the daily test

Naviabheeman:
Initial fuzz test implementation design:
Two AI-assisted pipelines that grow this repo's fuzz coverage, plus the local tooling and daily CI wiring around them. Nothing here has been run against real external credentials yet.

- **fuzz_code** (`contrib/fuzz/oss-fuzz/` + `contrib/fuzz/fuzz-introspector/`): Fuzz Introspector finds functions with no fuzz coverage → OSS-Fuzz-Gen drafts a `fuzz_test_file` via Claude (GCP Vertex AI) → a generated `review_code.html` page gets human approval → approved drafts land as permanent libFuzzer targets, run daily and free by `daily-test.yml`'s `fuzz-code-only` job with zero further AI involvement.
- **fuzz_script** (`contrib/fuzz/fuzz4all/`): Fuzz4All generates candidate Tapyrus Script programs via Claude (Anthropic API, alternating Haiku 4.5/Sonnet 5), verified against a new `tapyrus-verify` tool → a generated `review_scripts.html` page prunes the batch down to what's worth keeping → the daily `fuzz-script-sweep` job replays a rotating window of the committed pool for free, no AI involved.
- Both pipelines draw from **one shared $50/month budget** (`contrib/fuzz/fuzz_spend_ledger.py`), not two independent ones.
- New `tapyrus-verify` tool (`src/tapyrus-verify.cpp`), gated behind new `BUILD_SCRIPT_VERIFY` (default ON) and `BUILD_FUZZ_TEST` (default OFF) CMake options — deliberately lean, links only `tapyrus_consensus`/`tapyrus_common`, never touches storage or networking.
- All local orchestration scripts are Python + asyncio (not bash) — real concurrency where it matters (bounded-concurrency harness drafting, concurrent build+clone), not concurrency for its own sake.

Full architecture, both sequence diagrams, and a script-by-script reference table: [`contrib/fuzz/PIPELINE.md`](contrib/fuzz/PIPELINE.md) (same content also published as an [interactive diagram](https://claude.ai/code/artifact/511a3c9f-c31c-4a6f-a3ab-0d4c7e7580d1)

## Known gaps
- **Fuzz Introspector's instrumented build** has not been run end-to-end (needs a real multi-stage build via `google/oss-fuzz`'s own tooling).
- **OSS-Fuzz-Gen drafting** has not been run for real (needs GCP Vertex AI credentials, not provisioned in this session).
- **Fuzz4All generation** has not been run for real 
- Everything *around* those three integrations — budget tracking, review/land HTML workflows, path resolution after the directory reorg, CMake wiring, CI job structure — has been verified with real (non-mocked) unit and integration tests locally. See each subdirectory's README for what's confirmed vs. still to prove out on a real run.

## tapyrus-verify — standalone Script/transaction verifier

What it is: a small CLI tool (src/tapyrus-verify.cpp) that takes two raw transactions — a to_spend transaction and a spending transaction — and runs the real VerifyScript on them, exactly the way BIP-341/342's own test vectors are shaped: one output being spent, one input spending it. It's the oracle behind both fuzz pipelines' script-level checking, and it's also just directly useful on its own for reproducing any script-verification result from a hex pair.

Two modes

1. Plain mode: tapyrus-verify <to_spend_hex> <spending_hex> [input_index]. Deserializes both transactions, runs VerifyScript on the given input (default 0), prints PASS/FAIL plus the script error string to stderr. This is the mode a human uses to reproduce something interesting found elsewhere.
2. --fuzz mode (only compiled in when BUILD_FUZZ_TEST is on): tapyrus-verify --fuzz <script-mnemonic-file>. Takes a file of Script opcode mnemonics (what Fuzz4All's Claude-driven generation actually produces), assembles it via ParseScript, builds a to_spend/spending pair around it, and sweeps it against 16 nLockTime × nSequence combinations (4 locktimes: disabled, smallest height-based, the historical height-vs-time boundary, max uint32; 4 sequences: final, both-locks-enabled, relative-lock-disabled, one high-but-not-final value) rather than just one — this is deliberately not exhaustive, just enough variation to catch locktime/sequence-interaction bugs a single fixed pair would miss.


azuchi:
Reviewed at `46f8f32`. Taking it as the design draft it says it is, so design questions first, then the concrete defects.

Two things I checked first because they'd have been the serious ones, and both are right: **no secrets are referenced anywhere in the workflow** — AI generation is local-only and the daily jobs replay already-committed material, matching the "CREDENTIALS NEEDED: none" claim — and **nothing auto-commits**. Both land scripts end at `"Not committed -- review the working-tree diff and git add/commit yourself."` (`fuzz_script_land_approved.py:80`, `fuzz_code_land_approved.py:153`). Generated content reaching the repo only through human review is the right call and it's actually implemented that way.

## Nothing in this PR has been compiled by CI

- Smoke run 33520603716: Lint Check failed, and `needs: lint` (`smoke-test.yml:45`) skipped the entire build matrix.
- Daily run 33520603601: every job `cancelled`.

So `src/tapyrus-verify.cpp`, `src/test/fuzz/pstt_parse_fuzz.cpp` and the CMake wiring have not been built anywhere. Worth clearing before the design discussion goes much further — a lot of the review surface here is "does this compile and link as described".

Four lint failures:

| lint | file |
| --- | --- |
| `lint-filenames` | `contrib/fuzz/fuzz4all/TAPYRUSSCRIPT.py`, `src/test/fuzz/FuzzedDataProvider.h` |
| `lint-include-guards` | `src/test/fuzz/FuzzedDataProvider.h` |
| `lint-locale-dependence` | `std::strtoul` in `src/tapyrus-verify.cpp` |
| `lint-shell-locale` | `contrib/fuzz/oss-fuzz/project/build.sh` missing `export LC_ALL=C` |

The first two are about the vendored LLVM header, so the fix is a lint exclusion rather than editing the file — Bitcoin Core carries the same file and handles it that way.

## `VerifyPair` verifies against the wrong output

`src/tapyrus-verify.cpp:105-118`:

```cpp
VerifyScript(
    spending.vin[nIn].scriptSig,
    toSpend.vout[nIn].scriptPubKey,
    ...
    MutableTransactionSignatureChecker(&spending, nIn, toSpend.vout[nIn].nValue),
```

The output being spent is `spending.vin[nIn].prevout.n`, not `nIn`. Those coincide only when input *i* happens to spend output *i*. `--fuzz` mode is a fixed 1-in/1-out shape so it's unaffected, but plain mode takes `input_index` as an argument, and `tapyrus-verify <a> <b> 1` will check input 1 against `to_spend.vout[1]` — the wrong scriptPubKey and the wrong `nValue`. A valid spend reports FAIL, or worse passes against an unrelated output.

`RunVerify`'s bounds check (`nIn >= toSpend.vout.size()`) encodes the same assumption, so it won't catch it. There's also no check that `spending.vin[nIn].prevout.hashMalFix` actually refers to `to_spend`, so a mismatched pair silently produces a confident wrong answer.

This is exactly the use the description sells the tool on ("directly useful on its own for reproducing any script-verification result from a hex pair").

## `-DBUILD_BENCH=OFF` isn't an option this project defines

Both new jobs pass it; the real name is `ENABLE_BENCH` (`CMakeLists.txt:188`). CMake will emit "Manually-specified variables were not used by the project" and `ENABLE_BENCH` stays at its `ON` default, so `bench_tapyrus` gets built in both fuzz jobs.

For contrast, `-DSANITIZERS=` *is* real (`CMakeLists.txt:333`), so the instrumentation itself does take effect.

## Design: `BUILD_SCRIPT_VERIFY` defaults ON and installs

```cmake
option(BUILD_SCRIPT_VERIFY "Build tapyrus-verify executable." ON)
...
install_binary_component(tapyrus-verify)
```

That puts a draft fuzzing-support tool into every build and every release package, adding a user-facing binary and its support surface. It also isn't marked `HAS_MANPAGE`, unlike the neighbouring tools. Defaulting to OFF until it has settled — and flipping it on deliberately once it's a tool in its own right — seems the safer sequencing.

## Design: `STANDARD_SCRIPT_VERIFY_FLAGS`

Fine as a crash oracle. Less fine for the "reproduce any script-verification result" framing, since standard flags aren't consensus flags and a result here can differ from what the network would do. Either say so in the usage text/README, or make the flag set selectable.

## `FuzzedDataProvider.h` licensing

It's Apache-2.0 WITH LLVM-exception vendored into an MIT repo. Compatible, but I don't see the third-party attribution recorded anywhere. Worth sorting alongside the lint exclusions.

## Smaller points

- The new jobs use `actions/upload-artifact@v5` in six places while #455 moves the repo to v6 — whichever lands second will be inconsistent.
- `fuzz-code-only` has no `timeout-minutes`, while `fuzz-script-sweep` sets 50. With `RUN_SECONDS=600` per target that grows with every target added to `FUZZ_TARGETS.txt`.

## On the rest of the design

The exit-code contract (0 clean / 1 usage / 2 assembler-rejected / signal = real crash) is a good separation of LLM noise from genuine findings, and threading it through the sweep job so ordinary noise never reds the build is the right shape. Keeping generation local and CI replay-only means the daily jobs stay free and credential-free, which is what makes this sustainable. `BUILD_FUZZ_TEST` defaulting OFF with the existing matrix passing `-DBUILD_FUZZ_TEST=OFF` explicitly is belt-and-braces in a good way.

Naviabheeman:
#457 now contains the Tapyrus-verify tool with feedback addressed and with tests. it is split so that the CI can run it independent of this PR's unverified design.

azuchi:
Re-reviewed at `dcaf4d3`. Splitting `tapyrus-verify` out into #457 leaves this PR as the pipelines plus the `--fuzz` mode on top, which reads much better. Everything I raised last time is addressed:

- Lint is fixed by excluding the vendored files in `lint-filenames.sh`/`lint-include-guards.sh` rather than editing them — the right way round.
- `-DBUILD_BENCH=OFF` is now `-DENABLE_BENCH=OFF` in both jobs.
- `fuzz-code-only` has `timeout-minutes: 30`, with a note that it needs raising as the target list grows.
- `FuzzedDataProvider.h`'s licence is recorded in a new "Vendored source files" section in `doc/dependencies.md` — and `src/tinyformat.h` got documented alongside it, which I hadn't asked for.
- The output-index bug, the `BUILD_SCRIPT_VERIFY` default and the `STANDARD_SCRIPT_VERIFY_FLAGS` question all moved to #457 and are fixed there.

The `--fuzz` mode itself is consistent with #457's fixed pair handling: it builds its own 1-in/1-out pair and passes `VerifyContext{toSpend, spending, 0, 0}`, where `prevout.n` really is 0, and reuses `RunOne` so the flag matrix applies there too.

## `upload-artifact@v5` × 12 — #455 has landed

master is now at `3ef42b96` (#455), which moved every `upload-artifact` to v6. This branch's merge-base is `75082c425` (#452), so it still has v5 in all twelve places — the six it inherits and the six it adds.

Rebasing will conflict on the six inherited ones. Merging as-is would put six fresh v5 steps back in and reinstate the Node 20 deprecation warnings #455 just removed. Needs a rebase plus bumping the new steps to v6.

## Nothing fuzz-related is compiled on a pull request

`BUILD_FUZZ_TEST=ON` is set only by `fuzz-code-only` (`:1137`) and `fuzz-script-sweep` (`:1238`), and both carry `if: github.event_name != 'pull_request'` (`:1100`, `:1213`). The main matrix passes `-DBUILD_FUZZ_TEST=OFF` explicitly (`:621`).

So on a PR neither `src/test/fuzz/pstt_parse_fuzz.cpp` nor the `--fuzz` mode added in this PR is ever compiled.

Skipping the *runs* on PRs is well argued in the job comments — coverage-guided fuzzing is exploration over time, not a per-change gate. But the *build* is a different thing. `pstt_parse_fuzz.cpp` calls `ParsePsttInputEntries` through `rpc/rawtransaction.h`, so it breaks whenever that signature moves, and right now that break would only surface on the next scheduled run, well after the change that caused it.

Splitting the two would fix it cheaply: build the fuzz targets on PRs, keep the actual fuzzing on the schedule. (For what it's worth the signature does still match today — `(const UniValue&, uint32_t, bool)` — so nothing is broken, it's just unguarded.)

## `pstt_parse_fuzz.cpp` cites an RPC that doesn't exist

`src/test/fuzz/pstt_parse_fuzz.cpp:6-7`:

> Fuzz target for ParsePsttInputEntries ..., the entry point that turns the "inputs" argument of `createpsttfromtransaction` et al.

There's no `createpsttfromtransaction`. The two real callers are `createpstt` (`rpc/rawtransaction.cpp:1476`) and `walletcreatefundedpstt` (`wallet/rpcwallet.cpp:4265`).

azuchi:
Reviewed at `ebb7e38b`. All three findings from the last round are addressed: the branch is rebased onto master so every `upload-artifact` step is v6 (no v5 left anywhere under `.github/workflows/`), `daily-test.yml:640` now compiles `tapyrus-verify --fuzz` on every PR via `-DBUILD_SCRIPT_VERIFY=ON -DBUILD_FUZZ_TEST=ON` on native builds, and `src/test/fuzz/pstt_parse_fuzz.cpp:7` cites `createpstt`. The lint exclusions and the "Vendored source files" section in `doc/dependencies.md:61-68` survived the rebase.

One blocker, introduced by that new compile-check itself.

## `tapyrus-verify` fails to link when the wallet is off

`L-X86-Rel-No_GUI+No_Wallet` and `L-ARM-Rel-No_GUI+No_Wallet` both fail:

```
FAILED: [code=1] bin/tapyrus-verify
lib/libtapyrus_util.a(random.cpp.o): in function `FastRandomContext::FastRandomContext(uint256 const&)':
random.cpp:623: undefined reference to `ChaCha20::ChaCha20()'
```

It's static library ordering. The link line resolves to

```
... lib/libtapyrus_crypto.a lib/libtapyrus_util.a src/univalue/libunivalue.a lib/libtapyrus_clientversion.a
```

with crypto *before* util. `random.cpp` uses ChaCha20 (`src/random.h:9` includes `crypto/chacha20.h`), but `tapyrus_util` never declares a dependency on `tapyrus_crypto` (`src/CMakeLists.txt:112-120`), so nothing puts crypto after it.

Two things kept this hidden until now:

- `-DBUILD_FUZZ_TEST=ON` (new, `daily-test.yml:640`) compiles the `--fuzz` path in for the first time, which pulls `random.cpp.o` into the link. #457 is green with an identical `target_link_libraries(tapyrus-verify ...)` purely because it never sets `BUILD_FUZZ_TEST`.
- With the wallet enabled, `tapyrus_util` links `$<TARGET_NAME_IF_EXISTS:tapyrus_wallet>` PRIVATE (`src/CMakeLists.txt:119`), which drags `tapyrus_crypto` in *after* util via `tapyrus_server` and resolves it by accident. `-DENABLE_WALLET=OFF` empties that generator expression and the accident stops happening.

This also affects a job this PR adds: `fuzz-script-sweep` (`daily-test.yml:1222-1250`) configures with exactly `-DENABLE_WALLET=OFF -DBUILD_SCRIPT_VERIFY=ON -DBUILD_FUZZ_TEST=ON` and builds `--target tapyrus-verify`. It only skips here because it's schedule-only, so it would hit the same link error on its first scheduled run.

The real fix is adding `tapyrus_crypto` to `tapyrus_util`'s PRIVATE link libraries — that's the dependency edge that's genuinely missing, and it fixes every consumer. Listing `tapyrus_util` ahead of `tapyrus_crypto` in `tapyrus-verify`'s own `target_link_libraries` would also work, but only patches this one binary.

## Nit

`src/test/fuzz/FUZZ_TARGETS.txt`'s header says it's read by `contrib/fuzz/oss-fuzz/build.sh`; the actual reader is `contrib/fuzz/oss-fuzz/project/build.sh:15`.

azuchi:
Reviewed at `26671f23`. The update is a rebase onto master rather than new work, and it leaves the PR in a much cleaner shape: with #457 and #459 merged, the diff no longer carries `src/tapyrus-verify.cpp` wholesale, `test/CMakeLists.txt`, the `tapyrus-verify-test.{py,json}` pair or `cmake/tests.cmake`. What's left is the AI fuzz pipelines plus the `--fuzz` mode, and the only `src/` changes are 126 lines in `tapyrus-verify.cpp` and three in `src/CMakeLists.txt`.

The ChaCha20 link failure I reported is gone — #457 merged with `tapyrus_crypto` declared on `tapyrus_util`, so this branch inherits the fix. CI is 14 green. The rebase also merged the two workflow edits into one coherent line rather than duplicating them:

```bash
# Build and test tapyrus-verify on every native job; also
# compile-check its --fuzz mode (see comment above) on every PR.
CMAKE_OPTS+=("-DBUILD_SCRIPT_VERIFY=ON" "-DBUILD_FUZZ_TEST=ON")
```

Two things from last round are still open.

## 1. `fuzz_pstt_parse` still isn't compiled on PRs

`daily-test.yml:622-623` still says it's a follow-up, and turning `BUILD_FUZZ_TEST=ON` for native jobs doesn't change that, because the target is excluded from the default build:

```cmake
add_executable(fuzz_pstt_parse EXCLUDE_FROM_ALL
    fuzz/pstt_parse_fuzz.cpp
)
```

`cmake --build build` skips it; only `fuzz-code-only`'s explicit `--target ${FUZZ_TARGETS}` builds it, and that job is schedule-only. So `pstt_parse_fuzz.cpp` is configured but never compiled on a pull request.

That's the same exposure this PR itself argues against for `tapyrus-verify --fuzz` — a signature drift goes unnoticed until the next scheduled fuzz run — and it's more likely here, since `DecodePSTT` and the surrounding RPC surface were being actively changed in #454 right up to its merge.

Either add `cmake --build build --target fuzz_pstt_parse` to the native jobs, or drop `EXCLUDE_FROM_ALL` when `BUILD_FUZZ_TEST` is on, which also makes the fuzz jobs build it without naming it explicitly.

## 2. `FUZZ_TARGETS.txt` still cites the wrong path

`src/test/fuzz/FUZZ_TARGETS.txt:8` names `contrib/fuzz/oss-fuzz/build.sh` as a reader; the file that actually reads it is `contrib/fuzz/oss-fuzz/project/build.sh:15`.

## Unrelated, worth its own issue

The new comment in `src/test/CMakeLists.txt` notes that `test_tapyrus_fuzzy.cpp` sits in `test_tapyrus`'s source list and is therefore never exercised as a fuzz binary, flagged as pre-existing and deliberately not fixed here. That seems right to leave out of this PR, but it's the kind of thing that quietly stays broken forever unless it's tracked somewhere.

azuchi:
@Naviabheeman fix lint error.

azuchi:
Reviewed at `ea4e3a59`. The lint fix works — `Lint Check` is green again, and with it the smoke-test matrix actually runs (it had been skipped wholesale via `smoke-test.yml:45`'s `needs: lint`, which is what the odd `matrix.config.name / skipping` entry was). Everything I raised on this PR is now addressed.

Two things left before merging, neither of them a code change.

## 1. Rebase onto master to clear the macOS failure

`macOS x86_64` fails on `rpc_pstt.py | ✧ Timeout | 246 s`, and it's the race #461 fixed, not anything in this PR:

```
node2 06:39:03.685480Z UpdateTip: new best=d1513451... height=2
node0 06:39:03.713964Z UpdateTip: new best=dd95b193... height=2
node1 06:39:03.724729Z UpdateTip: new best=dd95b193... height=2
```

Two different height-2 blocks 28 ms apart, then `sync_blocks` polls `getbestblockhash` until the test times out.

This branch's merge-base is `e0c1459f0` (#459), so it's missing both:

```
a759173cb Fix ARM64 runner mislabeling, add minimum-supported-build testing, version-scoped depends patches (#453)
13e30f508 Fix rpc_pstt.py: node2 mines without syncing, causing a chain fork
```

`test/functional/rpc_pstt.py:161` on this branch still has no `sync_all()` after `nodes[2].generate()`. A rebase (or merge) picks up the fix and the job goes green.

I checked the rebase is clean, including the part I'd have expected trouble from: #453 also rewrites large parts of `daily-test.yml`, but the merged file correctly ends up with both its changes (the `Reset depends/ tracked files` step, the `depends/built/$HOST/$pkg` filter on the expected-versions list) and this PR's new fuzz jobs. #453's new "flat patches under `depends/patches/<pkg>/` are a build error" rule doesn't affect this PR either, since it touches no `depends/` files.

## 2. Please run the two scheduled jobs once by hand before merging

`fuzz-code-only` and `fuzz-script-sweep` are gated on `if: github.event_name != 'pull_request'`, so they have shown `skipping` on every run of this PR, and I don't see a `workflow_dispatch` run of them on this branch either. Meanwhile the history has:

```
0f9a32d86 Fix fuzz-code-only/fuzz-script-sweep: broken sanitizer flag + link order
```

So they were broken at one point and fixed, but the fixed state has never actually executed. As things stand the first real run is the first nightly after merge, and if anything is still off, master goes red every night until someone notices.

One `workflow_dispatch` run of `Cmake Daily Test` on this branch would settle it. The jobs need no credentials (the header comment says `CREDENTIALS NEEDED: none.`, and there's no `secrets.` reference anywhere in the workflow), so it should be cheap to do.

Worth being explicit about the ongoing cost while you're there: `fuzz-code-only` is `timeout-minutes: 90` (7 targets × `RUN_SECONDS` 600 plus margin) and `fuzz-script-sweep` budgets `SWEEP_TIME_BUDGET_SECONDS` 1800, so this adds up to roughly two hours of runner time per night. They run alongside the existing matrix rather than extending it, but it's a real increase in consumption that the team should be signing up for knowingly.

## Nit

The title and description still frame this as `draft ... for design review`, but what's here is a working CI integration. Worth updating both so the merge commit reads as what it is.

azuchi:
The rebase did its job — `675b95b2` sits on current master (`a759173cb`), the macOS `rpc_pstt.py` timeout is gone with #461, and all 14 checks pass. I also confirmed the rebase kept both sides of `daily-test.yml` intact: #453's `Reset depends/ tracked files` step and its `depends/built/$HOST/$pkg` filter on the expected-versions list are still there alongside this PR's fuzz jobs.

That leaves the other half of my last comment, which hasn't happened yet.

## Please still run the two scheduled jobs once by hand

`fuzz-code-only` and `fuzz-script-sweep` have shown `skipping` on all five runs of this branch, and there is no `workflow_dispatch` run of `Cmake Daily Test` in the workflow's history at all — not on this branch, not anywhere. So they remain the only part of this PR that has never executed, while the history says they were broken at one point:

```
d8a39f0c2 Fix fuzz-code-only/fuzz-script-sweep: broken sanitizer flag + link order
```

If anything is still off, the first run is the nightly after merge (~20:45 UTC), and master goes red every night until somebody notices. One `workflow_dispatch` run from this branch settles it, and the jobs need no credentials (`CREDENTIALS NEEDED: none.`, no `secrets.` reference in the workflow).

Everything else I raised on this PR is resolved, so this is the last thing I'd want before merge.

## Description

The title now matches what this is, but the body still opens with:

> Draft, for design review: two AI-assisted pipelines ... Nothing here has been run against real external credentials yet.

Also, the bullet describing `New tapyrus-verify tool (src/tapyrus-verify.cpp), gated behind new BUILD_SCRIPT_VERIFY` predates #457 — the tool itself is already on master, and what this PR adds is its `--fuzz` mode. Worth a pass over the body so the merge commit reads accurately.

azuchi:
Thanks for running these from your fork — and that run earned its keep. Confirmed at `d3e30dce`:

```
35080192986  workflow_dispatch  tapyrus-fuzz-test  success
  success  fuzz-script-sweep   2m15s
  success  fuzz-code-only      1h16m
```

`d3e30dce` is the right fix, and the commit message names the cause exactly. For anyone reading later: the step runs under `shell: /usr/bin/bash -e {0}`, and `rc=$?` on the following line isn't a "tested context", so `set -e` killed the shell the moment `tapyrus-verify` returned 2 — before the `rc` check could decide to tolerate it. The `|| rc=$?` form puts the command in a tested context, so `-e` stays out of it. Worth noting a re-run could never have fixed this: exit 2 is the documented "ParseScript rejected the candidate, ordinary LLM noise" path, so it recurs for as long as any unparseable candidate sits in the pool — which the design says should happen occasionally.

This is exactly why I wanted the manual run before merge: the `-e` interaction meant the exit-2 tolerance had never worked, and the first nightly after merge would have gone red and stayed red.

## Two numbers worth recording

**`fuzz-script-sweep` is far cheaper than its budget suggests.** It finished in 2m15s, not the 30 minutes `SWEEP_TIME_BUDGET_SECONDS` allows, because it laps the pool first:

```
Completed a full lap of the pool (30 candidates) before the time budget ran out.
Tested 30 of 30 pool candidates this run (started at offset 22).
```

So my earlier "roughly two hours a night" estimate was too pessimistic — at 30 candidates the sweep costs about two minutes, and only approaches its budget once the pool grows by an order of magnitude.

**`fuzz-code-only` has less headroom than its comment claims.** It took 76 minutes against `timeout-minutes: 90`. The comment says:

> 7 targets * 600s default = 70 min alone; 90 min leaves ~20 min margin.
> bump this when target count times RUN_SECONDS/60 approaches it.

Measured margin is 14 minutes, and one more harness at the default `RUN_SECONDS` of 600 adds ~10, landing at ~86 minutes — inside the timeout by four. Since the whole point of the `fuzz_code/*_fuzz.cpp` glob is that landing a new harness needs no CI edits, this will bite silently. Worth raising `timeout-minutes` or lowering `RUN_SECONDS` now, while it's a deliberate choice rather than a surprise.

## Description

Still the last loose end: the body opens with "Draft, for design review", and its `tapyrus-verify` bullet describes the tool as new here, which predates #457 — what this PR adds is the `--fuzz` mode. Everything else I raised is resolved.

azuchi (approved):

PR #457, opened 2026-09-02
Tapyrus verify tool

Naviabheeman:
This PR splits Tapyrus-verify tool from #456

What tapyrus-verify does

It's a standalone CLI tool that answers one question: does this spending transaction correctly spend one output of this to_spend transaction? — the same to_spend/spending pair shape Bitcoin's BIP-341/342 test vectors use.

Two modes:
- Plain mode: tapyrus-verify <to_spend_hex> <spending_hex> [input_index] [--flags=<spec>] — deserializes both, resolves which output spending's input actually claims to spend (from its own prevout, not from the index you pass), verifies it, prints PASS/FAIL.

Design points that came out of review this session:
- Deliberately lean: links only tapyrus_consensus + tapyrus_common, never tapyrus_server/leveldb — no storage, no networking, no block-level validation.
- Validates the pair, not just the script: confirms spending's referenced output actually matches to_spend (hash + index) before running anything, and does basic structural checks (money range, no duplicate inputs, valid script opcodes) — catching a bug where the wrong output could silently get checked.
- Explicit about its own limits: its header now documents that a PASS is necessary-but-not-sufficient for one input's script correctness — not a node-acceptance guarantee, since it checks one input in isolation, has no live chain state, and never runs mempool/relay policy checks.
- BUILD_SCRIPT_VERIFY defaults OFF — an opt-in, still-settling tool, not shipped in every build/release by default.
- Has its own util test suite (9 cases) and uses only locale-independent parsing


azuchi:
Reviewed at `64086ec`. Everything I raised on #456 is addressed, and the fixes go further than I asked in a couple of places. Two things left.

## Please enable `BUILD_SCRIPT_VERIFY` in at least one CI job

`BUILD_SCRIPT_VERIFY` now defaults OFF, which is right — but no workflow overrides it:

```
$ grep -rn "BUILD_SCRIPT_VERIFY" .github/workflows/
(nothing)
```

So `src/tapyrus-verify.cpp` (467 lines) is never compiled anywhere, and the nine-case suite added here sits under `if(BUILD_SCRIPT_VERIFY)` (`test/CMakeLists.txt:98`), meaning neither `check-tapyrus-verify` nor the `tapyrus_verify_test` ctest has ever run. The 20/20 green on this PR says nothing about any of it.

One job configuring with `-DBUILD_SCRIPT_VERIFY=ON` covers both the build and the tests in one go. This is the same gap #456 had, just reached a different way — there it was Lint blocking the matrix, here it's the default with nothing turning it on.

## Test coverage: no unit tests, and the CLI suite leaves specific gaps

There are no C++ unit tests — `src/test/` is untouched. That matches how `tapyrus-tx` is tested (`test/util/data/tapyrus-util-test.json` + `tapyrus-util-test.py`, no dedicated unit test), so it's consistent with the repo, and the helpers all live in an anonymous namespace in the `.cpp`, so unit-testing them would mean moving them into a header first. I'm not asking for that as-is.

What is worth closing is that the nine CLI cases don't touch two whole areas of the new code:

**None of `LoadPair`'s structural checks are exercised.** `CheckMoneyRange`, `CheckNoDuplicateInputs` and `CheckAllScriptsSanity` are new code with their own error paths, and no case reaches any of them — an output over `MAX_MONEY`, a spending tx with a duplicate outpoint, and an oversized or invalid-opcode script would be three cheap additions in the same JSON style.

**None of the flag-parsing branches beyond `standard` are exercised.** Worth adding:

- `--flags=0x...` hex form, plus the `>8` hex-digit rejection in `ParseHexUInt32`
- an unknown flag name
- `--flags=` with an empty spec — right now the `getline` loop never runs and it silently resolves to `0`, i.e. the same as `none`; erroring instead would be better, and either way the behaviour should be pinned by a test
- a `MaskToFlags`/`FlagsToNames` round-trip. The comment there notes the named flags sit at non-sequential bit positions (`nullfail` is bit 14), which is exactly the kind of mapping that's easy to get subtly wrong and currently has nothing checking it.

## Verified fixed from #456

- **Output resolution.** `LoadPair` (`:189-232`) now resolves the output from `spending.vin[nIn].prevout.n`, rejects a pair whose `prevout.hashMalFix` doesn't identify `to_spend`, and range-checks `prevout.n` — both halves of the bug are gone, with the reasoning recorded in the header.
- **`BUILD_SCRIPT_VERIFY` default.** Now OFF (`CMakeLists.txt:116`).
- **Locale dependence.** `std::strtoul` replaced by `ParseUInt32` and a hand-rolled hex parser, with a distinct message for a leading `-`.
- **The `STANDARD_SCRIPT_VERIFY_FLAGS` question** is answered better than by documenting it: the default is now the full 64-combination sweep. I checked the claim in the header holds — `kNamedFlags` is exactly STANDARD's five flags plus `SIGPUSHONLY`, and `MANDATORY_SCRIPT_VERIFY_FLAGS == 0` (`interpreter.h:93`), so the matrix genuinely contains both the mandatory and the exact standard combinations.
- **Stated limits.** The header now says a PASS is necessary-but-not-sufficient, and why.

A few things I checked that turned out fine, for the record: `CheckMoneyRange` matches `CheckTransaction`'s own output-value logic (`consensus/tx_verify.cpp:167-178`) including applying `MAX_MONEY` to colored outputs, so it won't wrongly reject colored transactions; the deleted `util_test_runner` ctest was dead (no `tapyrus-util` target exists, and `test/util/test_runner.py` doesn't either); and `check-tapyrus-verify`'s dependency on `tapyrus-util-test.py` is real, not a copy-paste leftover — `tapyrus-verify-test.py:24` importlib-loads it for the `bctester()` harness.

azuchi (changes requested):
Reviewed at `2d7cc187`. All eight test cases I asked for are there, and I confirmed the whole suite passes — I built `tapyrus-verify` from this branch locally and ran all 17 cases through `bctest`: 17 passed, 0 failed. Two problems remain, both in the CI/CMake side rather than the tool.

## 1. Dropping `tapyrus_wallet` from `tapyrus_util` breaks every GUI+Wallet job

```diff
 target_link_libraries(tapyrus_util
         PUBLIC
         tapyrus_clientversion
         PRIVATE
         core_interface
+        tapyrus_crypto
         univalue
         Boost::headers
-        $<TARGET_NAME_IF_EXISTS:tapyrus_wallet>
 )
```

Adding `tapyrus_crypto` is right and does fix the link I reported — `L-X86-Rel-No_GUI+No_Wallet` now builds `bin/tapyrus-verify` cleanly. Removing the `tapyrus_wallet` generator expression in the same commit is what turned five jobs red (`L-ARM-Rel-GUI+Wallet`, `L-ARM-Deb-GUI+USDT+Wallet`, `L-X86-Rel-GUI+USDT+Wallet`, `L-X86-Deb-GUI+USDT+Wallet`, `L-Cross-W-X86-Rel-GUI+Wallet`):

```
/usr/bin/ld: lib/libtapyrus_util.a(node.cpp.o): in function
  `interfaces::(anonymous namespace)::NodeImpl::getWallets()':
interfaces/node.cpp:224: undefined reference to
  `interfaces::MakeWallet(std::shared_ptr<CWallet> const&)'
collect2: error: ld returned 1 exit status
```

`src/interfaces/node.cpp` belongs to `tapyrus_util` (`src/CMakeLists.txt:98`) and calls `interfaces::MakeWallet`, which lives in `tapyrus_wallet`. That generator expression was the only thing declaring the edge; without it `libtapyrus_wallet.a` only ever appears *before* `libtapyrus_util.a` on the link line (visible in the failing `bin/tapyrus-qt` command), so the symbol goes unresolved.

Please restore `$<TARGET_NAME_IF_EXISTS:tapyrus_wallet>` and keep the `tapyrus_crypto` addition — that combination is what I verified locally.

(`tapyrus_util` depending on `tapyrus_wallet` is a layering inversion and I can see why it looked removable, but it's load-bearing. Untangling it means moving `interfaces/node.cpp` out of `tapyrus_util`, which is well outside this PR.)

## 2. The 17 CLI cases still never run in CI

`-DBUILD_SCRIPT_VERIFY=ON` in `daily-test.yml:614` gets the binary compiled on every native job, which is real progress, but that workflow never invokes the suite:

- the unit-test phase runs `./test_tapyrus` directly, not `ctest` (`daily-test.yml:757`)
- the only `ctest` call is `ctest --test-dir "$BASE_BUILD_DIR" -R "^test-tapyrus-qt$"` (`:767`)
- the util phase runs `python3 test/util/tapyrus-util-test.py` (`:784`) — `tapyrus-verify-test.py` is not run

So neither the `check-tapyrus-verify` target nor the `tapyrus_verify_test` ctest registered at `test/CMakeLists.txt:113` is ever reached, and none of the 17 cases execute.

`smoke-test.yml:247` already runs an unfiltered `ctest --output-on-failure --parallel "$CPU_COUNT"`, but doesn't set `BUILD_SCRIPT_VERIFY`, so the test isn't registered there. Adding `-DBUILD_SCRIPT_VERIFY=ON` to that job's `CMAKE_OPTS` gets both the build and the tests in one place, which is what I was after last time. The `daily-test.yml` change can stay — it just doesn't run anything.

## Not related to this PR

`macOS ARM64` and `macOS x86_64` fail at `brew install berkeley-db@4` falling back to 18.1.40 — that's #458, with #459 open against it.

azuchi (approved):

PR #459, opened 2026-09-03
cmake: find any installed Homebrew BerkeleyDB

### LINKED ISSUE #458: macOS CI broken: Homebrew dropped berkeley-db@4, FindBerkeleyDB.cmake hardcodes that formula
## Symptom

Every native macOS CI job now fails at CMake configure time, before building anything:

```
CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
  Could NOT find BerkeleyDB (missing: BerkeleyDB_LIBRARY BerkeleyDB_INCLUDE_DIR)
  (Required is at least version "4.8")
Call Stack (most recent call first):
  cmake/module/FindBerkeleyDB.cmake:103 (find_package_handle_standard_args)
  CMakeLists.txt:124 (find_package)
```

This is not caused by any pull request — it affects `master` and every branch equally. It is currently showing up as a red `macOS ARM64` check on unrelated PRs (e.g. #455, which only bumps `actions/upload-artifact`).

## Cause

Homebrew removed the `berkeley-db@4` formula. Two smoke-test runs the same day pin the window down:

| Time (UTC, 2026-09-02) | Run | Result |
| --- | --- | --- |
| 12:06 | [33628082194](https://github.com/chaintope/tapyrus-core/actions/runs/33628082194) | `Pouring berkeley-db@4--4.8.30.arm64_sonoma.bottle.2.tar.gz` → **success** |
| 12:47 | [33631871428](https://github.com/chaintope/tapyrus-core/actions/runs/33631871428) | `berkeley-db@4` not found; brew offers only `berkeley-db` and `berkeley-db@5` → **failure** |

## Why the existing fallback doesn't save us

`.github/workflows/smoke-test.yml:122-128` already anticipates this:

```bash
# Try berkeley-db@4 first, fall back to berkeley-db
if brew install --quiet berkeley-db@4 2>/dev/null; then
  echo "Installed berkeley-db@4"
else
  brew install --quiet berkeley-db
  echo "Installed berkeley-db (will use --with-incompatible-bdb)"
fi
```

The fallback does install `berkeley-db` 18.1.40, but nothing then tells CMake where it is. The only follow-up is `smoke-test.yml:207-208` (and `daily-test.yml:650-651`):

```bash
if ! brew list berkeley-db@4 &>/dev/null; then
  CMAKE_OPTS+=("-DWARN_INCOMPATIBLE_BDB=OFF")
fi
```

`WARN_INCOMPATIBLE_BDB` only suppresses the version-mismatch *warning* at `CMakeLists.txt:130`. It has no effect on locating the library, which is what actually fails here.

The reason the keg is never found is `cmake/module/FindBerkeleyDB.cmake:6-10`, which hardcodes the formula name:

```cmake
if(CMAKE_HOST_APPLE)
  include(BrewHelper)
  find_brew_prefix(_BerkeleyDB_homebrew_prefix berkeley-db@4)
endif()
```

With `berkeley-db@4` gone, `_BerkeleyDB_homebrew_prefix` is empty, so the `HINTS` in the `find_path`/`find_library` calls below are empty too and `/opt/homebrew/opt/berkeley-db` is never searched. So the fallback branch has apparently never actually been exercised until now.

## Possible directions

Not proposing one over the others — this needs a deliberate call, because BDB 18.1 wallet files are not interchangeable with 4.8 ones, which is exactly what `WARN_INCOMPATIBLE_BDB` exists to flag:

1. **Point CMake at the fallback keg** — have the workflows pass `-DBerkeleyDB_ROOT="$(brew --prefix berkeley-db)"` on the fallback path, and/or teach `FindBerkeleyDB.cmake` to try `berkeley-db` after `berkeley-db@4`. Cheapest fix, but it means macOS CI silently builds wallets in an incompatible format.
2. **Keep 4.8 on macOS CI** — install it from a tap that still carries it, or from a pinned bottle URL. Preserves what the jobs used to test.
3. **Build BDB through `depends` on macOS CI** — `depends/packages/bdb.mk` already pins 4.8.30 and is used everywhere else; this removes the Homebrew dependency for BDB entirely, at the cost of build time on those jobs.

Whichever way it goes, it would be worth not leaving `FindBerkeleyDB.cmake`'s hardcoded `berkeley-db@4` as the single point of failure, since the same thing will happen again the next time a formula is renamed or dropped.

## Affected files

- `cmake/module/FindBerkeleyDB.cmake:9`
- `.github/workflows/smoke-test.yml:122-128`, `:207-208`
- `.github/workflows/daily-test.yml:650-651`

Naviabheeman:
Fix #458
macOS CI was failing because Homebrew removed the berkeley-db@4 formula. FindBerkeleyDB.cmake's Homebrew lookup was hardcoded to that exact name, so even though the CI script already falls back to installing plain berkeley-db, CMake never looked there and failed to locate BDB at all.

Add find_brew_prefix_any_version() to BrewHelper.cmake: queries the list of installed Homebrew formulae for any berkeley-db or berkeley-db@N formula that's actually installed, tries version-suffixed ones first (natural-sorted, so @4 before @18), then the plain formula last, instead of guessing one hardcoded name. FindBerkeleyDB.cmake now calls this instead of the single-name lookup.

Verified against a real Homebrew install (berkeley-db, @4, and @5 all present) and against a synthetic "only plain berkeley-db installed" case matching the actual CI failure -- both resolve correctly, and a full -DWITH_BDB=ON configure finds berkeley-db@4 4.8.30 as before.

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -645,12 +645,12 @@ jobs:
           fi
 
           # Platform-specific options
-          if [ "${{ matrix.config.platform }}" == "macos" ]; then
-            # Native macOS builds
-            if ! brew list berkeley-db@4 &>/dev/null; then
-              CMAKE_OPTS+=("-DWARN_INCOMPATIBLE_BDB=OFF")
-            fi
-          elif [ "${{ matrix.config.target_platform }}" == "macos" ]; then
+          #
+          # Native macOS builds need no BDB warning override here: this
+          # job builds BDB from depends (always 4.8.30, from source),
+          # never from Homebrew, so the version this override used to
+          # guard against was never actually in play.
+          if [ "${{ matrix.config.target_platform }}" == "macos" ]; then
             # Cross-compilation to macOS from Linux
             CMAKE_OPTS+=("-DCMAKE_SYSTEM_NAME=Darwin")
             if [ "${{ matrix.config.target_arch }}" == "arm64" ]; then
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -119,12 +119,22 @@ jobs:
           fi
           clang --version
           brew install --quiet python@3 || brew link --overwrite python@3
-          # Try berkeley-db@4 first, fall back to berkeley-db
-          if brew install --quiet berkeley-db@4 2>/dev/null; then
+          # berkeley-db@5 is the validated baseline (Sleepycat-licensed,
+          # wallet-format compatible with @4 via shared DB_BTREEVERSION);
+          # @4 is a compatible fallback for runners that still pour it.
+          # Never fall back further to the plain berkeley-db formula --
+          # it tracks Homebrew's latest (presently 18.1.x), which is
+          # AGPL-3.0-only and incompatible with this project's MIT
+          # license; CMake's BerkeleyDB version ceiling rejects it anyway,
+          # so installing it here would just fail the configure step
+          # later instead of failing fast here.
+          if brew install --quiet berkeley-db@5 2>/dev/null; then
+            echo "Installed berkeley-db@5"
+          elif brew install --quiet berkeley-db@4 2>/dev/null; then
             echo "Installed berkeley-db@4"
           else
-            brew install --quiet berkeley-db
-            echo "Installed berkeley-db (will use --with-incompatible-bdb)"
+            echo "::error::Neither berkeley-db@5 nor berkeley-db@4 could be installed" >&2
+            exit 1
           fi
           brew install --quiet coreutils ninja pkgconf gnu-getopt ccache boost libevent zeromq qrencode
           echo 'export PATH="/opt/homebrew/bin:$PATH"' >> $HOME/.bashrc
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -212,12 +212,10 @@ jobs:
             # Native builds (Linux/macOS)
             CMAKE_OPTS+=("-DENABLE_TESTS=ON" "-DCP2SH_ACTIVATION_TEST_HEIGHT=120")
 
-            # Disable BDB version warning for macOS if using newer version
-            if [ "${{ matrix.config.platform }}" == "macos" ]; then
-              if ! brew list berkeley-db@4 &>/dev/null; then
-                CMAKE_OPTS+=("-DWARN_INCOMPATIBLE_BDB=OFF")
-              fi
-            fi
+            # No -DWARN_INCOMPATIBLE_BDB=OFF here: this job's own install
+            # step (above) only ever installs berkeley-db@5 or @4, and
+            # CMakeLists.txt's BDB warning no longer fires for either --
+            # there is nothing left for this override to suppress.
 
             # Add platform-specific options
             CMAKE_OPTS+=("-DCMAKE_CXX_FLAGS=-Wno-error=unused-member-function")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,12 +119,18 @@ option(BUILD_UTILS "Build tapyrus-tx executable." ${ENABLE_TESTS})
 
 option(ENABLE_WALLET "Build wallet functionality in tapyrusd, gui and tests." ON)
 option(WITH_BDB "Enable Berkeley DB (BDB) wallet support." ON)
-cmake_dependent_option(WARN_INCOMPATIBLE_BDB "Warn when using a Berkeley DB (BDB) version other than 4.8." ON "WITH_BDB" OFF)
+cmake_dependent_option(WARN_INCOMPATIBLE_BDB "Warn when using a Berkeley DB (BDB) version other than 4.8 or 5.3." ON "WITH_BDB" OFF)
 if(ENABLE_WALLET AND WITH_BDB)
-  find_package(BerkeleyDB 4.8 MODULE REQUIRED)
+  # Upper bound is a license floor, not a technical one: Homebrew's BDB
+  # major versions above 5 (the plain "berkeley-db" formula is presently
+  # 18.1.x) are AGPL-3.0-only, incompatible with this project's MIT
+  # license, unlike the Sleepycat-licensed 4.8/5.3. See
+  # cmake/module/FindBerkeleyDB.cmake and BrewHelper.cmake's
+  # find_brew_prefix_any_version() for the Homebrew-formula side of this.
+  find_package(BerkeleyDB 4.8...5.99 MODULE REQUIRED)
   set(USE_BDB ON)
-  if(NOT BerkeleyDB_VERSION VERSION_EQUAL 4.8)
-    message(WARNING "Found Berkeley DB (BDB) other than 4.8.\n"
+  if(NOT (BerkeleyDB_VERSION VERSION_EQUAL 4.8 OR BerkeleyDB_VERSION VERSION_EQUAL 5.3))
+    message(WARNING "Found Berkeley DB (BDB) other than the validated 4.8 or 5.3.\n"
                     "BDB (legacy) wallets opened by this build will not be portable!"
     )
     if(WARN_INCOMPATIBLE_BDB)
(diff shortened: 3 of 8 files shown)


azuchi (changes requested):
Reviewed at `2f359460`. The approach is right — asking Homebrew what's actually installed instead of hardcoding one formula name is the correct fix for #458, and the ordering logic does what its comment claims.

I extracted the selection logic and ran it standalone under CMake 3.28: input `berkeley-db;berkeley-db@18;berkeley-db@4;berkeley-db@5;boost;berkeley-dbx@9` yields `berkeley-db@4;berkeley-db@5;berkeley-db@18;berkeley-db`. So `COMPARE NATURAL` really does put `@4` ahead of `@18`, the anchored regex correctly rejects `berkeley-dbx@9`, and the plain formula is appended after the sort so it stays last. `list(SORT)` on an empty list is a no-op, so a machine with no versioned formula installed doesn't error. `COMPARE NATURAL` needs CMake >= 3.18 and this project requires 3.22 (`CMakeLists.txt:11`), so that's fine too. `find_brew_prefix` still has a caller (`src/qt/locale/CMakeLists.txt:7`), so nothing is left dead.

Three things before this merges.

## 1. The green macOS jobs never exercise the new path

In this PR's run (33760630276, 13:41):

```
==> Pouring berkeley-db@4--4.8.30.arm64_sonoma.bottle.2.tar.gz
🍺  /opt/homebrew/Cellar/berkeley-db@4/4.8.30
Installed berkeley-db@4
-- Found BerkeleyDB: /opt/homebrew/opt/berkeley-db@4/lib/libdb_cxx-4.8.dylib (found suitable version "4.8.30")
```

`berkeley-db@4` installed successfully, so the old hardcoded `find_brew_prefix(... berkeley-db@4)` would have passed this run identically. Compare the run three hours earlier the same day (33746806745, 10:58, on #454):

```
berkeley-db
berkeley-db@5
To install berkeley-db, run:
  brew install berkeley-db
==> Pouring berkeley-db--18.1.40_2.arm64_sonoma.bottle.tar.gz
Installed berkeley-db (will use --with-incompatible-bdb)
```

Formula availability is flapping between runners on the same day, which is presumably an image/index difference. The upshot is that the exact scenario this PR exists to fix is still untested. Worth confirming by hand with `@4` uninstalled, or with a temporary job that installs only the fallback, before trusting the green.

## 2. The fallback links an AGPL-licensed Berkeley DB

From formulae.brew.sh:

| formula | version | license |
|---|---|---|
| `berkeley-db` | 18.1.40 | **AGPL-3.0-only** |
| `berkeley-db@5` | 5.3.28 | Sleepycat |
| `berkeley-db@4` | — | removed from homebrew-core (though some runners still pour it, per above) |

This change accepts the plain formula as the last candidate, and `smoke-test.yml:123-127` — untouched here — installs plain `berkeley-db` when `@4` fails. Before this PR that combination simply failed to configure. After it, CMake finds the 18.1.40 keg and links MIT-licensed tapyrus against an AGPL-3.0-only library.

Please point the CI fallback at `berkeley-db@5` instead. It's also the better technical choice: 5.3.28 and 4.8.30 both define `DB_BTREEVERSION 9`, so a wallet.dat stays readable by a 4.8-linked build, whereas 18.1 is well outside anything this project has validated.

It would also be worth warning when the selected formula is the plain/unversioned one, either in the helper or at the `FindBerkeleyDB.cmake` call site, so a local build doesn't silently pick it up either.

## 3. Nothing resolves an 18.x keg even once the prefix is found

`FindBerkeleyDB.cmake:23-30`'s `PATH_SUFFIXES` (`4.8 48 db4.8 4 db4 5.3 db5.3 5 db5`) and the `NAMES` lists from `:44` (`db_cxx-4.8 db4_cxx db48 db_cxx-5.3 db_cxx-5 db_cxx libdb48`) have no 18.x entry, so picking up an 18.1 keg depends entirely on Homebrew shipping unversioned `db_cxx.h` and `libdb_cxx.dylib`. Per finding 1, CI has never reached that code, so it's unverified.

Adding `db_cxx-18.1 db_cxx-18` to `NAMES` is cheap insurance. Fixing finding 2 makes the question moot instead, since `db_cxx-5.3` and `db_cxx-5` are already in the list — which is the outcome I'd prefer.

azuchi:
Reviewed at `28040cfe`. All three points from the last round are resolved.

The macOS jobs now genuinely exercise the path this PR exists for:

```
==> Pouring berkeley-db@5--5.3.28_1.arm64_sonoma.bottle.tar.gz
Installed berkeley-db@5
-- Found BerkeleyDB: /opt/homebrew/opt/berkeley-db@5/lib/libdb_cxx-5.3.dylib
   (found suitable version "5.3.28", required range is "4.8...5.99")
```

`find_brew_prefix_any_version` resolves `@5`, and the existing `db_cxx-5.3` entry in `NAMES` picks the library up, so the 18.x question is moot. The AGPL fallback is gone in both places — CI now fails fast rather than installing the plain formula, and the helper never considers it. Adding `HANDLE_VERSION_RANGE` matters and is easy to miss: without it `find_package_handle_standard_args` silently ignores the upper bound entirely.

I re-ran the new selection logic standalone under CMake 3.28:

| installed formulae | result |
| --- | --- |
| `@5;@4;@18;berkeley-db;boost;@5.3` | candidates `@5;@4`, best `@5` |
| `@18;berkeley-db` | no candidate (falls through to system paths, then the ceiling rejects it) |
| `@4` only | `@4` |

`if(... MATCHES ... AND CMAKE_MATCH_1 LESS_EQUAL MAX_MAJOR)` evaluates in the right order and `ORDER DESCENDING` prefers the newest allowed major, as intended.

Three things left.

## 1. The version ceiling applies everywhere, and the documented escape hatch doesn't exist

`CMakeLists.txt:127` and `src/wallet/CMakeLists.txt:9` apply `4.8...5.99` on every platform, not just Homebrew, so any environment carrying only BDB 6.x or newer now fails to configure where it previously configured with a warning. Meanwhile five places in the docs tell users to work around exactly that with `-DWITH_INCOMPATIBLE_BDB=ON`:

```
doc/build-unix.md:107, :300, :302
doc/build-osx.md:103
doc/build-cmake.md:41
```

That option does not exist in any CMake file — only `WARN_INCOMPATIBLE_BDB` does. The docs were already wrong before this PR, but this change turns "docs are wrong, the build still works with a warning" into "docs are wrong and there is no way through". `build-unix.md:301-302` names Arch Linux, whose `db` package is 6.2, as the concrete case.

Worth being precise about what the ceiling protects, because it isn't the project's own licensing. Release binaries link the Sleepycat-licensed 4.8.30 from `depends`, so they are unaffected by this range either way. AGPL obligations attach on conveying a combined work, so offering a build option is not itself a problem: a user building locally conveys nothing, and a third party who distributes an AGPL-linked build takes on those obligations themselves. What the ceiling actually prevents is someone unknowingly producing such a build from system libraries — a good goal, but one that argues for a loud opt-out rather than a hard stop.

Suggest keeping `4.8...5.99` as the default and adding the option the docs already name:

- `WITH_INCOMPATIBLE_BDB` (default `OFF`) lifts the upper bound
- when it's `ON`, emit a `message(WARNING)` saying the build will link an AGPL-3.0-only library, that distributing the result puts the whole combined work under AGPL, and that it must not be used for official release binaries

That makes the five existing doc references true rather than needing deletion. If you'd rather keep the hard failure, those references need updating instead — either way the current mismatch shouldn't ship.

While you're in there, `doc/build-osx.md:37-38` still says "It is recommended to use Berkeley DB 4.8" and points at `contrib/install_db4.sh`, with no mention that Homebrew dropped `berkeley-db@4` and `@5` is now the baseline this PR validates.

## 2. The tightened regex no longer matches a dotted formula suffix

`^${BASE_NAME}@([0-9]+)$` drops what the old `[0-9.]+` accepted, so `berkeley-db@5.3` would be skipped. It doesn't matter for berkeley-db — homebrew-core only carries `@4` and `@5` — but the function is generic, and Homebrew does use `name@major.minor` elsewhere (`python@3.11`). `@([0-9]+)(\.[0-9]+)?$` keeps the major available for the `MAX_MAJOR` comparison while tolerating a minor.

## 3. The CI's `WARN_INCOMPATIBLE_BDB=OFF` condition no longer expresses anything

`smoke-test.yml:217` and `daily-test.yml:650` still gate on `if ! brew list berkeley-db@4`, which is now always true because CI installs `@5`. Harmless — 5.3 no longer triggers the warning at all — but the condition reads as if it still means something.

azuchi (approved):

PR #460, opened 2026-09-07
Upgrade Berkeley DB dependency from 4.8.30 to 5.3.28 

Naviabheeman:
Upgrade the wallet's Berkeley DB dependency from 4.8.30 to 5.3.28 across the depends build system, CI, and docs. The CMake version gate upgrade in in PR #459.  This PR moves what depends/ actually builds to 5.3, while keeping 4.8 usable for anyone with an existing system install (-DWITH_INCOMPATIBLE_BDB=ON remains the escape hatch for anything outside that range).

- depends/packages/bdb.mk: version, download URL, and sha256 bumped to 5.3.28; build target names updated to libdb_cxx-5.3.a/libdb-5.3.a (BDB's own LIBVERSION convention).
- depends/patches/bdb/5.3.28/clang_cxx_11.patch: regenerated against real 5.3.28 source — 5.3 nests everything under a new top-level src/ directory that 4.8 didn't have, so the patch's file paths needed updating (same atomic_init→atomic_init_db rename as before, just repathed).
- CI (daily-test.yml, weekly-heavy-tests.yml): updated libdb_cxx-4.8.a/libdb-4.8.a references to -5.3.a.
- Docs (dependencies.md, build-cmake.md, build-unix.md): 5.3 is now the recommended version; Ubuntu gets a direct libdb5.3-dev/libdb5.3++-dev package instead of the old bitcoin PPA; Fedora's unsuffixed libdb-devel already ships 5.3.28; 4.8 stays documented as supported (both share DB_BTREEVERSION 9. SO the same wallet file can be opened using a tapyrus_wallet using any version. However after a wallet is opened using bdb5.3, it can no longer be opened using bdb4.8) 
- contrib/install_db4.sh: kept as a 4.8-specific legacy helper (matches the docs' framing); fixed its trailing usage note, which still referenced pre-CMake ./configure flags, to the current cmake -DBerkeleyDB_ROOT=... invocation.
- Real native depends + CMake builds confirmed on both Linux (aarch64) and macOS (arm64): BDB 5.3.28 compiles, patches apply cleanly, tapyrusd links against it statically.

### DIFF
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -227,6 +227,12 @@ jobs:
           git checkout -- depends/
           git clean -fdx -- depends/patches/ depends/packages/ depends/hosts/ depends/builders/
 
+      - name: Re-checkout depends/packages and depends/patches
+        # A stale restore-keys cache hit (e.g. right after a version bump)
+        # can overwrite these with old content, so make rebuilds the wrong
+        # version. Force them back to HEAD after the cache restore.
+        run: git checkout -- depends/packages depends/patches
+
       - name: Restore Benchmark History cache
         # Benchmarks only run on scheduled/manually-dispatched runs, not on
         # every PR (see the gated benchmark step further down). Gated on the
@@ -417,8 +423,8 @@ jobs:
 
             # Wallet dependencies (Berkeley DB)
             if [ "${{ matrix.config.enable_wallet }}" == "true" ]; then
-              REQUIRED_FILES+=("$DEPENDS_PREFIX/lib/libdb_cxx-4.8.a")
-              REQUIRED_FILES+=("$DEPENDS_PREFIX/lib/libdb-4.8.a")
+              REQUIRED_FILES+=("$DEPENDS_PREFIX/lib/libdb_cxx-5.3.a")
+              REQUIRED_FILES+=("$DEPENDS_PREFIX/lib/libdb-5.3.a")
               REQUIRED_FILES+=("$DEPENDS_PREFIX/include/db_cxx.h")
               REQUIRED_FILES+=("$DEPENDS_PREFIX/include/db.h")
             fi
@@ -686,7 +692,7 @@ jobs:
           # Platform-specific options
           #
           # Native macOS builds need no BDB warning override here: this
-          # job builds BDB from depends (always 4.8.30, from source),
+          # job builds BDB from depends (always 5.3.28, from source),
           # never from Homebrew, so the version this override used to
           # guard against was never actually in play.
           if [ "${{ matrix.config.target_platform }}" == "macos" ]; then
@@ -708,7 +714,7 @@ jobs:
           # but not the specific header/library paths).
           if [ "${{ matrix.config.enable_wallet }}" == "true" ]; then
             CMAKE_OPTS+=("-DBerkeleyDB_INCLUDE_DIR=$DEPENDS_PREFIX/include")
-            CMAKE_OPTS+=("-DBerkeleyDB_LIBRARY=$DEPENDS_PREFIX/lib/libdb_cxx-4.8.a")
+            CMAKE_OPTS+=("-DBerkeleyDB_LIBRARY=$DEPENDS_PREFIX/lib/libdb_cxx-5.3.a")
           fi
 
           # Configure and build
--- a/.github/workflows/weekly-heavy-tests.yml
+++ b/.github/workflows/weekly-heavy-tests.yml
@@ -501,10 +501,11 @@ jobs:
         #
         # The rest of depends/packages/*.mk is left pinned at its current
         # version:
-        #   - Berkeley DB: find_package(BerkeleyDB 4.8 MODULE REQUIRED)
-        #     (CMakeLists.txt:124) is already satisfied by the 4.8.30 pinned in
-        #     bdb.mk (4.8.30 >= 4.8) -- no older, meaningfully-different 4.8.x
-        #     release to fetch instead, so bdb.mk is left as-is.
+        #   - Berkeley DB: find_package(BerkeleyDB 4.8...5.99 MODULE REQUIRED)
+        #     (CMakeLists.txt:149) is already satisfied by the 5.3.28 pinned
+        #     in bdb.mk, and doc/dependencies.md documents no separate
+        #     "Minimum required" for it (Version and Minimum required are
+        #     both 5.3.28), so bdb.mk is left as-is.
         #   - Qt: find_package(Qt 6.0 ...) (CMakeLists.txt:180) is 5+ minor
         #     versions behind the 6.10.1 pinned in qt.mk. Swapping in Qt 6.0.0
         #     isn't a one-line version+hash change the way the five above
@@ -654,7 +655,7 @@ jobs:
             -DZEROMQ_LIBRARY="$DEPENDS_PREFIX/lib/libzmq.a"
             -DBerkeleyDB_ROOT="$DEPENDS_PREFIX"
             -DBerkeleyDB_INCLUDE_DIR="$DEPENDS_PREFIX/include"
-            -DBerkeleyDB_LIBRARY="$DEPENDS_PREFIX/lib/libdb_cxx-4.8.a"
+            -DBerkeleyDB_LIBRARY="$DEPENDS_PREFIX/lib/libdb_cxx-5.3.a"
           )
           if [ "${{ matrix.config.platform }}" == "linux" ]; then
             CMAKE_OPTS+=(-DENABLE_TRACING=ON)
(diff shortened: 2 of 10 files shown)


azuchi:
Reviewed at `805b4b14`. The mechanical side of the bump is in good shape, and a few things I'd have flagged as unverified are already handled.

I re-checked the parts that are cheap to get wrong:

- The sha256 matches — I downloaded `db-5.3.28.NC.tar.gz` from Oracle in full and got `76a25560d9e52a198d37a31440fd07632b5f1f8f9f2b6d5438f4bc3e7c9013ef`.
- The regenerated `clang_cxx_11.patch` is correct, not just approximately right: 5.3.28 moved everything under a new top-level `src/`, and the repathed patch applies from the extract root with **zero fuzz and zero offsets** on all six files.
- `mingw_winioctl_case.patch` is a real 5.3-only problem (`win_db.h` includes `<WinIoCtl.h>`; mingw-w64 ships `winioctl.h`) and the patch header records exactly that. Good catch — cross-compilation was the part of this bump I'd have expected to bite.
- CI's `libdb_cxx-4.8.a` references are updated in both workflows, and `contrib/install_db4.sh`'s trailing note no longer prints pre-CMake `./configure` flags.

Four things before this merges.

## 1. The portability claim only holds in one direction

`doc/build-unix.md:94-96` (and the same wording at `:203-206` and `:307`) says:

> both share the same on-disk wallet format, `DB_BTREEVERSION` 9, so a wallet built against one opens fine with the other

The downgrade direction doesn't work. I built 5.3.28 with this PR's own patches and tested it against the 4.8.30 from `depends`, driving the environment exactly as `BerkeleyEnvironment::Open` does (same flags, same `set_lg_dir`, same `DB_LOG_AUTO_REMOVE`):

| | result |
| --- | --- |
| 4.8 creates → 4.8 reads | OK, 200 records |
| 4.8 creates → **5.3 reads** (upgrade) | **OK, 200 records** |
| 5.3 creates → **4.8 reads** (downgrade) | **fails** |

```
Unacceptable log file .../database/log.0000000001: unsupported log version 19
Invalid log file: log.0000000001: Invalid argument
PANIC: Invalid argument
DbEnv::open: DB_RUNRECOVERY: Fatal error, run database recovery
```

`DB_LOG_AUTO_REMOVE` (`src/wallet/db.cpp:157`) doesn't remove the current log, so `log.0000000001` survives even a clean shutdown, and 4.8 tops out at log version 16.

The `DB_BTREEVERSION` 9 part is true — opening the 5.3-written `wallet.dat` with 4.8 and **no environment** reads all 200 records back. What breaks is the environment/log layer, not the file. But from a user's point of view the effect is "I upgraded, then rolled back to the previous release, and the node won't start."

Please scope the docs to what actually holds — the file format is shared, but a wallet directory last opened by 5.3 will not open under 4.8 — and say what to do about it (remove `database/`, or run the old binary with `-salvagewallet`). Given release binaries will start writing 5.3 wallets, this belongs in the release notes too, not only in `build-unix.md`.

## 2. The S3 mirror doesn't have the 5.3.28 tarball yet

```
db-5.3.28.NC.tar.gz -> HTTP 403
db-4.8.30.NC.tar.gz -> HTTP 206
```

A 403 can't distinguish "absent" from "not public" while ListBucket is denied, but the 4.8 tarball answers 206 to the same request shape from the same bucket, so absent is the reasonable reading. This isn't theoretical: in this PR's own macOS run the qrencode upstream returned 404 and the S3 fallback is what rescued the build. Please upload it to `FALLBACK_DOWNLOAD_PATH` before merging — same ordering as #452.

## 3. The description documents a job that isn't in the diff

> weekly-heavy-tests.yml also gains a node-upgrade-test job (manual workflow_dispatch only) plus a build-image job that docker-testnet-sync now shares, intended to validate that a wallet built under an old BDB version reopens cleanly under the new one.

`grep -rn "node-upgrade-test\|build-image" .github/workflows/` finds nothing on this branch, and the `weekly-heavy-tests.yml` diff is ten lines — the library-name updates plus the re-checkout step. Worth noting that the job as described is precisely what would have caught finding 1. Either add it or drop the paragraph.

## 4. The patch layout collides with #453

#453 moves depends patches into version subdirectories and makes a flat patch a hard build error:

```make
$(if $(wildcard $(dir $($(1)_patches_path))*.patch),$(error Patch file(s) found directly under ... ))
```

- #453 has `depends/patches/bdb/4.8.30/clang_cxx_11.patch`
- this PR has `depends/patches/bdb/clang_cxx_11.patch` and `depends/patches/bdb/mingw_winioctl_case.patch`, both flat

If #453 lands first, this branch fails to build immediately. If this lands first, #453's rebase has to move both files into `depends/patches/bdb/5.3.28/`. Worth agreeing the merge order and who absorbs the move.

## Also

The two failing macOS jobs are `rpc_pstt.py --scheme SCHNORR` timing out at 245-248s. `M-ARM-Rel-GUI+Wallet` and `M-ARM-Rel-No_GUI+Wallet` are green on master's 09-07 and 09-08 runs, so this reproduces only here, and it's the wallet-enabled macOS jobs specifically. Could be a slow runner, could be 5.3 being slower on this workload — worth a re-run to separate the two.

Separately, `doc/dependencies.md:21` still lists 4.8 as the minimum, but after this change nothing in CI builds 4.8 at all. #453 redefines that column as the floor CI actively proves, which makes the gap more visible; either give 4.8 a job that backs the claim or reword it.

azuchi (changes requested):
Marking this changes-requested to match the findings in https://github.com/chaintope/tapyrus-core/pull/460#issuecomment-5594163797 — the detail is all there rather than repeated here.

The two that make it blocking:

1. `doc/build-unix.md`'s "a wallet built against one opens fine with the other" only holds in the upgrade direction. Measured against this PR's own 5.3.28 build: 4.8-written wallet opens under 5.3, but a 5.3-written wallet directory fails under 4.8 with `unsupported log version 19` / `DB_RUNRECOVERY`, even after a clean shutdown. Once release binaries write 5.3 wallets this is a one-way door for anyone rolling back, so the docs need scoping and the release notes need a recovery procedure.

2. `db-5.3.28.NC.tar.gz` isn't on the S3 mirror yet, and the mirror is load-bearing — this PR's own macOS run fell back to it when a qrencode upstream fetch 404'd.

3 (the description documenting a `node-upgrade-test` job that isn't in the diff) and 4 (flat patch directories colliding with #453's version-scoped layout, which makes a flat patch a hard build error) both need a decision but aren't blockers on their own.

azuchi:
@Naviabheeman fix conflicts.

azuchi:
Reviewed at `e1c6d00a`. Three of the four findings are resolved, CI is 14/14 (the macOS `rpc_pstt.py` timeout went away with the rebase onto master, which now carries #461), and the docs came out better than I asked for.

**The one-way framing** is now explicit in `doc/build-unix.md`, with the actual error text and the recovery procedure:

> once a wallet directory has been opened by 5.3, its transaction log is a version 4.8 cannot read ... (`PANIC: ... unsupported log version`). If you need to go back to a 4.8-linked build, remove the wallet's `database/` subdirectory first (or run the old binary with `-salvagewallet`).

And the release-version mapping table in `doc/dependencies.md` puts it in front of the people it affects, with "back up the wallet directory first" spelled out. That's the part I thought belonged in the release notes, and it's now in the docs a user actually reads. `Minimum tested` moving to 5.3.28 also closes the gap where 4.8 was claimed as a floor that nothing builds.

**The S3 mirror** is populated — `db-5.3.28.NC.tar.gz` answers 206 now.

**The patch layout** fits #453's scheme: patches under `depends/patches/bdb/5.3.28/`, `bdb.mk` using the same `ifeq`-per-version `_patches` plus generic `$(foreach patch,...)` pattern as `libevent.mk`, and `4.8.30/clang_cxx_11.patch` kept so a version swap still has its patch. `e1c6d00a0` also caught the `libdb_cxx-4.8.a` references still left in `weekly-heavy-tests.yml`, including the minimum-versions job's comment.

## Please update the PR description

It hasn't been touched since the first push, so it now disagrees with the branch in three places:

- `depends/patches/bdb/clang_cxx_11.patch` — the file is at `depends/patches/bdb/5.3.28/clang_cxx_11.patch`.
- "weekly-heavy-tests.yml also gains a node-upgrade-test job ... plus a build-image job that docker-testnet-sync now shares" — `grep -rn "node-upgrade-test\|build-image" .github/workflows/` still finds nothing, and the `weekly-heavy-tests.yml` diff is only the 4.8 → 5.3 substitutions.
- "4.8 stays documented as supported (both share `DB_BTREEVERSION` 9, **so wallets are portable between them**)" — this is the exact claim the documentation in this same PR now retracts. Leaving it in the description means the merge commit reproduces the misunderstanding the PR fixed.

The third one is the one I'd really like corrected before merge; the description is what ends up in the merge commit and in `git log`.

## Minor

`doc/build-unix.md:205-207` says "Berkeley DB 4.8 is also **validated** if you specifically need it". Now that nothing in CI builds 4.8 anywhere, "validated" reads stronger than what's actually true. Keeping `4.8.30/clang_cxx_11.patch` does mean a manual version swap still works, so something like "still supported, though no longer exercised in CI" would match reality.

Other than the description, I don't think anything else needs to change here.

Naviabheeman:
The description has been updated to match the current behavior.

azuchi (approved):

PR #462, opened 2026-09-10
Fix wallet_coloredcoin test failure

Naviabheeman:
Fixes a race in wallet_coloredcoin.py's only_token filter test that could mine a transaction the test requires to stay unconfirmed.

setup_network() connects nodes 0, 1, and 2 in a full mesh (0-1, 1-2, 0-2). The test only disconnected the direct 0-1 link before issuing res1 on node 0, leaving node 0 reachable via node 2 — so node 2 could still relay res1 on to node 1 (0 → 2 → 1). Whether that relay landed before nodes[1].generate(2) read its mempool was a race: usually lost locally, but hit often enough under CI load to mine res1 when the test's assertions require it to still be unconfirmed on node 1.

The fix isolates node 0 from both peers (0-1 and 0-2) before issuing res1, guaranteeing it can't reach node 1's mempool while isolated. Since p2p relay only announces transactions that arrive after a peer is already connected (not a reconnecting peer's pre-existing mempool), the nodes are reconnected afterward and a new relay_mempool() helper re-submits res1 to nodes 1 and 2 by RPC directly, so the subsequent sync_all() can actually converge — plain p2p reconnection alone would never propagate it.

relay_mempool() retries in passes (bounded by mempool size) since a child transaction can't be accepted on a node before its parent is, and on final failure reports the last RPC error seen per unsynced txid rather than swallowing it, so a genuine failure (fee rule, missing input, etc.) is distinguishable from a transient ordering issue.

### DIFF
--- a/test/functional/wallet_coloredcoin.py
+++ b/test/functional/wallet_coloredcoin.py
@@ -679,6 +679,7 @@ class WalletColoredCoinTest(BitcoinTestFramework):
         node before its parent is.
         """
         remaining = set(from_node.getrawmempool())
+        last_errors = {}
         for _ in range(len(remaining) + 1):
             if not remaining:
                 return
@@ -688,11 +689,14 @@ class WalletColoredCoinTest(BitcoinTestFramework):
                     if txid not in node.getrawmempool():
                         try:
                             node.sendrawtransaction(raw_tx)
-                        except JSONRPCException:
-                            pass  # parent not yet present on this node; retry next pass
+                        except JSONRPCException as e:
+                            # parent not yet present on this node; retry next pass
+                            last_errors[txid] = e
                 if all(txid in node.getrawmempool() for node in to_nodes):
                     remaining.discard(txid)
-        raise AssertionError("relay_mempool could not sync: %s" % remaining)
+        raise AssertionError("relay_mempool could not sync: %s" % {
+            txid: str(last_errors[txid]) for txid in remaining if txid in last_errors
+        })
 
     def test_only_token_filter(self):
 
--- a/test/functional/wallet_coloredcoin.py
+++ b/test/functional/wallet_coloredcoin.py
@@ -26,6 +26,7 @@ import json
 import math
 import time
 
+from test_framework.authproxy import JSONRPCException
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (
     assert_array_result,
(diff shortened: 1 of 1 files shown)


azuchi:
Reviewed at `c9eb5854`. The diagnosis holds up and the fix is the right shape — two small things, neither blocking.

The mesh claim checks out: `wallet_coloredcoin.py:120-122` wires 0-1, 1-2 **and** 0-2, so the old `disconnect_nodes(0, 1)` never isolated node 0 at all and left 0 → 2 → 1 open. `disconnect_nodes` itself is synchronous (`util.py:395` waits for the peer to disappear from `getpeerinfo`), so there's no residual window between the disconnect and `issuetoken`.

`relay_mempool` is justified too, for what it's worth. Once node 0 is genuinely isolated, res1 exists only there, and the following `sync_all([self.nodes[0:3]])` can never converge on mempools. Swapping that for `sync_blocks` wouldn't work either — it would drop the wait that res2 relies on to reach node 0 — so re-injecting res1 by RPC and keeping `sync_all` is the least invasive option. The loop bound `range(len(remaining) + 1)` is also exactly sufficient for the worst case (a chain of N where each pass places only the deepest available parent).

## 1. The description doesn't describe the bug being fixed

> p2p relay only announces a transaction to peers that were already connected. so node 1 and node 2 cannot synch their mempool using sync_mempools().

That's the consequence of the *fix* — the reason `relay_mempool` had to be added once node 0 became genuinely isolated — not the defect. The defect is the one your code comment states correctly: `setup_network()` builds a mesh, so disconnecting only 0-1 left node 2 relaying res1 on to node 1, and whether that landed before `nodes[1].generate(2)` was a race.

Worth fixing before merge, since the description is what ends up in the merge commit.

## 2. Swallowing the exception loses the reason on failure

```python
try:
    node.sendrawtransaction(raw_tx)
except JSONRPCException:
    pass  # parent not yet present on this node; retry next pass
...
raise AssertionError("relay_mempool could not sync: %s" % remaining)
```

A transaction that is permanently unacceptable for some other reason takes the same path, so the helper exhausts its passes and reports nothing but a list of txids. For a helper added specifically to make a flaky test deterministic, that's the wrong thing to lose — the next time this fires, whoever looks at it gets no indication whether it was `missing-inputs`, a fee rule, or something else entirely.

Keeping the last exception per txid and folding it into the final `AssertionError` message would be enough.

## One question

I tried to find the failing run to check the fix against the real symptom, and couldn't: no `wallet_coloredcoin.py` failure appears in the last 30 `daily-test` runs or the last 14 `smoke-test` runs (the failures in that window are the fuzz branch's lint/macOS ones and `library-split`'s compile error). The comment says it was "won often enough under CI load", so I assume I'm just looking in the wrong place — where was it observed? Handy to know so the fix can be confirmed against the actual failure mode rather than only against the reasoning.

Naviabheeman:
The CI run which failed was in smoke test on macOS: https://github.com/chaintope/tapyrus-core/actions/runs/34440860643/job/102756695913

azuchi:
Thanks for the link — and the miss was mine. That run is at attempt 2, which passed, and `gh run view --job` serves the latest attempt's log by default, so I was reading the re-run. Attempt 1 has it:

```
attempt 1, macOS ARM64: failure
wallet_coloredcoin.py --scheme SCHNORR failed, Duration: 41 s
```

It also didn't turn up in my search because it was a smoke-test on #460's branch rather than a `daily-test` run. Sorry for sending you to look it up.

The symptom matches the diagnosis exactly, which is the part I wanted to confirm:

```
TestFramework (INFO): Options [1, 30, [], False]: found 7 tokens, expected 6
AssertionError: not(7 == 6)
  wallet_coloredcoin.py, line 723, in test_only_token_filter
```

One of the seven dumped entries carries `'confirmations': 1` — amount 100, label equal to its own color, held by node 0, i.e. res1. Had it stayed unconfirmed as intended, `listunspent(1, 30)` would have excluded it and the count would be 6. So res1 really did get mined, which is precisely the 0 → 2 → 1 relay race the fix addresses, and landing in the second of the two generated blocks explains the single confirmation. Good.

## `53e475e2` looks right

Preserving the last exception per txid does what I was after. One small thing: the filter drops any txid that ended up in `remaining` without having recorded an exception, so in that case the message loses the txid list the old one printed (and could even render as `{}`). Not a path I'd expect to hit often, but

```python
{txid: str(last_errors.get(txid, "no error recorded")) for txid in remaining}
```

keeps both pieces of information. Entirely optional.

## Still open: the description

It's unchanged, and still explains the consequence of the fix rather than the defect:

> p2p relay only announces a transaction to peers that were already connected. so node 1 and node 2 cannot synch their mempool using sync_mempools().

The defect is what your code comment states: `setup_network()` wires a mesh, so `disconnect_nodes(0, 1)` left node 2 relaying res1 on to node 1, and whether that beat `nodes[1].generate(2)` was a race — now demonstrated by the attempt-1 log above. Since the description becomes the merge commit, it's worth a quick edit before this lands.

Nothing else from me — CI is 14/14.

azuchi (approved):

PR #463, opened 2026-09-10
Splitting tapyrus_common and tapyrus_server, fix broken circular deps, missing CMake wiring, dead config

Naviabheeman:
Implement PR1 from the library split plan https://claude.ai/code/artifact/7ca340c5-bff6-4008-bc18-123f627b767b

The library-split branch's CMake refactor (11 scoped static libraries replacing the old monolithic tapyrus_common/tapyrus_server) had never actually been built to completion. This PR finishes it: fixes two real circular dependencies the split was supposed to eliminate but didn't catch, wires up every target that the split silently broke, and cleans up config-header cruft surfaced along the way.

- federationparams.cpp → CDynamicParams/genesis-block split — separates the per-run genesis-block-reading/CXFieldHistory-seeding lifecycle (new src/dynamicparams.{cpp,h}, src/genesisblock.{cpp,h}) out of static per-deployment identity, breaking a real circular dependency (FederationParams needed key_io.cpp; key_io.cpp needed FederationParams). Updates every caller and the unit tests to SelectDynamicParams()/DynamicParams().
- Breaks tapyrus_wallet's undeclared reverse dependency on tapyrus_rpc — three call sites (WalletInit::RegisterRPC, and interfaces/wallet.cpp's IssueToken/IssueReissuableToken/BurnToken calls) reached up into tapyrus_rpc-only symbols despite the declared dependency running the other way. Moved the offending code down to where it actually belongs (new wallet/walletinit_rpc.cpp in tapyrus_rpc, new wallet/tokenissuance.cpp in tapyrus_wallet), rather than just adding a link edge that would reintroduce the cycle.
- Breaks tapyrus_chainstate's undeclared dependency on tapyrus_peer — policy/packages.cpp called RelayTransaction()/g_connman directly. Added a new CValidationInterface::PackageTransactionsRelay signal (deliberately distinct from TransactionAddedToMempool, which fires mid-package and would reintroduce a free-relay-amplification bug); PeerLogicValidation subscribes and relays using its own connman.
- CMake wiring fixes — tapyrus_rpc, bench_tapyrus, and tapyrusd were each missing a direct leveldb link needed for headers reached only through tapyrus_chainstate's PRIVATE leveldb link. tapyrus-genesis, bench_tapyrus, tapyrus-qt, and test-tapyrus-qt were each missing the "glue" files (file_io.cpp/shutdown.cpp/ui_interface.cpp/etc.) that fatal-error paths in the split libraries call directly. tapyrus-qt itself was still linking the long-deleted tapyrus_server, never updated when the split originally landed.
- Smaller fixes found along the way: WITH_QRENCODE/WITH_DBUS checked an undefined ENABLE_GUI CMake variable instead of BUILD_GUI (always evaluated false); qt/test/test_main.cpp used #ifdef ENABLE_WALLET instead of #if (always true, since it's a #cmakedefine01 macro); addressbooktests.cpp was never gated behind ENABLE_WALLET despite needing CWallet directly; tapyrus_cli's dependencies were all PUBLIC with two unused; crypto/CMakeLists.txt and tapyrus-build-config.h.in carried tapyrus-only ENABLE_* SIMD macros with zero consumers and no upstream equivalent, now removed in favor of upstream's own HAVE_*-gated pattern.


azuchi:
Reviewed at `c8ee3792`. First, the part that's working: the lint output is real evidence the refactor achieves what it set out to.

```
Good job! The circular dependency "chainparams -> federationparams -> key_io -> chainparams" is no longer present.
... "chainparams -> federationparams -> primitives/block -> chainparams"
... "chainparams -> federationparams -> validation -> chainparams"
... "federationparams -> validation -> federationparams"
... "federationparams -> xfieldhistory -> federationparams"
... "txdb -> xfieldhistory -> txdb"
```

Six cycles gone, none introduced, and `EXPECTED_CIRCULAR_DEPENDENCIES` trimmed to match. Moving code to where it belongs rather than adding a link edge to paper over the cycle is the right instinct throughout.

CI is 11 red / 9 green, though, across three distinct causes. The first is a design question rather than a wiring gap.

## 1. `WalletInit`'s vtable can't be resolved across the archive boundary

Every Linux job and the Linux smoke build die here:

```
/usr/bin/ld: lib/libtapyrus_wallet.a(init.cpp.o):(.data.rel.ro._ZTV10WalletInit+0x20):
  undefined reference to `WalletInit::RegisterRPC(CRPCTable&) const'
FAILED: bin/tapyrusd
```

The referencing symbol is `_ZTV10WalletInit` — the vtable, not a call site. The class body is in `wallet/init.cpp` (`tapyrus_wallet`) while `RegisterRPC` alone compiles into `tapyrus_rpc` (`src/CMakeLists.txt:426`), and `tapyrusd` links `tapyrus_rpc` at `:474` before `tapyrus_wallet` at `:477`. GNU ld resolves static archives in a single left-to-right pass, so by the time `init.cpp.o` is pulled in, the archive holding the definition has already been scanned.

Reordering doesn't help: `tapyrus_rpc` genuinely needs `tapyrus_wallet` (`rpcwallet.cpp` uses `CWallet`), so wallet has to come after rpc. Splitting one virtual method of a polymorphic class across the two archives creates a symbol-level cycle between them — the same cycle the refactor is removing, relocated from the include graph to the vtable.

Three ways out, in the order I'd prefer them:

1. Keep `RegisterRPC` in `tapyrus_wallet` and invert the dependency for real — have `tapyrus_rpc` populate a registration hook, or let `tapyrusd`'s glue do the registering instead of `WalletInit`.
2. Move the whole `WalletInit` class up (into `tapyrus_rpc` or the glue) rather than one method of it.
3. Wrap the pair in `-Wl,--start-group/--end-group`. This is what #456's fuzz binaries had to do, but it conceals a real cycle and doesn't work the same way with Apple's ld64.

## 2. macOS: `tapyrus_peer` doesn't get libevent's include path

```
src/torcontrol.cpp:24:10: fatal error: 'event2/bufferevent.h' file not found
```

## 3. Windows cross: `ws2_32` missing from whichever target now holds `util.cpp`

```
build/src/./util.cpp:1215: undefined reference to `__imp_WSAStartup'
collect2: error: ld returned 1 exit status
```

2 and 3 are the same family — transitive include/link properties the monolith used to supply, now dropped.

## 4. `PackageTransactionsRelay`'s doc contradicts its implementation

The new signal's comment says "Called on a background thread", but `CMainSignals::PackageTransactionsRelay` dispatches synchronously (`validationinterface.cpp:188-190`), unlike every neighbouring signal, which goes through `m_schedulerClient.AddToProcessQueue`.

Synchronous is the right choice here — it preserves the old inline `RelayTransaction()` timing inside `SubmitPackageToMempool()`, so relay still happens before the RPC returns. It's the comment that should change. The reasoning for keeping this separate from `TransactionAddedToMempool` (not gossiping the admitted prefix of a partially-failed package) is sound, and `PeerLogicValidation`'s override reads correctly.

## One thing I couldn't check

At the previous head (`d36d66fd`), macOS built and `test_tapyrus` then aborted across most of the suite:

```
miner_tests/CreateNewBlock_validity:
  Assertion failed: (nThreadsServicingQueue == 0), function ~CScheduler, scheduler.cpp:20
then, for nearly every subsequent suite:
  Assertion failed: (secp256k1_context_sign == nullptr), function ECC_Start, key.cpp:360
```

Both are the signature of a fixture constructor throwing: the member `CScheduler` gets destroyed with its thread still running, and `ECC_Stop()` — which lives only in `~BasicTestingSetup` — never runs, so every later test's `ECC_Start()` asserts. The newly inserted `SelectDynamicParams(ReadGenesisBlock())` sits in that constructor after `ECC_Start()` and does file I/O that can throw, which is where I'd look first.

At this head macOS fails earlier (finding 2), so I can't tell whether that's still present. Worth re-checking once the build is green. Separately, making the fixture exception-safe would be worth doing regardless — as it stands, one failure poisons the entire rest of the run and hides its own cause.

## On scope

The plan described PR 1 as target definitions only, with `FederationParams` as the single code exception. What's here also adds a validation-interface signal, relocates wallet RPC registration across libraries, and touches init/qt/bench. I think that's the right call — those were real layering violations, and moving the code beats adding link edges. But it does change what reviewers should be looking at, so it's worth saying so in the description.

azuchi (changes requested):
Marking this changes-requested to match the findings in https://github.com/chaintope/tapyrus-core/pull/463#issuecomment-5657807803 — the detail is all there rather than repeated here.

CI is 11 red / 9 green at `c8ee3792`, across three causes:

1. `WalletInit`'s vtable (`_ZTV10WalletInit` in `libtapyrus_wallet.a`) references `RegisterRPC`, which compiles into `tapyrus_rpc` — an archive that has already been scanned by the time wallet is pulled in, and can't be reordered because `tapyrus_rpc` genuinely needs `CWallet`. This is the blocking one: splitting one virtual method of a polymorphic class across the two libraries recreates the cycle at the vtable level, so it needs a different approach rather than more wiring.
2. macOS: `torcontrol.cpp` can't find `event2/bufferevent.h` — `tapyrus_peer` isn't getting libevent's include path.
3. Windows cross: `util.cpp`'s `__imp_WSAStartup` is unresolved — `ws2_32` is missing from whichever target now holds it.

Also worth doing before this lands: `PackageTransactionsRelay`'s doc comment says it's called on a background thread while the implementation dispatches synchronously (the implementation is right, the comment isn't).

The six eliminated circular dependencies with none introduced is a genuinely good result, and none of the above changes that — it's the link-time consequences that still need working through.

azuchi:
Reviewed at `b7e0bff3` — all 20 checks green now, and the three build failures are properly fixed rather than worked around. The three test fixes found along the way are good work in their own right:

- `394c6be4` traced the `federationparams_tests` segfault to `CXFieldHistoryMap::Get()` dereferencing an `end()` iterator on a never-seeded map, and correctly noted it only crashed under libstdc++. Worth saying out loud that the UB is in `xfieldhistory.cpp:51` (`find(type)->second` with no presence check), not in the test — seeding the history fixes this suite, but a guard in `Get()` itself would be worth a follow-up, even if no production path can reach it today.
- `8759cf4d` nailed why `miner_tests` was flaky: re-deriving genesis from the same key relies on `createGenesisBlock()`'s `blockTime = time(0)` landing in the same second as the fixture's own derivation. Confirming it by forcing a 2-second gap is exactly the right way to prove that.
- `b7e0bff3` identified the ECC cascade more precisely than I could from the logs, and reproducing it with a deliberate `assert(false)` is convincing.

One note on that last one: `ECC_NeedsReset()` adds public API to `key.h` purely for the test harness. The same self-healing works entirely inside `test_tapyrus.cpp` with a file-static flag — set it after `ECC_Start()`, clear it in the destructor. A SIGABRT skips the destructor, so the flag survives as `true` and the next fixture knows to call `ECC_Stop()` first. That keeps a security-sensitive header free of a test-only entry point.

## `WalletInit`'s vtable: the cycle is now declared rather than removed

`f6cf0081` resolves the link failure by adding `tapyrus_wallet -> tapyrus_rpc` and letting CMake repeat the archives. It builds everywhere, and the comment explaining the key-function mechanism (`AddWalletOptions()` being the first non-inline virtual, so the vtable lands in `init.cpp.o`) is accurate. But it does mean the top two libraries in a refactor whose purpose was to remove circular dependencies now have a declared cycle between them, and `tapyrus_wallet` can never be linked without `tapyrus_rpc` again.

For reference, Bitcoin Core doesn't have this edge. Wallet RPC registration lives in the wallet library and reaches the RPC table through an abstract interface rather than a direct symbol:

```cpp
// src/wallet/interfaces.cpp — part of bitcoin_wallet
void registerRpcs() override {
    for (const CRPCCommand& command : GetWalletRPCCommands()) {
        ...
        m_rpc_handlers.emplace_back(m_context.chain->handleRpc(m_rpc_commands.back()));
    }
}
```

`Chain::handleRpc` is pure virtual (`interfaces/chain.h:350`), implemented node-side (`node/interfaces.cpp:794`), and `init.cpp` drives it polymorphically over `node.chain_clients`. The result is that `bitcoin_wallet` links only `bitcoin_common` — no node, no rpc — and the graph stays a DAG.

I'm not suggesting you build `interfaces::Chain` here. Tapyrus has no equivalent (`src/interfaces/` is just `handler`/`node`/`wallet`), and creating that boundary is a project of its own, well outside this PR.

What does look in scope is the tier this PR already invented. `WalletInit` is init-time wiring, not wallet logic: it's instantiated exactly once (`wallet/init.cpp:18`, as `g_wallet_init_interface`) and consumed only by `src/init.cpp`. That is precisely the description of the glue files this PR compiles directly into executables rather than into a library. Moving `wallet/init.cpp` and `wallet/walletinit_rpc.cpp` (255 lines together) into that tier removes the cycle with no new abstraction and no ABI subtleties, and it uses the PR's own vocabulary. The executables that need `g_wallet_init_interface` would have to pick up the glue, but the PR has already done that distribution for tapyrus-genesis, bench, tapyrus-qt and test-tapyrus-qt.

(There's also a one-line trick: declare `RegisterRPC()` first in `walletinit.h` so it becomes the key function, putting the vtable in `tapyrus_rpc` and leaving only the existing rpc → wallet direction. It works, but it makes the layering depend on declaration order plus an ABI rule, and a future reorder would silently reintroduce the cycle. I'd rather not see that as the permanent answer.)

If you'd prefer to keep the cycle and clean it up separately, that's a reasonable call too — but then please leave a TODO at the declaration pointing at the Core pattern, so it doesn't quietly become permanent.

## `PackageTransactionsRelay`'s doc still contradicts the code

`validationinterface.h:162`'s comment says "Called on a background thread", while `CMainSignals::PackageTransactionsRelay` dispatches synchronously (`validationinterface.cpp:188-190`) — unlike every neighbouring signal, which goes through `m_schedulerClient.AddToProcessQueue`.

Synchronous is the right behaviour: it preserves the inline `RelayTransaction()` timing that `SubmitPackageToMempool()` had, so relay still happens before the RPC returns. It's the comment that needs the edit.

azuchi:
`060d1db4` and `3e686923` resolve both open findings, and the `WalletInit` move came out cleaner than I expected. The declared graph is a DAG again (`tapyrus_wallet` no longer lists `tapyrus_rpc`), the glue is distributed to tapyrus-qt and test-tapyrus-qt alongside tapyrusd, and as a bonus the key-function question disappears entirely — both TUs now land in the same executable, so which one emits the vtable stops mattering. `tapyrus_wallet` is also linkable on its own again, which keeps the door open for a wallet-only tool later. The `PackageTransactionsRelay` comment now explains *why* synchronous is correct rather than just stating it, which is better than what I asked for.

One thing left, and it's the smallest item on the list.

## `ECC_NeedsReset()` in `key.h`

`b7e0bff3`'s self-healing is worth keeping — one aborted fixture poisoning every later test is exactly the failure mode that hid the real cause here. My only reservation is where the mechanism lives: `key.h`/`key.cpp` gain a public entry point that exists solely for the test harness, in a file whose surface I'd rather keep to what production code needs.

The same behaviour fits entirely inside `test_tapyrus.cpp`:

```cpp
static bool g_ecc_started = false;   // file-static in test_tapyrus.cpp

BasicTestingSetup::BasicTestingSetup(...)
{
    ...
    if (g_ecc_started) ECC_Stop();   // a prior fixture died before its destructor ran
    ECC_Start();
    g_ecc_started = true;
    ...
}

BasicTestingSetup::~BasicTestingSetup()
{
    ...
    ECC_Stop();
    g_ecc_started = false;
}
```

It covers the case your commit message describes for the same reason your version does: a SIGABRT unwinds past the destructor, so the flag is never cleared and the next fixture knows to call `ECC_Stop()` first. The `catch (...)` path you added already calls `ECC_Stop()` directly, so it just needs the flag cleared there too.

The one behavioural difference is that the flag tracks "this fixture started ECC", whereas `ECC_NeedsReset()` reads the real global. If anything outside the fixtures ever called `ECC_Start()`, the flag wouldn't know — but nothing in the test binary does, and if that changed, the assert would be the correct response anyway.

Not a blocker, and I won't hold the PR on it. If you'd rather keep `ECC_NeedsReset()`, please at least mark it test-only in its doc comment so nobody mistakes it for part of the key API.

With that said, everything else I raised is addressed and CI is 20/20 — happy to see this go in.

azuchi (approved):

